Issue #20, opened 2017-12-08
Saving text file back to Azure loses properties and metadata


StephenWeatherford:

StephenWeatherford:
Original file as uploaded and edited in Storage Explorer:

![image](https://user-images.githubusercontent.com/6913354/33784799-0b98732a-dc17-11e7-8b8f-fe962e86c7c0.png)

After modification and upload from extension:
![image](https://user-images.githubusercontent.com/6913354/33784832-2ed2ccdc-dc17-11e7-9432-e82dfa24937d.png)

PR #24, opened 2017-12-11
Fixed editing files in folders. Updated edit logic to use existing contentType.

JakeRadMSFT:
Fixed editing files in folders.
Updated edit logic to use existing contentType.


JakeRadMSFT:
Resolves #20

StephenWeatherford:
This only partial resolves #20

Issue #16, opened 2017-12-08
Storage accounts are not sorted alphabetically


StephenWeatherford:
Makes it difficult if you have a lot of accounts.

JakeRadMSFT:
Resolved in V0.1.3

PR #26, opened 2017-12-12
Upgrade to use common Azure TreeDataProvider.

JakeRadMSFT:


JakeRadMSFT:
(FYI - I just did it compiled and ran it. There are likely issues  (possibly many) just throwing it out for some extra eyes on it.

src/azureStroageExplorer/blobContainers/blobContainerGroupNode.ts
```diff
@@ -3,47 +3,39 @@
...
-            label: this.label,
-            collapsibleState: TreeItemCollapsibleState.Collapsed,
-            contextValue: 'azureBlobContainerGroup',
-            iconPath: {
-				light: path.join(__filename, '..', '..', '..', '..', '..', 'resources', 'light', 'AzureBlob_16x.png'),
-				dark: path.join(__filename, '..', '..', '..', '..', '..', 'resources', 'dark', 'AzureBlob_16x.png')
-			}
-        }
-    }
+    public id: string = "Blob Containers";
```
JakeRadMSFT: What should ids be? Do they need to be globally unique?

src/azureStroageExplorer/storageAccountProvider.ts
```diff
@@ -0,0 +1,28 @@
...
+ *--------------------------------------------------------------------------------------------*/
+
+import StorageManagementClient = require('azure-arm-storage');
+import { StorageAccount } from '../../node_modules/azure-arm-storage/lib/models';
+import { IChildProvider, IAzureTreeItem, IAzureNode } from 'vscode-azureextensionui';
+import { StorageAccountNode } from './storageAccounts/storageAccountNode';
+import { Subscription } from 'azure-arm-resource/lib/subscription/models';
+
+export class StorageAccountProvider implements IChildProvider { 
+    async loadMoreChildren(node: IAzureNode, _clearCache: boolean): Promise<IAzureTreeItem[]> {
```
JakeRadMSFT: This is an odd interface for me. I don't like the IAzureNode wrapper... why not just give me the IAzureTreeItem I made and then another context object of some kind.

src/azureStroageExplorer/blobContainers/blobContainerGroupNode.ts
EricJizbaMSFT: "A unique id to identify this tree item. This id is also used for openInPortal"

https://github.com/Microsoft/vscode-azuretools/blob/master/ui/index.d.ts#L74

src/azureStroageExplorer/storageAccountProvider.ts
EricJizbaMSFT: Just want to make sure you read the README [here](https://github.com/Microsoft/vscode-azuretools/tree/master/ui) and the docs [here](https://github.com/Microsoft/vscode-azuretools/blob/master/ui/index.d.ts) before we go to deep on this

The node interface has several helpful methods and I'm not sure how we would provide that value without it (although I'm open to suggestions). I'll just give one example: if your tree item implements `IAzureTreeItem.deleteTreeItem` - you're never supposed to call `deleteTreeItem` directly. You're supposed to call `IAzureNode.deleteNode` (which wraps your implementation and ensures the item is removed from the tree)

src/azureStroageExplorer/fileShares/fileShareNode.ts
```diff
@@ -3,57 +3,46 @@
...
-				dark: path.join(__filename, '..', '..', '..', '..', '..', 'resources', 'dark', 'AzureFileShare_16x.png')
-			}
-        }
+        public readonly key: StorageAccountKey) {		
     }
+    public id: string = this.share.name;
+    public label: string = this.share.name;
+    public contextValue: string = 'azureFileShare';
+    public iconPath: { light: string | Uri; dark: string | Uri } =  {
+        light: path.join(__filename, '..', '..', '..', '..', '..', 'resources', 'light', 'AzureBlob_16x.png'),
```
JakeRadMSFT: Wrong image.

JakeRadMSFT:
Resolves #16

src/azureStroageExplorer/blobContainers/blobContainerGroupNode.ts
JakeRadMSFT: Why though? I'd like it to be optional ... otherwise I'm inventing unique IDs where they don't need to be.

src/azureStroageExplorer/storageAccountProvider.ts
JakeRadMSFT: Yeah still odd to me ...

JakeRadMSFT: We can discuss offline ...

src/azureStroageExplorer/blobContainers/blobContainerActionHandler.ts
```diff
@@ -20,11 +21,11 @@ export class BlobContainerActionHandler extends BaseActionHandler {
         this.initEvent(context, 'azureStorage.blobEditor.onDidSaveTextDocument', vscode.workspace.onDidSaveTextDocument, (doc: vscode.TextDocument) => this._editor.onDidSaveTextDocument(context.globalState, doc));
     }
 
-    openBlobContainerInStorageExplorer(node: BlobContainerNode) {
-        var resourceId = node.storageAccount.id;
-        var subscriptionid = node.subscription.subscriptionId;
+    openBlobContainerInStorageExplorer(node: IAzureNode<BlobContainerNode>) {
+        var resourceId = node.treeItem.storageAccount.id;
+        var subscriptionid = node.treeItem.subscription.subscriptionId;
```
EricJizbaMSFT: Just use `node.subscription` directly instead of using a subscription stored in the treeItem. Part of the purpose of `IAzureNode` is to store relevant information about the tree (aka parent, subscription, etc.) so you don't have to pass it down the tree when you're creating children

src/azureStroageExplorer/blobContainers/blobContainerGroupNode.ts
```diff
@@ -3,47 +3,39 @@
...
+import { IAzureParentTreeItem, IAzureTreeItem } from 'vscode-azureextensionui';
+import { Uri } from 'vscode';
 
-export class BlobContainerGroupNode extends AzureLoadMoreTreeNodeBase {
+export class BlobContainerGroupNode implements IAzureParentTreeItem {
     private _continuationToken: azureStorage.common.ContinuationToken;
 
     constructor(
-        public readonly subscription: SubscriptionModels.Subscription, 
+        public readonly subscription: Subscription,
```
EricJizbaMSFT: No need to store the subscription. It's already given to you in `IAzureNode`

src/azureStroageExplorer/blobContainers/blobEditor.ts
```diff
@@ -8,32 +8,33 @@ import * as azureStorage from "azure-storage";
 import { BaseEditor } from '../../azureServiceExplorer/editors/baseEditor';
 import { AzureStorageOutputChanel } from '../azureStorageOutputChannel';
 
-export class BlobEditor extends BaseEditor<BlobNode> {
+import { IAzureNode } from 'vscode-azureextensionui';
+export class BlobEditor extends BaseEditor<IAzureNode<BlobNode>> {
```
EricJizbaMSFT: Seems like you could've left this file alone and just passed in node.treeItem to the blobEditor

src/azureStroageExplorer/fileShares/fileShareActionHandler.ts
```diff
@@ -21,11 +22,11 @@ export class FileShareActionHandler extends BaseActionHandler {
         this.initEvent(context, 'azureStorage.fileEditor.onDidSaveTextDocument', vscode.workspace.onDidSaveTextDocument, (doc: vscode.TextDocument) => this._editor.onDidSaveTextDocument(context.globalState, doc));
     }
 
-    openFileShareInStorageExplorer(node: FileShareNode) {
-        var resourceId = node.storageAccount.id;
-        var subscriptionid = node.subscription.subscriptionId;
+    openFileShareInStorageExplorer(node: IAzureNode<FileShareNode>) {
+        var resourceId = node.treeItem.storageAccount.id;
+        var subscriptionid = node.treeItem.subscription.subscriptionId;
```
EricJizbaMSFT: Again, use node.subscription directly

src/azureStroageExplorer/blobContainers/blobContainerGroupNode.ts
```diff
@@ -3,47 +3,39 @@
...
+    public contextValue: string = 'azureBlobContainerGroup';
+    public iconPath: { light: string | Uri; dark: string | Uri } =  {
+        light: path.join(__filename, '..', '..', '..', '..', '..', 'resources', 'light', 'AzureBlob_16x.png'),
+        dark: path.join(__filename, '..', '..', '..', '..', '..', 'resources', 'dark', 'AzureBlob_16x.png')
+    };
+
 
-    async getMoreChildren(): Promise<AzureTreeNodeBase[]> {
+    async loadMoreChildren(): Promise<IAzureTreeItem[]> {
         var containers = await this.listContainers(this._continuationToken);
```
EricJizbaMSFT: You're currently ignoring the "clearCache" parameter. If `clearCache` is true, you're supposed to restart loading (aka don't use the continuation token)

EricJizbaMSFT: NOTE: Since you don't let users refresh a specific node in the tree, this isn't causing a bug in your code yet. It will lead to problems once they _can_ refresh a specific node.

src/azureStroageExplorer/blobContainers/blobEditor.ts
JakeRadMSFT: Perhaps... not worried about see other review.

PR #32, opened 2018-01-28
Creating and deleting blob containers

StephenWeatherford:


src/azureServiceExplorer/actions/baseActionHandler.ts
```diff
@@ -4,82 +4,32 @@
...
+    private _azureActionHandler: AzureActionHandler;
 
-    initCommand(context: vscode.ExtensionContext, commandId: string, callback: (...args: any[]) => any) {
-        this.initAsyncCommand(context, commandId, (...args: any[]) => Promise.resolve(callback(...args)));
-    }
+    abstract registerActions(context: vscode.ExtensionContext);
 
-    initAsyncCommand(context: vscode.ExtensionContext, commandId: string, callback: (...args: any[]) => Promise<any>) {
-        context.subscriptions.push(vscode.commands.registerCommand(commandId, this.wrapAsyncCallback(commandId, callback)));
+    private ensureHandler(context: vscode.ExtensionContext): void {
```
JakeRadMSFT: Is this the pattern other extensions are using? Do we still need BaseActionHandler?

package.json
```diff
@@ -140,17 +148,53 @@
...
+				{
+					"command": "azureStorage.refresh",
+					"when": "never"
+				},
+				{
+					"command": "azureStorage.openBlobContainer",
+					"when": "never"
+				},
+				{
+					"command": "azureStorage.openBlobContainer",
```
EricJizbaMSFT: openBlobContainer is listed here twice

package.json
```diff
@@ -205,8 +249,8 @@
 		"azure-storage": "^2.6.0",
 		"ms-rest": "^2.2.2",
 		"ms-rest-azure": "^2.3.1",
-		"vscode-extension-telemetry": "^0.0.6",
-		"vscode-azureextensionui": "~0.3.0",
+		"vscode-extension-telemetry": "^0.0.10",
+		"vscode-azureextensionui": "^0.5.3",
```
EricJizbaMSFT: I think we should continue using '~' instead of '^' for the ui package. It's still in 'preview' and so minor version updates could correspond to breaking changes

package.json
```diff
@@ -140,17 +148,53 @@
...
+					"when": "never"
+				},
+				{
+					"command": "azureStorage.copyConnectionString",
+					"when": "never"
+				},
+				{
+					"command": "azureStorage.copyPrimaryKey",
+					"when": "never"
+				}
```
EricJizbaMSFT: Add create/delete here since the node picker is not being used yet

src/azureServiceExplorer/actions/baseActionHandler.ts
StephenWeatherford: We should get rid of BaseActionHandler completely, as a separate PR.

package.json
StephenWeatherford: was meant to be createBlobContainer/deleteBlobContainer, fixed.

package.json
StephenWeatherford: Done.

StephenWeatherford:
Part of fix for #35

PR #38, opened 2018-02-01
Create/delete blobs, directories and files

StephenWeatherford:


src/azureStroageExplorer/fileShares/childKind.ts
```diff
@@ -0,0 +1,9 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+export enum ChildKind {
+    Subdirectory,
```
EricJizbaMSFT: Nit: It feels like you could use FileNode.contextValue and DirectoryNode.contextValue rather than create an enum representing the same information

You'd have to make the field static. We do it like this in other repos to support the node picker:
```
public static contextValue: string = 'azureFileShareDirectory';
public contextValue: string = DirectoryNode.contextValue;
```

src/azureStroageExplorer/fileShares/directoryNode.ts
```diff
@@ -63,4 +71,78 @@ export class DirectoryNode implements IAzureParentTreeItem {
...
+        if (<ChildKind>userOptions === ChildKind.File) {
+            return this.askAndCreateEmptyTextFile(showCreatingNode);
+        } else {
+            return askAndCreateChildDirectory(this.share, this.fullPath, this.storageAccount, this.key, showCreatingNode);
+        }
+    }
+
+    public async deleteTreeItem(_node: IAzureNode): Promise<void> {
+        // Note: Azure will fail the directory delete if it's not empty, so no need to ask about deleting contents
+        const message: string = `Are you sure you want to delete the directory '${this.label} (it must be empty)'?`;
```
EricJizbaMSFT: In other contexts we ask 'Are you sure you want to delete ... and all its contents?'

Why don't we just delete the contents for them? Although this could be a separate feature...

Also - can we check if the directory is empty _before_ asking if they want to delete it? Then we would just show them a warning like "You cannot delete a directory that is not empty"

src/azureStroageExplorer/fileShares/validateNames.ts
```diff
@@ -0,0 +1,44 @@
...
+
+
+    return undefined;
+}
+
+export function validateFileName(name: string): string | undefined | null {
+    const validLength = { min: 1, max: 255 };
+
+    if (!name) {
+        return "Filename cannot be empty";
```
EricJizbaMSFT: nit: 'File name' instead of 'Filename'

package.json
```diff
@@ -323,11 +407,11 @@
...
 		"azure-storage": "^2.6.0",
+		"copy-paste": "^1.3.0",
+		"fs-extra": "4.0.2",
 		"ms-rest": "^2.2.2",
 		"ms-rest-azure": "^2.3.1",
 		"vscode-extension-telemetry": "^0.0.10",
-		"vscode-azureextensionui": "~0.5.3",
-		"fs-extra": "4.0.2",
-		"copy-paste": "^1.3.0"
+		"vscode-azureextensionui": "0.5.5"
```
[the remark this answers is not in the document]
StephenWeatherford: No, wasn't trying to get npm i to work, missed setting this back.

src/azureStroageExplorer/fileShares/childKind.ts
StephenWeatherford: Like that, thanks.

src/azureStroageExplorer/fileShares/directoryNode.ts
```diff
@@ -3,17 +3,21 @@
...
+import { askAndCreateChildDirectory } from './createDirectories';
+import { DialogBoxResponses } from '../../constants';
+import { ChildKind } from './childKind';
+import { validateFileName } from './validateNames';
 
 export class DirectoryNode implements IAzureParentTreeItem {
     constructor(
-        public readonly relativeDirectory: string,
-        public readonly directory: azureStorage.FileService.DirectoryResult,
+        public readonly relativeDirectory: string, // full path of the parent
```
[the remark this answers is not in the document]
StephenWeatherford: I agree.  Was not my code, but changing it now.

src/azureStroageExplorer/fileShares/directoryNode.ts
StephenWeatherford: I'm considering it a separate feature (also could take a very long time).  Maybe I'll get time for it today.

src/azureStroageExplorer/fileShares/validateNames.ts
StephenWeatherford: Why?  Filename is most often written as one word.

package.json
```diff
@@ -189,6 +217,34 @@
...
+					"command": "azureStorage.createSubdirectory",
+					"when": "view == azureStorage && viewItem == azureFileShareDirectory"
+				},
+				{
+					"command": "azureStorage.deleteDirectory",
+					"when": "view == azureStorage && viewItem == azureFileShareDirectory"
+				},
+				{
+					"command": "azureStorage.createTextFile",
+					"when": "view == azureStorage && viewItem == azureFileShareDirectory"
```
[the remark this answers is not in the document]
StephenWeatherford: Good catch.

src/azureStroageExplorer/fileShares/validateNames.ts
EricJizbaMSFT: English is dumb 🙃. I guess both are acceptable spellings - I just figured all the other ones were 'Blob name' and 'Table name' and whatnot

PR #41, opened 2018-02-01
Add lint (not enforced) and travis integration

StephenWeatherford:


whiskeyjay:
Why include package-lock.json? I thought it is supposed to be in the ignored list.

nturinski:
https://stackoverflow.com/questions/44206782/do-i-commit-the-package-lock-json-file-created-by-npm-5

I think we ended up ignoring it because of all the merge conflicts it would create though?

PR #58, opened 2018-02-07
Add screenshots for create blob container/create blob

StephenWeatherford:


README.md
```diff
@@ -2,13 +2,19 @@
...
    ![Explore Containers, Shares, Queues, and Tables](resources/exploreScreenshot.png)
+   ![Create Blob Containers](resources/createBlobContainerScreenshot.png)
+
  * Access Connection String and Primary Key
 
    ![Copy Connection String and Primary Key](resources/storageActionsScreenshot.png)
+
  * Create/Edit/Delete Block Blobs and Files
+
+   ![Create Blob Blobs](resources/blobContainerActionsScreenshot.png)
```
EricJizbaMSFT: Block Blob not Blob Blob?

README.md
```diff
@@ -2,13 +2,19 @@
 The Azure Storage extension for VS Code lets you browse Azure Blob Containers, File Shares, Tables, and Queues.
 
 ## Features
- * Explore/Create/Delete Blob Containers, File Shares, Queues and Tables
+* Explore/Create/Delete Blob Containers, File Shares, Queues and Tables
 
    ![Explore Containers, Shares, Queues, and Tables](resources/exploreScreenshot.png)
+   ![Create Blob Containers](resources/createBlobContainerScreenshot.png)
```
EricJizbaMSFT: Can we combine these two screenshots into one? The 'Create' picture pretty much has all the same info except the right-click-menu is open

README.md
```diff
@@ -2,13 +2,19 @@
...
 
    ![Copy Connection String and Primary Key](resources/storageActionsScreenshot.png)
+
  * Create/Edit/Delete Block Blobs and Files
+
+   ![Create Blob Blobs](resources/blobContainerActionsScreenshot.png)
+
  * Open in Storage Explorer for memory or computationally heavy tasks, or for upload/download of large or non-text files.
 
    ![Open In Storage Explorer](resources/openBlobContainerScreenshot.png)
```
EricJizbaMSFT: Do we need to update this picture with the new 'Storage Explorer' wording?

README.md
JakeRadMSFT: So I'm not the only on to make typos :)

MRayermannMSFT: Super nit pick: the node highlighting in this picture is weird (two nodes have highlighting). Won't block the review for this, but just in case you care. *shrug*

MRayermannMSFT: The highlight color is also different from the existing pics.

README.md
StephenWeatherford: Done.

README.md
StephenWeatherford: Fixed

PR #61, opened 2018-02-07
Pull out some lint

StephenWeatherford:


src/azureStorageExplorer/fileShares/directoryUtils.ts
```diff
@@ -67,7 +67,9 @@ export async function deleteDirectoryAndContents(directory: string, share: strin
     const parallelOperations = 5;
     const maxResults = 50;
 
+    // tslint:disable-next-line:no-unnecessary-initializer
```
EricJizbaMSFT: I mean...you could just remove `= undefined` from the line below rather than disable the rule

tslint.json
```diff
@@ -38,19 +38,19 @@
         "mocha-no-side-effect-code": true,
         "no-any": false /*TODO: true*/,
         "no-arg": true,
-        "no-backbone-get-set-outside-model": false /*TODO: true*/,
+        "no-backbone-get-set-outside-model": false, // changed
         "no-bitwise": true,
         "no-conditional-assignment": true,
         "no-console": [
-            false /*TODO: true*/,
+            false, // changed
```
EricJizbaMSFT: Just going on record saying I would prefer to keep "no-console" turned on - but by no means will I block the PR for this.

Here's my reasoning: It's fine for personal development, but I don't like including it in production. It doesn't provide any value once it's installed on a user's machine, but it _does_ increase the likelihood for bugs (for example if someone calls `console.log(example.toString())`, but `example` is null).  If we're actually logging something important, it should go to the outputChannel or to our telemetry.

StephenWeatherford: Yeah, we disagree on this one.  There is information that is useful to us in developing.  Ideally this would be handled with more finesse but I'm not planning on fixing everything in this codebase for a backfitting of tslint.  This falls below the bang for buck line for me.

Thanks for the comment. We appreciate your business.

src/azureStorageExplorer/fileShares/directoryUtils.ts
StephenWeatherford: I considered that, but it seems important in this case to indicate that the variable was intentionally left undefined. It's even passed in to a function while still undefined, which would feel like an error to me if it was simply declared without a value.

So I'm hesitating on this rule but not ready to give up completely on it yet.  But my opinion is less strong on this one.

PR #67, opened 2018-02-08
Yet more lint

StephenWeatherford:


src/azureStorageExplorer/blobContainers/blobContainerGroupNode.ts
```diff
@@ -19,7 +19,7 @@ export class BlobContainerGroupNode implements IAzureParentTreeItem {
         public readonly key: StorageAccountKey) {
     }
 
-    public id: string = undefined;
+    public id: string;
```
EricJizbaMSFT: It shouldn't make any functional difference, but just FYI the id field is optional in v0.6.0 of the ui package and so you could just remove all of these if you so desired

(See this PR for more info https://github.com/Microsoft/vscode-azuretools/pull/69)

JakeRadMSFT: Nice!

StephenWeatherford: Will do separately.

PR #80, opened 2018-02-15
regedit -> node-winreg, fixes #79

### LINKED ISSUE #79: Reading Windows registry to open in Storage Explorer sometimes fails
Complaining that the swap file is too small (which doesn't seem to be the case). 

C:\Repos\vscode-azurestorage\node_modules\regedit\vbs\regListStream.wsf(0, 2) SWbemObjectEx: Die Auslagerungsdatei ist zu klein, um diesen Vorgang durchzuführen.

StephenWeatherford:
node-winreg is already used by vscode, so it makes sense to use that.  It uses a less complicated method and is working for me when regedit is not.


package.json
```diff
@@ -409,7 +409,7 @@
 		"watch": "tsc -watch -p ./"
 	},
 	"optionalDependencies": {
```
JakeRadMSFT: I'd try making this required

StephenWeatherford: Worked fine.

src/storageExplorerLauncher/windowsStorageExplorerLauncher.ts
```diff
@@ -9,17 +9,18 @@ import * as os from "os";
...
 
-// regedit doesn't exist for Mac. I have to import like this so it builds.
-let regedit: any;
+// winreg doesn't exist for Mac. I have to import like this so it builds.
+let winreg: WinregStatic;
 if (os.platform() === "win32") {
     // tslint:disable-next-line:no-require-imports
     // tslint:disable-next-line:no-var-requires
-    regedit = require("regedit");
+    winreg = require("winreg");
```
craxal: Depending on the version of Typescript being used, you can do this *and* `import * as winreg from "winreg";`. The import will function as a typings import, and the conditional one will contain the actual module exports.

src/storageExplorerLauncher/windowsStorageExplorerLauncher.ts
```diff
@@ -70,22 +76,22 @@ export class WindowsStorageExplorerLauncher implements IStorageExplorerLauncher
...
-    private static getWindowsRegistryValue(key: string): Promise<string> {
+    private static getWindowsRegistryValue(hive: string, key: string): Promise<string> {
         return new Promise((resolve, reject) => {
-            regedit.list([key])
-                .on('data', (entry) => {
-                    let value = <string>entry.data.values[""].value.split("\"")[1];
-                    resolve(value);
-                })
-                .on('error', (err) => {
+            let regKey = new winreg({ hive, key });
```
craxal: `winreg` is going to be `undefined` if this isn't running in Windows.

StephenWeatherford: np. This code never runs on windows.

craxal: How do we know this won't run on Windows?

StephenWeatherford: Because it checks for os before instantiating.

PR #88, opened 2018-02-27
Support text blob upload/download

StephenWeatherford:
@EricJizbaMSFT One issue... For upload, I've got it currently set up via createChild. But if the blob already exists and user chooses to overwrite, you end up with two nodes. So either I need to delete the original node before upload, or need to not use createChild, in which case I need to manually added a node, not sure how to do that.  Suggestions?

### DIFF
--- a/package.json
+++ b/package.json
@@ -121,8 +121,16 @@
 				"title": "Create Empty Text Blob"
 			},
 			{
+				"command": "azureStorage.uploadBlockBlob",
+				"title": "Upload block blob..."
+			},
+			{
 				"command": "azureStorage.deleteBlob",
-				"title": "Delete Blob"
+				"title": "Delete"
+			},
+			{
+				"command": "azureStorage.downloadBlob",
+				"title": "Download..."
 			},
 			{
 				"command": "azureStorage.createFileShare",
@@ -150,7 +158,7 @@
 			},
 			{
 				"command": "azureStorage.deleteFile",
-				"title": "Delete File"
+				"title": "Delete"
 			},
 			{
 				"command": "azureStorage.createTable",
@@ -235,7 +243,15 @@
 					"when": "view == azureStorage && viewItem == azureBlobContainer"
 				},
 				{
+					"command": "azureStorage.uploadBlockBlob",
+					"when": "view == azureStorage && viewItem == azureBlobContainer"
+				},
+				{
 					"command": "azureStorage.deleteBlob",
+					"when": "view == azureStorage && viewItem == azureBlob"
+				},
+				{
+					"command": "azureStorage.downloadBlob",
 					"when": "view == azureStorage && viewItem == azureBlob"
 				},
 				{
@@ -301,7 +317,15 @@
 					"when": "never"
 				},
 				{
+					"command": "azureStorage.uploadBlockBlob",
+					"when": "never"
+				},
+				{
 					"command": "azureStorage.deleteBlob",
+					"when": "never"
+				},
+				{
+					"command": "azureStorage.downloadBlob",
 					"when": "never"
 				},
 				{
--- a/package.json
+++ b/package.json
@@ -454,7 +454,7 @@
 		"ms-rest": "^2.2.2",
 		"ms-rest-azure": "^2.3.1",
 		"vscode-extension-telemetry": "^0.0.10",
-		"vscode-azureextensionui": "~0.7.0",
+		"vscode-azureextensionui": "~0.7.1",
 		"winreg": "^1.2.3"
 	}
 }
--- a/src/azureServiceExplorer/editors/RemoteFileEditor.ts
+++ b/src/azureServiceExplorer/editors/RemoteFileEditor.ts
@@ -35,26 +35,26 @@ export class RemoteFileEditor<ContextT> implements vscode.Disposable {
     async showEditor(context: ContextT): Promise<void> {
         let fileName = await this.remoteFileHandler.getFilename(context);
 
-        this.appendLineToOutput(`Opening '${fileName}' ...`);
         try {
             let parsedPath: path.ParsedPath = path.posix.parse(fileName);
             let temporaryFilePath = await TemporaryFile.create(parsedPath.base);
             await this.remoteFileHandler.downloadFile(context, temporaryFilePath);
             await this.showEditorFromFile(context, temporaryFilePath);
-            this.appendLineToOutput(`Successfully opened '${fileName}'`);
         } catch (error) {
-            let details: string;
+            if (!(error instanceof UserCancelledError)) {
+                let details: string;
 
-            if (!!error.message) {
-                details = error.message;
-            } else {
-                details = JSON.stringify(error);
+                if (!!error.message) {
+                    details = error.message;
+                } else {
+                    details = JSON.stringify(error);
+                }
+
+                this.appendLineToOutput(`Unable to open '${fileName}'`);
+                this.appendLineToOutput(`Error Details: ${details}`);
+
+                await window.showWarningMessage(`Unable to open "${fileName}". Please check Output for more information.`);
             }
-
-            this.appendLineToOutput(`Unable to open '${fileName}'`);
-            this.appendLineToOutput(`Error Details: ${details}`);
-
-            await window.showWarningMessage(`Unable to open "${fileName}". Please check Output for more information.`);
         }
     }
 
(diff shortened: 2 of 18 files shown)


package.json
```diff
@@ -120,9 +120,17 @@
...
+				"title": "Upload block blob..."
+			},
 			{
 				"command": "azureStorage.deleteBlob",
-				"title": "Delete Blob"
+				"title": "Delete"
+			},
+			{
+				"command": "azureStorage.downloadBlob",
+				"title": "Download..."
```
EricJizbaMSFT: How do you decide which titles have '...' and which don't?

src/azureServiceExplorer/editors/RemoteFileEditor.ts
```diff
@@ -86,9 +86,16 @@ export class RemoteFileEditor<ContextT> implements vscode.Disposable {
...
     private async showEditorFromFile(context: ContextT, localFilePath: string): Promise<void> {
+        this.appendLineToOutput("Opening...");
         const document = await vscode.workspace.openTextDocument(localFilePath);
-        this.fileMap[localFilePath] = [document, context];
-        await vscode.window.showTextDocument(document);
+        if (document) {
+            this.fileMap[localFilePath] = [document, context];
+            await vscode.window.showTextDocument(document);
+            this.appendLineToOutput(`Successfully opened '${localFilePath}'`);
+        } else {
```
EricJizbaMSFT: I still think we should switch over to the shared editor in the ui package ASAP. Every time we edit this file, they diverge even more and we lose fixes on each side.

src/azureStorageExplorer/blobContainers/blobContainerNode.ts
```diff
@@ -142,4 +167,94 @@ export class BlobContainerNode implements IAzureParentTreeItem {
...
+    private async createChildFromUpload(node: IAzureNode<BlobContainerNode>, showCreatingNode: (label: string) => void): Promise<IAzureTreeItem> {
+        let uris = await vscode.window.showOpenDialog(
+            <vscode.OpenDialogOptions>{
+                canSelectFiles: true,
+                canSelectFolders: false,
+                canSelectMany: false,
+                defaultUri: lastUploadFolder,
+                filters: {
+                    "Text files": [
+                        'csv',
```
EricJizbaMSFT: How'd you come up with this list? What happens if we don't restrict the user? If they get a helpful error message somewhere down the line, then I would prefer not to use a hard-coded list like this

src/azureStorageExplorer/blobContainers/blobFileHandler.ts
```diff
@@ -18,24 +26,79 @@ export class BlobFileHandler implements IRemoteFileHandler<IAzureNode<BlobNode>>
...
+            await Limits.askOpenInStorageExplorer(
+                `Please use Storage Explorer to upload files larger than ${Limits.maxUploadDownloadSizeMB}MB.`,
+                node.treeItem.storageAccount.id,
+                node.subscription.subscriptionId,
+                'Azure.BlobContainer',
+                node.treeItem.container.name);
+        }
+    }
+
+    private getLocalFileSize(localPath: string): Promise<number> {
```
EricJizbaMSFT: use 'fs-extra' and there should be an async version of fs.stat

src/azureStorageExplorer/limits.ts
```diff
@@ -0,0 +1,25 @@
+import { storageExplorerLauncher } from "../storageExplorerLauncher/storageExplorerLauncher";
+import { window } from "vscode";
+import { ResourceType } from "../storageExplorerLauncher/ResourceType";
+import { UserCancelledError } from "vscode-azureextensionui";
+
+/*
```
EricJizbaMSFT: nit: move the copyright header above the imports

src/azureStorageExplorer/limits.ts
```diff
@@ -0,0 +1,25 @@
...
+  *  Licensed under the MIT License. See License.txt in the project root for license information.
+  **/
+
+export namespace Limits {
+    //  VS Code currently supports at least 256MB, but not 512MB. But it won't open anything larger than 4MB through the APIs.
+    export const maxUploadDownloadSizeMB = 4;
+    export const maxUploadDownloadSizeBytes = maxUploadDownloadSizeMB * 1000 * 1000;
+
+    export async function askOpenInStorageExplorer(errorMessage: string, resourceId: string, subscriptionId: string, resourceType: ResourceType, resourceName: string): Promise<void> {
+        const message = "Open container in Storage Explorer";
```
EricJizbaMSFT: Is 'container' always accurate? It seems like the resource could be something other than a container

src/azureStorageExplorer/blobContainers/blobContainerNode.ts
```diff
@@ -142,4 +167,94 @@ export class BlobContainerNode implements IAzureParentTreeItem {
...
+            let handler = new BlobFileHandler();
+            await handler.checkCanUpload(node, filePath);
+
+            let blobPath = await vscode.window.showInputBox({
+                prompt: 'Enter a name for the uploaded block blob (may include a path)',
+                value: path.basename(filePath)
+            });
+            if (blobPath) {
+                if (await this.doesBlobExist(blobPath)) {
+                    const result = await vscode.window.showWarningMessage(
```
EricJizbaMSFT: In terms of your question - you can move most of this logic outside createChild, and only call createChild when you know the blob doesn't exist. createChild would basically only need these four lines of code:
```
showCreatingNode(blobPath);
await this.uploadBlockBlob(filePath, blobPath);
const actualBlob = await this.getBlob(blobPath);
return new BlobNode(actualBlob, this.container, this.storageAccount, this.key);
```

For the existing blob case, you could also consider using `tree.findNode(id)` (I think `${node.id}/${blobPath}` would be the id) combined with something like [runWithTemporaryState](https://github.com/Microsoft/vscode-azurefunctions/blob/master/src/tree/FunctionAppTreeItem.ts#L77) to append 'Overwriting...' or 'Uploading...' to the node in the tree

src/azureStorageExplorer/blobContainers/blobFileHandler.ts
```diff
@@ -18,24 +26,79 @@ export class BlobFileHandler implements IRemoteFileHandler<IAzureNode<BlobNode>>
...
             });
         });
     }
 
+    public async downloadFile(node: IAzureNode<BlobNode>, filePath: string): Promise<void> {
+        await this.checkCanDownload(node);
+
+        const blob = node.treeItem.blob;
+        const treeItem = node.treeItem;
+        const linkablePath = Uri.file(filePath); // Allows CTRL+Click in Output panel
```
EricJizbaMSFT: ctrl+click is great - but I suggest we also open the file in VSCode immediately after downloading

EricJizbaMSFT: Although then it's eerily similar to editing the blob, which opens it in VS Code. idk

package.json
StephenWeatherford: ... means that it will bring up an additional dialog/window for more input. This line seems a little blurred in vscode. I'll take a look and see if it's being followed in their own menus.

src/azureServiceExplorer/editors/RemoteFileEditor.ts
StephenWeatherford: Wasn't even aware there was one. Added issue #90.

src/azureStorageExplorer/blobContainers/blobContainerNode.ts
StephenWeatherford: Just looked up common extensions on the web.  They can still manually enter *.* and get to any file they want. The restrictions are about scope more than functionality (long discussion with Jake/Matt). We're just trying to encourage simple dev usage. We do plan on adding images to this list. Not fully happy with this but not sure there's a better solution. I'll discuss with Matt, though, and see what he thinks.

src/azureStorageExplorer/limits.ts
StephenWeatherford: Ah. The auto-import does that if you don't have any other imports yet.  Will fix.

src/azureStorageExplorer/limits.ts
StephenWeatherford: Will fix when I add file share support.  Thanks.

src/azureStorageExplorer/blobContainers/blobFileHandler.ts
StephenWeatherford: Yeah, those were my thoughts, too (did consider this).  I think best to keep the behavior separate and don't open on download.  The only real other difference is the location where it's downloaded.

src/azureStorageExplorer/blobContainers/blobContainerNode.ts
```diff
@@ -142,4 +167,94 @@ export class BlobContainerNode implements IAzureParentTreeItem {
...
+                        'json',
+                        'log',
+                        'md',
+                        'rtf',
+                        'txt',
+                        'text',
+                        'xml',
+                    ]
+                },
+                openLabel: "Select file to upload"
```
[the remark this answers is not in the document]
StephenWeatherford: Done.

src/azureStorageExplorer/blobContainers/blobFileHandler.ts
StephenWeatherford: Done.

src/azureStorageExplorer/blobContainers/blobContainerNode.ts
EricJizbaMSFT: Unless I'm missing something, I don't think there's a way to manually enter on a mac...

src/azureStorageExplorer/blobContainers/blobContainerNode.ts
StephenWeatherford: Separate PR.

PR #104, opened 2018-03-27
Validate new blob doesn't exist, fixes #93

### LINKED ISSUE #93: Creating new blob will silently overwrite existing blob with same name
Repro Stpes:
1. Expand a stoeage account and select one blob container.
2. Create a new blob named 'test'.
3. Create the same blob under the blob container.

Expect:
Create fail in step3 and pop up an error like below.
error

Actual:
The blob can be created successfully.
test

More info:
1. This issue doesn't reproduce for directory/table.
2. This issue also reproduces for file/queue.

StephenWeatherford:


src/azureStorageExplorer/blobContainers/blobContainerNode.ts
```diff
@@ -202,7 +202,16 @@ export class BlobContainerNode implements IAzureParentTreeItem {
     private async createChildAsNewBlockBlob(showCreatingNode: (label: string) => void): Promise<IAzureTreeItem> {
         const blobName = await vscode.window.showInputBox({
             placeHolder: 'Enter a name for the new block blob',
-            validateInput: BlobContainerNode.validateBlobName
+            validateInput: async (name: string) => {
+                let nameError = BlobContainerNode.validateBlobName(name);
```
EricJizbaMSFT: So I was curious if you already checked `this.doesBlobExist` for the upload scenario and it turns out you do (as a warning message instead of validateInput). However, that made me think: should you add `validateInput: BlobContainerNode.validateBlobName` when getting the blob path for the upload case?

StephenWeatherford: Good point, fixed.

PR #109, opened 2018-03-28
Copy all content settings, fixes #20

### LINKED ISSUE #20: Saving text file back to Azure loses properties and metadata

StephenWeatherford:


src/azureStorageExplorer/fileShares/fileFileHandler.ts
```diff
@@ -36,7 +36,8 @@ export class FileFileHandler implements IRemoteFileHandler<IAzureNode<FileNode>>
         let createOptions: azureStorage.FileService.CreateFileRequestOptions = {};
 
         if (fileProperties && fileProperties.contentSettings && fileProperties.contentSettings.contentType) {
```
EricJizbaMSFT: nit: No need to check `fileProperties.contentSettings.contentType` in the if case anymore

StephenWeatherford: Thanks, that's not a nit, it's a bug (it might be empty but others non-empty).  Handled it for blob but missed it for files.

EricJizbaMSFT: Oh true - didn't consider that

Issue #116, opened 2018-04-11
Suggestion: Ability to browse blob container (Copy URL)


chrisdias:
I've got a static site deployed to a blob storage container and i would like to be able to right click on the container and browse to it. 

Today there is no context menu to do this:

![image](https://user-images.githubusercontent.com/1487073/38595327-0474b4ec-3d00-11e8-92d3-dbaa0400b1b1.png)

StephenWeatherford:
Are you asking for a way to copy the URI to the clipboard? Depending on settings, you may or may not be able to actually browse to it (if it's public and publicly enumerable).

StephenWeatherford:
@chrisdias I'm assuming you meant a "Get URL" context menu item and am implementing it that way.

chrisdias:
@StephenWeatherford yes, that would be sufficient.

StephenWeatherford:
Fixed via #129

PR #129, opened 2018-04-25
Copy URL (files, shares, blobs, containers)

StephenWeatherford:


StephenWeatherford:
Fixes #116

src/azureStorageExplorer/fileShares/fileShareNode.ts
```diff
@@ -50,6 +54,14 @@ export class FileShareNode implements IAzureParentTreeItem {
             }));
     }
 
+    public async copyUrl(_node: IAzureParentNode): Promise<void> {
+        let fileService = azureStorage.createFileService(this.storageAccount.name, this.key.value);
+        let url = fileService.getUrl(this.share.name, "");
+        copypaste.copy(url);
+        azureStorageOutputChannel.show();
+        azureStorageOutputChannel.appendLine(`Share URL copied to clipboard: ${url}`);
+    }
```
PrashanthCorp: I notice that in all the use cases, the only difference is line 1, in the method being called. 
Is there an easy way to remove the duplicated code (getUrl to appendLine)? 

I think having a createService method (with `FileShareNode` calling `createFileService`, etc.) would free up the other lines to a common method call.

StephenWeatherford: I considered it but decided it wouldn't be worth the gain, given that the logic is dead simple.

StephenWeatherford: ping

Issue #153, opened 2018-06-05
Enable static site hosting from extension


fiveisprime:
It seems like we'll end up shipping this in two parts, 1) supporting deployments and 2) supporting enabling and configuring storage accounts for web hosting.

As part of this first step, it would be helpful to be able to launch into the portal directly to a storage account to simplify the process for enabling web hosting.

Related to #146

fiveisprime:
URL format seems to elude to being able to deep link.

/resource/subscriptions/:id/resourcegroups/:group/providers/Microsoft.Storage/storageAccounts/:name/staticWebsite

StephenWeatherford:
Right now we're sending them to the portal, can allow enabling ourselves when we get SDK support.

PrashanthCorp:
The above PR should close this issue since deployment has been implemented [here](https://github.com/Microsoft/vscode-azurestorage/commit/4e729dfd304e2485a4e0a84d6c65ee9dd1330086). @fiveisprime  let me know if there's more to the issue you'd like aside from the PR. :)

PrashanthCorp:
Closing via PR.

PR #158, opened 2018-06-11
Configure static website...

StephenWeatherford:
Part of #146, #153 

Menu to get to the portal's Static Website tab.  Since the preview is not yet public, we have to temporarily add a flag to enable the feature in the portal.

Depends on https://github.com/Microsoft/vscode-azuretools/pull/194

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,6 +66,10 @@
 				"title": "Open in Portal"
 			},
 			{
+				"command": "azureStorage.configureStaticWebsite",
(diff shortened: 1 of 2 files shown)


StephenWeatherford:
This will ship with next update.  I should be able to remove the flag when the preview goes public soon.

StephenWeatherford:
Turns out this will be under a feature flag.

PR #163, opened 2018-06-15
Fix reversed condition

StephenWeatherford:
Accidentally checked in.


src/azureStorageExplorer/storageAccounts/storageAccountActionHandlers.ts
```diff
@@ -102,7 +102,7 @@ async function deployStaticWebsite(this: IActionContext, target?: vscode.Uri | I
             destContainerNode = enabledContainers[0];
         }
     }
-    console.assert(!destContainerNode, "Should have a destination container node by now");
+    console.assert(!!destContainerNode, "Should have a destination container node by now");
```
PrashanthCorp: Is the 'by now' phrase necessary? If yes, I'd prefer 'Should have a destination container node at this point'

PR #234, opened 2018-09-28
Fix missing activation events, add tests

### LINKED ISSUE #232: Commands need to be listed as activationEvents
Repro steps:
1. Open storage explorer view in vs code
1. Hit the refresh button really quickly

Expected:
Extension activates before the refresh event occurs

Actual:
"command 'azureStorage.refresh' not found"

The commands need to be listed in the activationEvents here: https://github.com/Microsoft/vscode-azurestorage/blob/master/package.json#L31

StephenWeatherford:
Fixes #232


package.json
```diff
@@ -31,7 +31,9 @@
 	"activationEvents": [
 		"onView:azureStorage",
 		"onCommand:azureStorage.deployStaticWebsite",
-		"onCommand:azureStorage.configureStaticWebsite"
+		"onCommand:azureStorage.configureStaticWebsite",
```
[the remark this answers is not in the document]
EricJizbaMSFT: I say we force every command to be in activationEvents. It's really not that hard to do and it could prevent future headaches.

In fact, I would suggest we check the result of `vscode.commands.getCommands()` instead of parsing package.json for the tests. Any command we register can technically be called by another extension, even if its not in package.json (which is more important when we actually "support" apis for partner teams). I believe users can also set up key bindings to specific commands:
screen shot 2018-09-28 at 1 18 05 pm

package.json
```diff
@@ -2,7 +2,7 @@
 	"name": "vscode-azurestorage",
 	"displayName": "Azure Storage",
 	"description": "Manage your Azure Storage accounts including Blob Containers, File Shares, Tables and Queues",
-	"version": "0.4.1",
+	"version": "0.4.2",
```
EricJizbaMSFT: Does this mean you want to do a hot fix for this? The bug has existed for almost a year and no one has complained...  I'm not sure how urgent it is

CHANGELOG.md
```diff
@@ -1,6 +1,11 @@
 # Change Log
 All notable changes to the "vscode-azurestorage" extension will be documented in this file.
 
+## 0.4.2 - 2018-09-28
+
+### Fixed
+- Can get a command not found error attempting to run some commands before extension is activatd [#232](https://github.com/Microsoft/vscode-azurestorage/issues/232)
```
EricJizbaMSFT: "activatd"

PR #235, opened 2018-09-29
Fix missing activation events (#234)

StephenWeatherford:
This is the release branch for 0.4.2, just porting changes from fix


StephenWeatherford:
#234

Issue #141, opened 2018-05-10
[Linux]'Copy Connection String' doesn't work


v-xuanzh:
Repro Steps:
1. Right click one storage account and select 'Copy Connection String'.
2. Paste the connection string to notepad or Storage Explorer.

Expect:
The connection string of the account can pasted successfully.

Actual:
The connection string of the account isn't pasted. 

More info:
'Copy Primary Key' also doesn't work.

PrashanthCorp:
For the repro, has xclip been installed?

EricJizbaMSFT:
That doesn't matter. You can use clipboardy like in the [functions extension](https://github.com/Microsoft/vscode-azurefunctions/blob/34e03574739cd19ce3ade422d560b97cfd4d96bf/src/commands/copyFunctionUrl.ts#L17), which will work without xclip

PR #256, opened 2018-12-06
copy-paste -> clipboardy

PrashanthCorp:
Note: The npm install here also updated a few other packages.  

Fixes https://github.com/Microsoft/vscode-azurestorage/issues/141

### DIFF
--- a/package.json
+++ b/package.json
@@ -662,6 +662,7 @@
 		"all": "npm i && npm run lint && npm test"
 	},
 	"devDependencies": {
+		"@types/clipboardy": "^1.1.0",
 		"@types/fs-extra": "=5.0.2",
 		"@types/glob": "^5.0.35",
 		"@types/mocha": "^5.2.5",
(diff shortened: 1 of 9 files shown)


thirdpartynotices.txt
```diff
@@ -7,9 +7,8 @@ are grateful to these developers for their contribution to open source.
 
 1. node-winreg (https://github.com/fresc81/node-winreg)
 2. fs-extra (https://github.com/jprichardson/node-fs-extra)
-3. copy-paste (https://github.com/xavi-/node-copy-paste)
-4. glob (https://github.com/isaacs/node-glob)
-5. opn (https://github.com/sindresorhus/opn)
+3. glob (https://github.com/isaacs/node-glob)
```
StephenWeatherford: You need to add clipboardy.

PrashanthCorp: Looks like the opn license was also missing. 

Adding that too.

package.json
```diff
@@ -685,10 +684,11 @@
 		"ms-vscode.azure-account"
 	],
 	"dependencies": {
+		"@types/clipboardy": "^1.1.0",
```
EricJizbaMSFT: This should be a dev dependency

Issue #216, opened 2018-08-07
Implement pre-publish action for Static Websites


gabrielvdk:
Would be great if you can do some pre-processing before the files go to Azure.

Eg.
- Configure some settings based on target storage account
- Increase a build number / date-time
- etc.

StephenWeatherford:
@fiveisprime Do we do something like this for publishing in any of the other extensions?

fiveisprime:
We don't; however it makes sense to be able to configure some kind of pre-publish build for SPAs. Maybe we could hook into VS Code's build system to trigger the command that you could customie how you see fit?

EricJizbaMSFT:
In Functions, we run `dotnet publish` before deploying a C# project. VS Code has a tasks api that you can leverage to list and execute tasks from an extension. See here for more info: https://github.com/Microsoft/vscode-azurefunctions/blob/master/src/commands/deploy.ts#L201

I'm not sure what the user experience would be in the storage extension, but I'm guessing you could leverage the same tasks api

StephenWeatherford:
I like that idea.  prepublishStaticWebsite?

PR #268, opened 2018-12-10
Add preDeployTask setting

EricJizbaMSFT:
Fixes https://github.com/Microsoft/vscode-azurestorage/issues/216

This was another piece of low-hanging fruit I thought I'd stick in the next release. Basically copied all of this logic from functions/app service which already support a `preDeployTask`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
 	"publisher": "ms-azuretools",
 	"aiKey": "AIF-d9b70cd4-b9f9-4d70-929b-a071c400b217",
 	"engines": {
-		"vscode": "^1.23.0"
+		"vscode": "^1.24.0"
 	},
 	"repository": {
 		"type": "git",
@@ -647,6 +647,11 @@
 					"type": "boolean",
 					"default": true,
 					"description": "Show warning dialog on Blob file save."
+				},
+				"azureStorage.preDeployTask": {
+					"scope": "resource",
+					"type": "string",
+					"description": "The name of the task to run before deploying."
 				}
 			}
 		}
--- a/src/azureStorageExplorer/storageAccounts/storageAccountActionHandlers.ts
+++ b/src/azureStorageExplorer/storageAccounts/storageAccountActionHandlers.ts
@@ -7,8 +7,7 @@ import * as clipboardy from 'clipboardy';
 import * as vscode from "vscode";
 import { IActionContext, registerCommand, TelemetryProperties } from 'vscode-azureextensionui';
 import { isPathEqual, isSubpath } from '../../components/fs';
-import { extensionPrefix } from '../../constants';
-import * as ext from "../../constants";
+import { configurationSettingsKeys, extensionPrefix } from '../../constants';
 import { storageExplorerLauncher } from '../../storageExplorerLauncher/storageExplorerLauncher';
 import { BlobContainerTreeItem } from "../blobContainers/blobContainerNode";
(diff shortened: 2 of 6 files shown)


src/azureStorageExplorer/storageAccounts/storageAccountActionHandlers.ts
```diff
@@ -78,5 +80,42 @@ async function deployStaticWebsite(this: IActionContext, target?: vscode.Uri | S
         throw new Error(`Could not find $web blob container for storage account "${destAccountTreeItem.label}"`);
     }
 
+    await runPreDeployTask(sourcePath, properties);
+
     return destContainerTreeItem.deployStaticWebsite(this, sourcePath);
 }
+
+async function runPreDeployTask(deployFsPath: string, telemetryProperties: TelemetryProperties): Promise<void> {
+    const preDeployTaskKey: string = 'preDeployTask';
```
StephenWeatherford: -> constants.configurationSettingsKeys

package.json
```diff
@@ -647,6 +647,11 @@
 					"type": "boolean",
 					"default": true,
 					"description": "Show warning dialog on Blob file save."
+				},
+				"azureStorage.preDeployTask": {
```
StephenWeatherford: Shouldn't we have a default?

src/azureStorageExplorer/storageAccounts/storageAccountActionHandlers.ts
```diff
@@ -78,5 +80,42 @@ async function deployStaticWebsite(this: IActionContext, target?: vscode.Uri | S
         throw new Error(`Could not find $web blob container for storage account "${destAccountTreeItem.label}"`);
     }
 
+    await runPreDeployTask(sourcePath, properties);
```
StephenWeatherford: Have your users been finding the predeploy option easily?  Or do we need to do some to make it more discoverable, such as indicating in the output window that there's no predeploy task, or even asking if they want to set one up?

EricJizbaMSFT: Well in functions we create the project for them so we set this. In App Service we haven't had any issues filed since adding the setting but idk. Is this common for static web sites? Seems like a very project-specific thing

package.json
EricJizbaMSFT: Correct. The task could be anything they defined and is likely project/runtime specific

src/azureStorageExplorer/storageAccounts/storageAccountActionHandlers.ts
EricJizbaMSFT: What's the reasoning for that? I usually only put stuff in constants if it's used in multiple files, but no strong preference...

StephenWeatherford: I guess I default to more consistency, but also not strong preference.  Up to you.

src/azureStorageExplorer/storageAccounts/storageAccountActionHandlers.ts
StephenWeatherford: @fiveisprime Any thoughts on whether we need to do work to make this more discoverable?

EricJizbaMSFT: I'm not gonna wait on this question to merge. Seems like we could always make it more discoverable after the fact

PR #275, opened 2018-12-15
Allow user to configure static website properties

PrashanthCorp:
Update icon wrt a) blob is $web, b) account is hosting static website

I redid part of the PR of saw/configure. A cherry pick was not possible due to the azure tree refactoring done. 

Addresses #153


PrashanthCorp:
Here's a short gif how it now looks: 

conf_stat_web

PrashanthCorp:
The [previous commit](https://github.com/Microsoft/vscode-azurestorage/pull/275/commits/ecec1865d4d69669f398d934b55459df70b748e0) has some changes I made to comply with strict-boolean-expression. However I don't think the code is more readable, so any thoughts on making the code better would be appreciated.

src/azureStorageExplorer/storageAccounts/storageAccountNode.ts
```diff
@@ -34,14 +36,36 @@ export type WebsiteHostingStatus = {
...
 
     private readonly _blobContainerGroupTreeItem: BlobContainerGroupTreeItem;
     private readonly _fileShareGroupTreeItem: FileShareGroupTreeItem;
     private readonly _queueGroupTreeItem: QueueGroupTreeItem;
     private readonly _tableGroupTreeItem: TableGroupTreeItem;
     private _root: IStorageRoot;
+    private _websiteHostingEnabled: boolean;
+
+    // Call this before giving this node to vscode
+    public set websiteHostingEnabled(value: boolean) {
```
EricJizbaMSFT: If you put this logic inside `createStorageAccountTreeItem` then it forces people to call it before creating the node and you won't have to do any asserts

src/azureStorageExplorer/storageAccounts/storageAccountNode.ts
```diff
@@ -249,11 +294,45 @@ export class StorageAccountTreeItem extends AzureParentTreeItem<IStorageRoot> {
     }
 
     public async configureStaticWebsite(): Promise<void> {
-        let hostingStatus = await this.getWebsiteHostingStatus();
-        await this.ensureHostingCapable(hostingStatus);
+        const defaultIndexDocumentName = 'index.html';
+        let oldStatus = await this.getWebsiteHostingStatus();
+        await this.ensureHostingCapable(oldStatus);
+        let indexDocument = await ext.ui.showInputBox({
+            ignoreFocusOut: true,
```
EricJizbaMSFT: nit: I believe ext.ui.showInputBox already defaults `ignoreFocusOut` to true, so this shouldn't be necessary

src/azureStorageExplorer/storageAccounts/storageAccountNode.ts
```diff
@@ -249,11 +294,45 @@ export class StorageAccountTreeItem extends AzureParentTreeItem<IStorageRoot> {
...
+            prompt: "Enter the index document name",
+            value: oldStatus.indexDocument ? oldStatus.indexDocument : defaultIndexDocumentName
+        });
+
+        let errorDocument404Path: string = await ext.ui.showInputBox({
+            ignoreFocusOut: true,
+            prompt: "Enter the 404 document path",
```
EricJizbaMSFT: nit: How about "Enter the error document path" instead of "Enter the 404 document path"

src/azureStorageExplorer/storageAccounts/storageAccountNode.ts
```diff
@@ -249,11 +294,45 @@ export class StorageAccountTreeItem extends AzureParentTreeItem<IStorageRoot> {
...
+            value: oldStatus.errorDocument404Path ? oldStatus.errorDocument404Path : "",
+            placeHolder: 'e.g. error/documents/error.html',
```
EricJizbaMSFT: How'd you pick this placeHolder? The docs that I found used "404.html"
https://docs.microsoft.com/en-us/azure/storage/blobs/storage-blob-static-website
https://azure.microsoft.com/en-us/blog/azure-storage-static-web-hosting-public-preview/

Also I like it better as the default value (like you did for "index.html") instead of placeHolder

src/azureStorageExplorer/storageAccounts/storageAccountNode.ts
```diff
@@ -249,11 +294,45 @@ export class StorageAccountTreeItem extends AzureParentTreeItem<IStorageRoot> {
...
+            validateInput: (value: string): string | undefined => {
+                if (value) {
+                    if (value.startsWith('/') || value.endsWith('/')) {
+                        return "If specified, the error document path must not begin or end with a '/' character.";
```
EricJizbaMSFT: nit: I don't think "If specified" is necessary. We don't do that for other optional input boxes

src/azureStorageExplorer/storageAccounts/storageAccountNode.ts
```diff
@@ -249,11 +294,45 @@ export class StorageAccountTreeItem extends AzureParentTreeItem<IStorageRoot> {
...
     public async configureStaticWebsite(): Promise<void> {
-        let hostingStatus = await this.getWebsiteHostingStatus();
-        await this.ensureHostingCapable(hostingStatus);
+        const defaultIndexDocumentName = 'index.html';
+        let oldStatus = await this.getWebsiteHostingStatus();
+        await this.ensureHostingCapable(oldStatus);
+        let indexDocument = await ext.ui.showInputBox({
+            ignoreFocusOut: true,
+            prompt: "Enter the index document name",
+            value: oldStatus.indexDocument ? oldStatus.indexDocument : defaultIndexDocumentName
```
EricJizbaMSFT: validateInput for this?

src/azureStorageExplorer/storageAccounts/storageAccountNode.ts
```diff
@@ -249,11 +294,45 @@ export class StorageAccountTreeItem extends AzureParentTreeItem<IStorageRoot> {
...
+            ignoreFocusOut: true,
+            prompt: "Enter the 404 document path",
+            value: oldStatus.errorDocument404Path ? oldStatus.errorDocument404Path : "",
+            placeHolder: 'e.g. error/documents/error.html',
+            validateInput: (value: string): string | undefined => {
+                if (value) {
+                    if (value.startsWith('/') || value.endsWith('/')) {
+                        return "If specified, the error document path must not begin or end with a '/' character.";
+                    } else if (value.length < 3 || value.length > 255) {
+                        return "If specified, the error document path must be between 3 and 255 characters in length";
```
EricJizbaMSFT: I prefer to make a const for the min and max values to make sure that the if condition and message are using the same numbers (especially if they're changed in the future)

src/azureStorageExplorer/storageAccounts/storageAccountNode.ts
```diff
@@ -249,11 +294,45 @@ export class StorageAccountTreeItem extends AzureParentTreeItem<IStorageRoot> {
...
+                return undefined;
+            }
+        });
+        let newStatus: azureStorage.common.models.ServicePropertiesResult.StaticWebsiteProperties = {
+            Enabled: true,
+            ErrorDocument404Path: errorDocument404Path,
+            IndexDocument: indexDocument
+        };
+        await this.setWebsiteHostingProperties(newStatus);
+        let msg = oldStatus.enabled ?
```
EricJizbaMSFT: Is there an open issue to allow users to disable static website hosting? Doesn't seem high priority to me, but usually if you support "enable" you should also support "disable"

src/azureStorageExplorer/storageAccounts/storageAccountNode.ts
```diff
@@ -249,11 +294,45 @@ export class StorageAccountTreeItem extends AzureParentTreeItem<IStorageRoot> {
...
+            'Static website hosting configuration updated' :
+            'The storage account has been enabled for static website hosting';
```
EricJizbaMSFT: Missing periods in several of your messages

src/azureStorageExplorer/storageAccounts/storageAccountNode.ts
```diff
@@ -249,11 +294,45 @@ export class StorageAccountTreeItem extends AzureParentTreeItem<IStorageRoot> {
...
+        window.showInformationMessage(msg);
+        if (!oldStatus.enabled) {
+            await ext.tree.refresh();
```
EricJizbaMSFT: It can be pretty time-consuming to refresh the entire tree. Only refresh the node you need to refresh

src/azureStorageExplorer/storageAccounts/storageAccountNode.ts
```diff
@@ -234,6 +255,30 @@ export class StorageAccountTreeItem extends AzureParentTreeItem<IStorageRoot> {
...
     }
 
+    public async setWebsiteHostingProperties(staticWebsiteProperties: azureStorage.common.models.ServicePropertiesResult.StaticWebsiteProperties): Promise<WebsiteHostingStatus> {
+        let blobService = this.root.createBlobService();
+        return await new Promise<WebsiteHostingStatus>((resolve, reject) => {
+            blobService.getServiceProperties((err: Error | undefined, props: azureStorage.common.models.ServicePropertiesResult.BlobServiceProperties = {}) => {
+                if (err) {
+                    reject(err);
+                } else {
+                    props.StaticWebsite = {
```
EricJizbaMSFT: Why not just `props.StaticWebsite = staticWebsiteProperties`?

src/azureStorageExplorer/blobContainers/blobContainerNode.ts
```diff
@@ -32,13 +33,29 @@ interface ICreateChildOptions {
     blobPath: string;
 }
 
+const defaultIconPath: { light: string | Uri; dark: string | Uri } = {
+    light: path.join(__filename, '..', '..', '..', '..', '..', 'resources', 'light', 'AzureStorageAccount.svg'),
+    dark: path.join(__filename, '..', '..', '..', '..', '..', 'resources', 'dark', 'AzureStorageAccount.svg')
+};
+
+const websiteIconPath: { light: string | Uri; dark: string | Uri } = {
+    light: path.join(__filename, '..', '..', '..', '..', '..', 'resources', 'light', 'Website.svg'),
```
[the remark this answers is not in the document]
PrashanthCorp: IMO, the web app icon is more indicative of a website than the portal icon.

src/azureStorageExplorer/storageAccounts/storageAccountNode.ts
PrashanthCorp: staticWebsiteProperties has fields populated through the input box. The default false-y value there is an empty string rather than undefined. I can handle this conversion ("" -> undefined) at the call site, configureStaticWebsite, but the logic there would be too bloated. The separation of logic (collecting input vs pruning + API call) made sense to me.

src/azureStorageExplorer/storageAccounts/storageAccountNode.ts
PrashanthCorp: On the portal, I can set the index document to be a file not ending in `.htm` or `.html`. Given the lack of documentation,  I'd prefer not enforcing a check vs being more restrictive/blocking.

src/azureStorageExplorer/storageAccounts/storageAccountNode.ts
PrashanthCorp: From my understanding, we don't use periods in single-sentence prompts. I could be wrong however.

src/azureStorageExplorer/storageAccounts/storageAccountNode.ts
EricJizbaMSFT: Yeah I wouldn't validate the file extension. I feel like there are quite a few options. I would just do the same validation that the portal does:
screen shot 2019-01-04 at 11 49 47 am

src/azureStorageExplorer/storageAccounts/storageAccountNode.ts
EricJizbaMSFT: In general you want to modify the same object you retrieved from Azure. For example if Azure adds another property in the future ("errorDocument401Path" maybe?) you might be accidentally clearing out that value since you're creating a "new" object with just your three known properties

As a side note you're calling `getServiceProperties` within both `getWebsiteHostingStatus` and `setWebsiteHostingProperties` - I would actually prefer if that was only called once (less duplicate code and less calls to Azure)

EricJizbaMSFT: Also what happens if you just set it to an empty string? It doesn't work?

src/azureStorageExplorer/storageAccounts/storageAccountNode.ts
PrashanthCorp: Scratch my above comment.

src/azureStorageExplorer/storageAccounts/storageAccountNode.ts
PrashanthCorp: From what we've discussed offline, @fiveisprime disagrees - he feels this should be left blank.

src/azureStorageExplorer/storageAccounts/storageAccountNode.ts
PrashanthCorp: The portal's tooltip mentions 404 specifically: 
error404documentpath

EricJizbaMSFT: Yeah I'm fine mentioning 404 - I guess I just wanted to make sure "error" was in there which you already changed

src/azureStorageExplorer/storageAccounts/storageAccountNode.ts
```diff
@@ -249,11 +292,49 @@ export class StorageAccountTreeItem extends AzureParentTreeItem<IStorageRoot> {
...
-        let resourceId = `${this.fullId}/staticWebsite`;
-        this.openInPortal(resourceId);
+    }
+
+    private validateDocumentNameInput(documentType: string, documentpath: string | undefined): undefined | string {
+        const minLengthDocumentPath = 3;
+        const maxLengthDocumentPath = 255;
+        if (documentpath) {
+            if (documentpath.startsWith('/') || documentpath.endsWith('/')) {
+                return `The ${documentType} document path must not begin or end with a '/' character.`;
```
EricJizbaMSFT: Not entirely accurate for the index page. It can't have _any_ '/' character

src/azureStorageExplorer/storageAccounts/storageAccountNode.ts
```diff
@@ -249,11 +292,49 @@ export class StorageAccountTreeItem extends AzureParentTreeItem<IStorageRoot> {
...
+            ErrorDocument404Path: errorDocument404Path,
+            IndexDocument: indexDocument
+        };
+        await this.setWebsiteHostingProperties(newStatus);
+        let msg = oldStatus.enabled ?
+            'Static website hosting configuration updated.' :
+            'The storage account has been enabled for static website hosting.';
+        window.showInformationMessage(msg);
+        if (!oldStatus.enabled) {
+            await ext.tree.refresh(this);
```
EricJizbaMSFT: nit: I think you can just do `this.refresh();`

src/azureStorageExplorer/storageAccounts/storageAccountNode.ts
```diff
@@ -70,10 +92,6 @@ export class StorageAccountTreeItem extends AzureParentTreeItem<IStorageRoot> {
     public label: string = this.storageAccount.name;
     public static contextValue: string = 'azureStorageAccount';
     public contextValue: string = StorageAccountTreeItem.contextValue;
-    public iconPath: { light: string | Uri; dark: string | Uri } = {
```
EricJizbaMSFT: I guess I would prefer if you just fully reverted your changes to the iconPath in this file. For example, you're setting iconPath to `defaultIconPath` twice (Once in the constructor and once in the property definition) which is kinda confusing

src/azureStorageExplorer/storageAccounts/storageAccountNode.ts
```diff
@@ -53,12 +71,16 @@ export class StorageAccountTreeItem extends AzureParentTreeItem<IStorageRoot> {
...
         this._tableGroupTreeItem = new TableGroupTreeItem(this);
+        this.iconPath = defaultIconPath;
     }
 
     public static async createStorageAccountTreeItem(parent: AzureParentTreeItem, storageAccount: StorageAccountWrapper, client: StorageManagementClient): Promise<StorageAccountTreeItem> {
         const ti = new StorageAccountTreeItem(parent, storageAccount, client);
         // make sure key is initialized
         await ti.refreshImpl();
+        const hostingStatus = await ti.getWebsiteHostingStatus();
+        ti.websiteHostingEnabled = hostingStatus.enabled;
```
EricJizbaMSFT: Is all this `websiteHostingEnabled` still necessary in `StorageAccountTreeItem` since you're not changing the icon anymore?

PrashanthCorp: We also use it to error-check 'Browse Static Website' and 'Configure Static Website'.

package.json
```diff
@@ -368,6 +379,11 @@
 					"when": "view == azureStorage && viewItem == azureBlobContainer",
 					"group": "6_staticwebsites@2"
 				},
+				{
+					"command": "azureStorage.disableStaticWebsite",
+					"when": "view == azureStorage && viewItem == azureBlobContainer",
+					"group": "6_staticwebsites@2"
```
EricJizbaMSFT: Again - the number after the @ sign is the order within the static website group of the context menu. The ordering here should be the same as the ordering on the account level (Not entirely your fault - looks like the orders weren't the same to begin with)

src/azureStorageExplorer/WebsiteHostingStatus.ts
```diff
@@ -0,0 +1,10 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+export type WebsiteHostingStatus = {
```
EricJizbaMSFT: Where is this used? If nowhere - please delete

src/azureStorageExplorer/selectStorageAccountNodeForCommand.ts
```diff
@@ -51,14 +51,14 @@ export async function selectStorageAccountTreeItemForCommand(
 
         if (options.askToConfigureWebsite && !hostingStatus.enabled) {
             let result = await window.showInformationMessage(
-                `Website hosting is not enabled on storage account "${accountTreeItem.label}". Would you like to go to the portal to enable it?`,
+                `Website hosting is not enabled on storage account "${accountTreeItem.label}". Would you like to enable it?`,
```
EricJizbaMSFT: A few things about this dialog:
1. This should be modal. This dialog blocks users from the action they literally just started, so it should be in their face for them to solve right away. Also if it's modal, then I can just hit "enter" to select the default button. If it's not modal, I have to use my mouse to click a button.
1. This should be a warning message not an information message. Right? Doesn't this seems like a warning? It's stopping the action they thought they could perform...
1. I'm not a huge fan of "yes" or "no" buttons. It requires me to parse the question and the button separately. Instead, I would prefer a message that was just `Website hosting is not enabled on storage account "${accountTreeItem.label}". ` with "Enable" and "Cancel" buttons

I know all of this wasn't your code, but since you were changing the message anyways I noticed it... plus it is integral to the experience you're changing.

src/azureStorageExplorer/storageAccounts/storageAccountNode.ts
EricJizbaMSFT: This is closer to what you had before, but I still want it to be exactly what you had before. I prefer not to make unnecessary refactoring changes as part of a larger feature because it can lead to hard-to-find bugs. I don't think you have a bug with this code, but as an example I think you _do_ have a bug in `blobContainerNode.ts` where the container node now looks like the storage account node:
screen shot 2019-01-11 at 10 26 20 am

instead of:
screen shot 2019-01-11 at 10 29 01 am

src/azureStorageExplorer/blobContainers/blobContainerNode.ts
```diff
@@ -32,13 +33,29 @@ interface ICreateChildOptions {
...
 export class BlobContainerTreeItem extends AzureParentTreeItem<IStorageRoot> implements ICopyUrl {
     private _continuationToken: azureStorage.common.ContinuationToken | undefined;
 
     constructor(
         parent: AzureParentTreeItem,
         public readonly container: azureStorage.BlobService.ContainerResult) {
         super(parent);
+        const storageAccountNode = <StorageAccountTreeItem>(parent.parent);
+        const websiteHostingEnabled = storageAccountNode.websiteHostingEnabled;
+        this.iconPath =
```
EricJizbaMSFT: If you set the iconPath in the constructor then it will never be updated when you refresh this node (aka when you call `this.refresh()`). You have to set it as a property. This is how I would do it:
```typescript
public get iconPath(): { light: string | Uri; dark: string | Uri } {
    const iconName: string = this._websiteHostingEnabled ? 'BrandAzureStaticWebsites.svg' : 'AzureBlobContainer.svg';
    return {
        light: path.join(__filename, '..', '..', '..', '..', '..', 'resources', 'light', iconName),
        dark: path.join(__filename, '..', '..', '..', '..', '..', 'resources', 'dark', iconName)
    };
}
```

Of course you'd have to define `_websiteHostingEnabled` in this file and make sure it's updated on a refresh. `AzureTreeItem` has a method `refreshImpl` that you can implement to "execute any async code when this node is refreshed". Something like this:
```typescript
public async refreshImpl(): Promise<void> {
    const hostingStatus = await this.parent.parent.getWebsiteHostingStatus();
}
```

src/azureStorageExplorer/storageAccounts/storageAccountNode.ts
```diff
@@ -249,11 +289,54 @@ export class StorageAccountTreeItem extends AzureParentTreeItem<IStorageRoot> {
...
+        });
+        let newStatus: azureStorage.common.models.ServicePropertiesResult.StaticWebsiteProperties = {
+            Enabled: true,
+            ErrorDocument404Path: errorDocument404Path,
+            IndexDocument: indexDocument
+        };
+        await this.setWebsiteHostingProperties(newStatus);
+        let msg = oldStatus.enabled ?
+            'Static website hosting configuration updated.' :
+            `The storage account ${this.label} has been enabled for static website hosting.`;
```
EricJizbaMSFT: Wrap `${this.label}` in quotes

src/azureStorageExplorer/storageAccounts/storageAccountNode.ts
```diff
@@ -249,11 +289,54 @@ export class StorageAccountTreeItem extends AzureParentTreeItem<IStorageRoot> {
...
+        window.showInformationMessage(msg);
+        if (!oldStatus.enabled) {
+            await this.refresh();
+        }
+
+    }
+
+    public async disableStaticWebsite(): Promise<void> {
+        await this.setWebsiteHostingProperties({ Enabled: false, ErrorDocument404Path: undefined, IndexDocument: undefined });
```
EricJizbaMSFT: I think the only thing you should change is "Enabled". The rest of the properties should probably stay the same so they don't have to re-enter them if they decide to re-enable

src/azureStorageExplorer/selectStorageAccountNodeForCommand.ts
```diff
@@ -46,19 +46,20 @@ export async function selectStorageAccountTreeItemForCommand(
...
-            let result = await window.showInformationMessage(
-                `Website hosting is not enabled on storage account "${accountTreeItem.label}". Would you like to go to the portal to enable it?`,
-                DialogResponses.yes,
-                DialogResponses.no);
+            let result = await window.showWarningMessage(
+                `Website hosting is not enabled on storage account "${accountTreeItem.label}".`,
+                { modal: true },
+                { title: "Enable web hosting." },
+                DialogResponses.cancel);
             let enableResponse = (result === DialogResponses.yes);
```
EricJizbaMSFT: You have to fix this line since you're no longer using `DialogResponses.yes`

src/azureStorageExplorer/selectStorageAccountNodeForCommand.ts
```diff
@@ -46,19 +46,20 @@ export async function selectStorageAccountTreeItemForCommand(
...
 
         if (options.askToConfigureWebsite && !hostingStatus.enabled) {
-            let result = await window.showInformationMessage(
-                `Website hosting is not enabled on storage account "${accountTreeItem.label}". Would you like to go to the portal to enable it?`,
-                DialogResponses.yes,
-                DialogResponses.no);
+            let result = await window.showWarningMessage(
+                `Website hosting is not enabled on storage account "${accountTreeItem.label}".`,
+                { modal: true },
+                { title: "Enable web hosting." },
```
EricJizbaMSFT: Your messages throughout this PR are inconsistent on two cases:
1. website hosting vs web hosting
1. account vs storage account

I think it should be "website hosting" and "storage account"

(NOTE: In this case if "Enable website hosting" is too long for the button, I'm okay to use just "Enable")

package.json
```diff
@@ -361,12 +372,17 @@
...
-					"group": "6_staticwebsites@2"
+					"group": "6_staticwebsites@3"
+				},
+				{
+					"command": "azureStorage.disableStaticWebsite",
+					"when": "view == azureStorage && viewItem == azureBlobContainer",
+					"group": "6_staticwebsites@4"
 				},
 				{
 					"command": "azureStorage.browseStaticWebsite",
```
EricJizbaMSFT: This browse needs to be `6_staticwebsites@1`

src/azureStorageExplorer/storageAccounts/storageAccountNode.ts
```diff
@@ -249,19 +284,84 @@ export class StorageAccountTreeItem extends AzureParentTreeItem<IStorageRoot> {
...
+
+        return undefined;
+    }
+
+    private validateErrorDocumentName(documentpath: string): undefined | string {
+        const minLengthDocumentPath = 3;
+        const maxLengthDocumentPath = 255;
+        if (documentpath) {
+            if (documentpath.startsWith('/') || documentpath.endsWith('/')) {
+                return "The error document path cannot contain a '/' character.";
```
EricJizbaMSFT: `The error document path must not begin or end with a '/' character.`

src/azureStorageExplorer/storageAccounts/storageAccountNode.ts
```diff
@@ -249,19 +284,84 @@ export class StorageAccountTreeItem extends AzureParentTreeItem<IStorageRoot> {
...
+    }
+
+    public async disableStaticWebsite(): Promise<void> {
+        if (!this.websiteHostingEnabled) {
+            window.showInformationMessage(`Account '${this.label}' does not currently have staitc web hosting enabled.`);
+            return;
+        }
+        let confirmDisable: MessageItem = await ext.ui.showWarningMessage(`Are you sure you want to disable static web hosting for the account '${this.label}'?`, DialogResponses.yes, DialogResponses.cancel);
+        if (confirmDisable === DialogResponses.yes) {
+            await this.setWebsiteHostingProperties({ Enabled: false });
```
EricJizbaMSFT: This still doesn't work. Follow these repro steps:
1. Configure static website hosting and use "example.html" as the index document
1. Disable static website hosting
1. Configure static website hosting

Expected:
example.html is used as the default value for index prompt

Actual:
index.html is used as the default value for index prompt

src/azureStorageExplorer/storageAccounts/storageAccountNode.ts
```diff
@@ -249,19 +284,84 @@ export class StorageAccountTreeItem extends AzureParentTreeItem<IStorageRoot> {
...
+        let newStatus: azureStorage.common.models.ServicePropertiesResult.StaticWebsiteProperties = {
+            Enabled: true,
+            ErrorDocument404Path: errorDocument404Path,
+            IndexDocument: indexDocument
+        };
+        await this.setWebsiteHostingProperties(newStatus);
+        let msg = oldStatus.enabled ?
+            'Static website hosting configuration updated.' :
+            `The storage account '${this.label}' has been enabled for static website hosting.`;
+        this.websiteHostingEnabled = newStatus.Enabled;
```
EricJizbaMSFT: I still think you should implement `refreshImpl` in `blobContainerNode` (as described in the other comment) and remove the `websiteHostingEnabled` property in the file. For example, follow these steps:
1. Expand the storage account until you can see $web node
1. Change the website hosting status in the portal
1. Refresh the $web node

Expected: icon for $web node changes
Actual: Icon doesn't change

src/azureStorageExplorer/storageAccounts/storageAccountNode.ts
```diff
@@ -249,19 +284,84 @@ export class StorageAccountTreeItem extends AzureParentTreeItem<IStorageRoot> {
...
+            await this.refresh();
+        }
 
-        let resourceId = `${this.fullId}/staticWebsite`;
-        this.openInPortal(resourceId);
+    }
+
+    public async disableStaticWebsite(): Promise<void> {
+        if (!this.websiteHostingEnabled) {
+            window.showInformationMessage(`Account '${this.label}' does not currently have staitc web hosting enabled.`);
```
EricJizbaMSFT: `staitc` 🙃

src/azureStorageExplorer/storageAccounts/storageAccountNode.ts
```diff
@@ -249,19 +284,84 @@ export class StorageAccountTreeItem extends AzureParentTreeItem<IStorageRoot> {
...
+        if (oldStatus.enabled !== newStatus.Enabled) {
+            await this.refresh();
+        }
 
-        let resourceId = `${this.fullId}/staticWebsite`;
-        this.openInPortal(resourceId);
+    }
+
+    public async disableStaticWebsite(): Promise<void> {
+        if (!this.websiteHostingEnabled) {
```
EricJizbaMSFT: This should call `getWebsiteHostingStatus` to get the latest value. We don't want to block them from disabling if a property on this class (aka `websiteHostingEnabled`) happens to be out of date

src/azureStorageExplorer/storageAccounts/storageAccountNode.ts
```diff
@@ -249,19 +284,84 @@ export class StorageAccountTreeItem extends AzureParentTreeItem<IStorageRoot> {
...
+            window.showInformationMessage(`Account '${this.label}' does not currently have staitc web hosting enabled.`);
+            return;
+        }
+        let confirmDisable: MessageItem = await ext.ui.showWarningMessage(`Are you sure you want to disable static web hosting for the account '${this.label}'?`, DialogResponses.yes, DialogResponses.cancel);
+        if (confirmDisable === DialogResponses.yes) {
```
EricJizbaMSFT: Again - I almost never use `DialogResponses.yes`. In this case I would prefer a button that said "Disable"

src/azureStorageExplorer/storageAccounts/storageAccountNode.ts
PrashanthCorp: Can you please define working? In the portal, disabling and enabling the capability carries no memory of previous preferences. This commit is on par with the portal behavior. I cannot ask Azure Storage to remember these preferences either - specifying the index document and error document path in this request leads to an XML invalid syntax error. 
image

Do you want us to remember these preferences (as opposed to the platform)?

src/azureStorageExplorer/selectStorageAccountNodeForCommand.ts
PrashanthCorp: I'm noticing spots outside the PR where there are inconsistencies too. (use of 'static site hosting' in the README.md). I'll fix those that I find, but I feel making them consistent across the extension would be PR creep.

src/azureStorageExplorer/storageAccounts/storageAccountNode.ts
```diff
@@ -249,19 +284,86 @@ export class StorageAccountTreeItem extends AzureParentTreeItem<IStorageRoot> {
...
-        this.openInPortal(resourceId);
+    }
+
+    public async disableStaticWebsite(): Promise<void> {
+        if (!this.websiteHostingEnabled) {
+            window.showInformationMessage(`Account '${this.label}' does not currently have static website hosting enabled.`);
+            return;
+        }
+        let confirmDisable: MessageItem = await ext.ui.showWarningMessage(`Are you sure you want to disable static web hosting for the account '${this.label}'?`, { title: "Disable" }, DialogResponses.cancel);
+        if (confirmDisable.title === "Disable") {
```
EricJizbaMSFT: The typical pattern I use for checking dialog responses is this:
```typescript
const disable: MessageItem = { title: "Disable" };
let result: MessageItem = await ext.ui.showWarningMessage(`Are you sure you want to disable static web hosting for the account '${this.label}'?`, disable, DialogResponses.cancel);
if (result === disable) {
```

It's better for two reasons:
1. You don't have to hard code the string "Disable" twice
1. If you accidently use `vscode.window.showInformationMessage` or `vscode.window.showWarningMessage` you could get a null ref when you try to access `result.title` (since those return undefined on cancel)

That being said if there's only one button other than "cancel", you can also do this:
```
// no need to check result since cancel throws UserCancelledError and only other option is 'Disable'
await ext.ui.showWarningMessage(`Are you sure you want to disable static web hosting for the account '${this.label}'?`, { title: "Disable" });
```

I think that option is a little more confusing and requires a comment every time, but it's less code. No strong preference on this aspect.

src/azureStorageExplorer/storageAccounts/storageAccountNode.ts
```diff
@@ -249,19 +284,86 @@ export class StorageAccountTreeItem extends AzureParentTreeItem<IStorageRoot> {
...
-        let resourceId = `${this.fullId}/staticWebsite`;
-        this.openInPortal(resourceId);
+    }
+
+    public async disableStaticWebsite(): Promise<void> {
+        if (!this.websiteHostingEnabled) {
+            window.showInformationMessage(`Account '${this.label}' does not currently have static website hosting enabled.`);
+            return;
+        }
+        let confirmDisable: MessageItem = await ext.ui.showWarningMessage(`Are you sure you want to disable static web hosting for the account '${this.label}'?`, { title: "Disable" }, DialogResponses.cancel);
```
EricJizbaMSFT: This should be modal. Again my rule of thumb is if it blocks the immediate action I was trying to take, then it should be in my face and I should be able to just click "Enter" to continue

src/azureStorageExplorer/selectStorageAccountNodeForCommand.ts
```diff
@@ -46,19 +45,21 @@ export async function selectStorageAccountTreeItemForCommand(
...
         await accountTreeItem.ensureHostingCapable(hostingStatus);
 
         if (options.askToConfigureWebsite && !hostingStatus.enabled) {
-            let result = await window.showInformationMessage(
-                `Website hosting is not enabled on storage account "${accountTreeItem.label}". Would you like to go to the portal to enable it?`,
-                DialogResponses.yes,
-                DialogResponses.no);
-            let enableResponse = (result === DialogResponses.yes);
+            let enableWebHostingPrompt = "Enable website hosting";
+            let result: { title: string } = await ext.ui.showWarningMessage(
```
EricJizbaMSFT: You switched from `window.showInformationMessage` to `ext.ui.showWarningMessage`, which I think is a good thing, but it has different behavior. Specifically it'll throw `UserCancelledError` instead of returning undefined. Which means the logic below related to `cancelStep` will no longer work. Here's how I would fix:
```typescript
actionContext.properties.cancelStep = 'StorageAccountWebSiteNotEnabled';
actionContext.properties.enableResponse = 'false';
// no need to check result since cancel throws UserCancelledError and only other option is 'Enable'
await ext.ui.showWarningMessage(
    `Website hosting is not enabled on storage account "${accountTreeItem.label}".`,
    { modal: true },
    { title: "Enable website hosting" });
actionContext.properties.enableResponse = 'true';
await accountTreeItem.configureStaticWebsite();
// Either way can't continue
```

(Also note you don't have to pass in `DialogResponses.cancel` - that's always included by default)

src/azureStorageExplorer/storageAccounts/storageAccountNode.ts
```diff
@@ -249,19 +275,83 @@ export class StorageAccountTreeItem extends AzureParentTreeItem<IStorageRoot> {
...
+            `The storage account '${this.label}' has been enabled for static website hosting.`;
+        window.showInformationMessage(msg);
+        if (oldStatus.enabled !== newStatus.Enabled) {
+            await ext.tree.refresh(this);
+        }
+
+    }
+
+    public async disableStaticWebsite(): Promise<void> {
+        if (!(await this.getActualWebsiteHostingStatus())) {
```
EricJizbaMSFT: You have to check `.enabled` on the result of this call

src/azureStorageExplorer/selectStorageAccountNodeForCommand.ts
```diff
@@ -46,19 +45,22 @@ export async function selectStorageAccountTreeItemForCommand(
...
-            actionContext.properties.enableResponse = String(enableResponse);
             actionContext.properties.cancelStep = 'StorageAccountWebSiteNotEnabled';
+            actionContext.properties.enableResponse = 'false';
+            let enableWebHostingPrompt = "Enable website hosting";
+            // don't check result since cancel throws UserCancelledError and only other option is 'Enable'
+            await ext.ui.showWarningMessage(
+                `Website hosting is not enabled on storage account "${accountTreeItem.label}".`,
+                { modal: true },
+                { title: enableWebHostingPrompt });
+            let enableResponse = 'true';
```
EricJizbaMSFT: I mean...this code works, but you don't need the `enableResponse` variable. No need to convert a string to a string in the below line and no need to check `if (enableResponse)` for a value that's always true

PR #276, opened 2019-01-03
Clean up pipelines and use typescript for gulp

EricJizbaMSFT:
Cosmos builds were failing because mac had Node 6 by default - it was only a matter of time until we ran into an issue in this repo as well. Mostly copied from functions repo. 

Related to https://github.com/Microsoft/vscode-cosmosdb/pull/987


.azure-pipelines/common/build.yml
```diff
@@ -1,4 +1,9 @@
 steps:
+- task: NodeTool@0
+  displayName: 'Use Node 8.x'
+  inputs:
+    versionSpec: 8.x
```
nturinski: Any reason to use node 8 instead of 10?

EricJizbaMSFT: Because I know that works and it's what we're using in other repos. Switching to 10 would be a bigger change and would require more testing

PR #293, opened 2019-02-08
Allow user to delete storage account

### LINKED ISSUE #291: Allow deleting a storage account
If we allow the user to create an account, we should also allow the user to delete an account.

PrashanthCorp:
Fixes #291

### DIFF
--- a/src/azureStorageExplorer/storageAccounts/storageAccountNode.ts
+++ b/src/azureStorageExplorer/storageAccounts/storageAccountNode.ts
@@ -63,7 +63,7 @@ export class StorageAccountTreeItem extends AzureParentTreeItem<IStorageRoot> {
     public static async createStorageAccountTreeItem(parent: AzureParentTreeItem, storageAccount: StorageAccountWrapper, client: StorageManagementClient): Promise<StorageAccountTreeItem> {
         const ti = new StorageAccountTreeItem(parent, storageAccount, client);
         // make sure key is initialized
-        await ti.refreshImpl();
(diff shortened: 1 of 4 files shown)


src/azureStorageExplorer/storageAccounts/storageAccountNode.ts
```diff
@@ -137,6 +137,19 @@ export class StorageAccountTreeItem extends AzureParentTreeItem<IStorageRoot> {
...
     }
 
+    public async deleteTreeItemImpl(): Promise<void> {
+        const message: string = `Are you sure you want to delete account '${this.label}' and all its contents?`;
+        const result = await window.showWarningMessage(message, { modal: true }, DialogResponses.deleteResponse, DialogResponses.cancel);
+        if (result === DialogResponses.deleteResponse) {
+            let storageManagementClient = createAzureClient(this.root, StorageManagementClient);
+            let parsedId = this.parseAzureResourceId(this.storageAccount.id);
+            let resourceGroupName = parsedId.resourceGroups;
+            await storageManagementClient.storageAccounts.deleteMethod(resourceGroupName, this.storageAccount.name);
```
StephenWeatherford: Do you need to show progress?

PrashanthCorp: I don't think so. We don't know the intermediate steps, so a progress bar does not communicate much better than the inProgress spin-wheel on the treeItem (we do that)

PR #296, opened 2019-02-11
Webpack

StephenWeatherford:
I won't merge before the current release unless we decide to combine them.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -5,9 +5,6 @@
 
 'use strict';
 
-const loadStartTime: number = Date.now();
-let loadEndTime: number;
-
 import * as vscode from 'vscode';
 import { commands } from 'vscode';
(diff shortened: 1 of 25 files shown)


src/extension.ts
```diff
@@ -63,7 +60,7 @@ export async function activate(context: vscode.ExtensionContext): Promise<AzureE
         registerTableGroupActionHandlers();
 
         vscode.window.registerTreeDataProvider('azureStorage', tree);
-        registerCommand('azureStorage.refresh', async (treeItem?: AzureTreeItem) => await tree.refresh(treeItem));
+        registerCommand('azureStorage.refresh', async (treeItem?: AzureTreeItem) => tree.refresh(treeItem));
```
EricJizbaMSFT: nit: why the change to this?

StephenWeatherford: I think that was an auto-fix, I remember seeing lint complain about that but don't think I changed it.  Seems harmless here either way, you agree?

EricJizbaMSFT: Yeah I guess it probably doesn't make a difference

PR #311, opened 2019-02-16
NOTICE.html generated

StephenWeatherford:
The TPN generator now is titled "NOTICE File Generator" and produces a different output format than before, so I'm calling it NOTICE.html.


StephenWeatherford:
Ping, going to release today. Thanks.

PR #343, opened 2019-05-15
Add creating and deleting storage account tests and related changes

v-etzha:
1. Change window.showWarningMessage() to ext.ui.WarningMessage().
2. Add a string util to get a random string.
3. Add creating and deleting storage account tests.


src/utils/stringUtils.ts
```diff
@@ -0,0 +1,11 @@
+/*---------------------------------------------------------------------------------------------
```
nturinski: Could you use the stringUtil file here as well please?

https://github.com/microsoft/vscode-azurestorage/blob/3cc88f32c5a308faa6486e553650e5700ca29f03/src/components/temporaryFile.ts

test/storageAccountActions.test.ts
```diff
@@ -0,0 +1,78 @@
...
+import { AzureTreeDataProvider, DialogResponses, TestAzureAccount, TestUserInput } from 'vscode-azureextensionui';
+import { ext, getRandomHexString, StorageAccountProvider } from '../extension.bundle';
+import { longRunningTestsEnabled } from './global.test';
+
+suite('Storage Account Actions', async function (this: ISuiteCallbackContext): Promise<void> {
+    this.timeout(1200 * 1000);
+    const resourceGroupsToDelete: string[] = [];
+    const testAccount: TestAzureAccount = new TestAzureAccount();
+    let storageAccountClient: StorageManagementClient;
+    const resourceGroupName: string = getRandomHexString();
```
nturinski: Can you just make `resourceGroupName` lowercase and get rid of `storageAccountName`?

test/storageAccountActions.test.ts
```diff
@@ -0,0 +1,78 @@
...
+        assert.ok(createdAccount);
+        assert.equal(createdAccount[0].name, storageAccountName, `The created Storage Account '${createdAccount[0].name}' is not equal to the expected one '${storageAccountName}'.`);
+    });
+
+    test("deleteStorageAccount", async () => {
+        const createdAccount1: StorageAccountListResult = await storageAccountClient.storageAccounts.listByResourceGroup(resourceGroupName);
+        assert.ok(createdAccount1);
+        ext.ui = new TestUserInput([storageAccountName, DialogResponses.deleteResponse.title]);
+        await vscode.commands.executeCommand('azureStorage.deleteStorageAccount');
+        const createdAccount2: StorageAccountListResult = await storageAccountClient.storageAccounts.listByResourceGroup(resourceGroupName);
```
nturinski: Try to get the specific resource that you deleted and see if it returns a value.  If we add tests later that create more resources, then this test will break.

test/storageAccountActions.test.ts
```diff
@@ -0,0 +1,78 @@
...
+            }
+        }
+        ext.tree.dispose();
+    });
+
+    test("createStorageAccount", async () => {
+        resourceGroupsToDelete.push(resourceGroupName);
+        ext.ui = new TestUserInput([storageAccountName, '$(plus) Create new resource group', resourceGroupName, 'East US']);
+        await vscode.commands.executeCommand('azureStorage.createGpv2Account');
+        const createdAccount: StorageAccountListResult = await storageAccountClient.storageAccounts.listByResourceGroup(resourceGroupName);
```
nturinski: This function always returns an Array type which will always be defined even if it's length is 0.  The assertion below only checks if it is not `undefined` so it's not actually checking that the storage account was created.

Use the `getProperties` command.  It will return a `StorageAccount` object that you can validate.

PR #347, opened 2019-05-29
npm audit fix

EricJizbaMSFT:


PrashanthCorp:
How are there changes in `package-lock.json` while the `package.json` is unchanged?

EricJizbaMSFT:
package.json only specifies direct dependencies. package-lock.json specifies all dependencies. I'm guessing it updated a dependency of a dependency

PR #350, opened 2019-05-31
Add Create Blob Container, Copy Key test cases

v-etzha:
Add new test cases:
1. Create Blob Container.
2. Copy Connection String.
3. Copy Primary Key.

### DIFF
--- a/test/storageAccountActions.test.ts
+++ b/test/storageAccountActions.test.ts
@@ -6,7 +6,8 @@
 import * as assert from 'assert';
 import { ResourceManagementClient } from 'azure-arm-resource';
 import { StorageManagementClient } from 'azure-arm-storage';
-import { StorageAccount } from 'azure-arm-storage/lib/models';
+import { BlobContainer, StorageAccount, StorageAccountKey } from 'azure-arm-storage/lib/models';
+import * as clipboardy from 'clipboardy';
 import { IHookCallbackContext, ISuiteCallbackContext } from 'mocha';
 import * as vscode from 'vscode';
 import { AzExtTreeDataProvider, DialogResponses, TestAzureAccount, TestUserInput } from 'vscode-azureextensionui';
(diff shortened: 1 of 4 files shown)


test/storageAccountActions.test.ts
```diff
@@ -59,6 +60,40 @@ suite('Storage Account Actions', async function (this: ISuiteCallbackContext): P
...
+        await vscode.commands.executeCommand('azureStorage.createBlobContainer');
+        const createdAccount: BlobContainer = await storageAccountClient.blobContainers.get(resourceName, resourceName, blobContainerName);
+        assert.ok(createdAccount);
+    });
+
+    test("copyConnectionString", async () => {
+        let key: string | undefined;
+        ext.ui = new TestUserInput([resourceName]);
+        await vscode.commands.executeCommand('azureStorage.copyConnectionString');
+        const connectionString: string = await clipboardy.read();
```
nturinski: Whenever you read from the clipboard, you need to write an empty string to it first since we don't know what will be there before you read from it.

test/storageAccountActions.test.ts
```diff
@@ -59,6 +60,40 @@ suite('Storage Account Actions', async function (this: ISuiteCallbackContext): P
...
+        const listKeys: StorageAccountKey[] | undefined = (await storageAccountClient.storageAccounts.listKeys(resourceName, resourceName)).keys;
```
nturinski: This `copyConnectionString` test is testing the clipboard API more than our extension.  A test that would test the connection string more thoroughly would be attempting to connect with the `connectionString` and verifying that it works.

v-etzha: Hi Nathan, yes, you're correct. Connecting with the connection string would be more thorough. Does it mean we will operate Storage Explorer? If yes, could you please provide some methods that we can use?

nturinski: I think you can use the storage API to call this command `createBlobService` which accepts a connectionString as a parameter.  Once you run that, you can basically run anything against that to see if it was a valid connectionString.

This is an example of it:
https://github.com/microsoft/vscode-azurefunctions/blob/654bcbe7834c4b399db1bf2853509c77992d23e7/src/debug/validatePreDebug.ts#L55-L67

EricJizbaMSFT:

PR #363, opened 2019-06-11
Can view the file hierarchy of azure storage file shares in file explorer and can delete files

### LINKED ISSUE #355: [file share] List files and folders in file explorer

monalgarg:
- Implemented the vscode.FileSystemProvider API in fileShareFileSystemProvider.ts. 

- Created the lookup method which finds files and directories starting from the file share folder. 

- Deleting files uses the already pre-existing method to delete files. 

Fixes #355


main.js
```diff
@@ -14,11 +14,13 @@ let perfStats = {
...
 
-Object.defineProperty(exports, "__esModule", { value: true });
+Object.defineProperty(exports, "__esModule", {
+    value: true
+});
 
 const ignoreBundle = !/^(false|0)?$/i.test(process.env.AZCODE_STORAGE_IGNORE_BUNDLE || '');
 const extensionPath = ignoreBundle ? "./out/src/extension" : "./dist/extension.bundle";
-const extension = require(extensionPath);
+const extension = require("./out/src/extension");
```
EricJizbaMSFT: This was just for your own debugging and shouldn't be checked in

package.json
```diff
@@ -712,6 +728,7 @@
 		"ms-vscode.azure-account"
 	],
 	"dependencies": {
+		"az": "^0.2.3",
```
EricJizbaMSFT: I'm guessing you added "az" by accident?

package.json
```diff
@@ -88,6 +90,15 @@
 			]
 		},
 		"commands": [
+			{
+				"command": "azureStorage.fileShareTest",
```
EricJizbaMSFT: Remember to change all things named "Test". This should have an actual name haha

package.json
```diff
@@ -665,6 +676,11 @@
 					"scope": "resource",
 					"type": "string",
 					"description": "The name of the task to run before deploying."
+				},
+				"azureStorage.enableTest": {
```
EricJizbaMSFT: Same here - name this something other than test and finish the description

src/azureStorageExplorer/fileShares/fileNode2.ts
```diff
@@ -0,0 +1,53 @@
...
+import * as path from 'path';
+import { Uri, window } from 'vscode';
+import { AzureParentTreeItem, AzureTreeItem, DialogResponses, UserCancelledError } from 'vscode-azureextensionui';
+import { getResourcesPath } from "../../constants";
+import { ext } from "../../extensionVariables";
+import { ICopyUrl } from '../../ICopyUrl';
+import { IStorageRoot } from "../IStorageRoot";
+import { deleteFile } from './fileUtils';
+
+export class FileTreeItem2 extends AzureTreeItem<IStorageRoot> implements ICopyUrl {
```
EricJizbaMSFT: Are these exact duplicates of the original or did you change stuff? Ideally you should merge whatever you changed with the original files. The less copied code, the better.

src/extension.ts
```diff
@@ -61,6 +63,15 @@ export async function activateInternal(context: vscode.ExtensionContext, perfSta
         registerTableActionHandlers();
         registerTableGroupActionHandlers();
 
+        // tslint:disable-next-line: strict-boolean-expressions
+        if (vscode.workspace.getConfiguration("azureStorage").get("enableTest")) {
```
EricJizbaMSFT: Use extensionPrefix from constants.ts (https://github.com/microsoft/vscode-azurestorage/blob/master/src/constants.ts#L15)

And add "enableTest" to `configurationSettingsKeys`

The reason is so that you can reference a variable instead of typing out `"azureStorage"` each time - aka it should prevent accidental typos.

src/extension.ts
```diff
@@ -61,6 +63,15 @@ export async function activateInternal(context: vscode.ExtensionContext, perfSta
...
+            context.subscriptions.push(vscode.workspace.registerFileSystemProvider('azurestorage', new FileShareFS(), { isCaseSensitive: true }));
+        }
+
+        registerCommand('azureStorage.fileShareTest', async (_actionContext: IActionContext, treeItem: FileShareTreeItem) => {
+            // tslint:disable-next-line: prefer-template
+            commands.executeCommand('vscode.openFolder', vscode.Uri.parse('azurestorage://' + treeItem.fullId));
```
EricJizbaMSFT: add an `await` before `commands.executeCommand`

src/azureStorageExplorer/fileShares/fileShareFileSystemProvider.ts
```diff
@@ -0,0 +1,200 @@
...
+        throw new Error("Method not implemented.");
+    }
+
+    writeFile(_uri: vscode.Uri, _content: Uint8Array, _options: { create: boolean; overwrite: boolean; }): void | Thenable<void> {
+        throw new Error("Method not implemented.");
+    }
+
+    // tslint:disable-next-line: no-reserved-keywords
+    async delete(uri: vscode.Uri, _options: { recursive: boolean; }): Promise<void> {
+        let fileFound: FileTreeItem2 = await this.lookupAsFile(uri, false);
```
EricJizbaMSFT: Let's do `throw new Error("Method not implemented.");` for now.  I think this would fail for directories and I'd rather save "delete" for another PR

src/azureStorageExplorer/fileShares/fileShareFileSystemProvider.ts
```diff
@@ -0,0 +1,200 @@
...
+                } else {
+                    if (!silent) {
+                        throw vscode.FileSystemError.FileNotFound(uri);
+                    } else {
+                        return undefined;
+                    }
+                }
+            }
+
+            context.errorHandling.rethrow = true;
```
EricJizbaMSFT: You should set these at the beginning of `callWithTelemetryAndErrorHandling`. If you throw an error before these lines, it'll never reach these lines.

src/azureStorageExplorer/fileShares/fileShareFileSystemProvider.ts
```diff
@@ -0,0 +1,200 @@
...
+        size: number
+    ) {
+        this.type = type;
+        this.ctime = ctime;
+        this.mtime = mtime;
+        this.size = size;
+    }
+}
+
+export class FileShareFS implements vscode.FileSystemProvider {
```
EricJizbaMSFT: In general the file should be named the same as the main class - aka `FileShareFS.ts`

src/azureStorageExplorer/fileShares/fileShareFileSystemProvider.ts
```diff
@@ -0,0 +1,200 @@
...
+            }
+        }
+
+        throw vscode.FileSystemError.FileNotFound(uri);
+    }
+
+    async readDirectory(uri: vscode.Uri): Promise<[string, vscode.FileType][]> {
+        let entry: DirectoryTreeItem | FileShareTreeItem2 = await this.lookupAsDirectory(uri, false);
+
+        let _continuationToken: azureStorage.common.ContinuationToken | undefined;
```
EricJizbaMSFT: I don't see you ever assigning to `_continuationToken`, which means you will only ever get the first "batch" of files. That's fine for this PR - but I would remove the variable, pass `undefined` below, and add a comment saying something like:
```
// Intentionally passing undefined for token - only supports listing first batch of files for now
```

src/azureStorageExplorer/fileShares/fileShareFileSystemProvider.ts
```diff
@@ -0,0 +1,200 @@
...
+
+                        let entries = listShareResult.entries;
+                        let fileShareResultChild = entries.find(element => element.name === part);
+
+                        if (fileShareResultChild) {
+                            shareResult = fileShareResultChild;
+                            entry = new FileShareTreeItem2(entry, shareResult);
+                            break;
+                        }
+                    } while (continuationToken !== undefined);
```
EricJizbaMSFT: This could be a LOT of files that takes a really long time. I propose you ignore the `continuationToken` for now and just display the first batch. We can handle other batches in a later PR. It'll probably end up being either "Load More..." or a setting describing how many pages to load or something.

monalgarg: This is not actually listing the files. This is just to look up certain files or directories. So, I think I should be going through all the children tree items.

src/azureStorageExplorer/fileShares/FileShareFS.ts
```diff
@@ -0,0 +1,205 @@
...
+    private async lookupAsDirectory(uri: vscode.Uri, silent: boolean): Promise<DirectoryTreeItem | FileShareTreeItem2> {
+        let entry = await this.lookup(uri, silent);
+        if (entry instanceof DirectoryTreeItem || entry instanceof FileShareTreeItem2) {
+            return entry;
+        }
+        throw vscode.FileSystemError.FileNotADirectory(uri);
+    }
+
+    private async lookup(uri: vscode.Uri, silent: boolean): Promise<EntryTreeItem | undefined> {
+        return <EntryTreeItem>await callWithTelemetryAndErrorHandling('fs.readDirectory', async (context) => {
```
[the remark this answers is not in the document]
monalgarg: readDirectory is a method in FileSystemProvider interface. So, I think I need to keep the name as fs.readDierectory.

PR #366, opened 2019-06-13
read file in file explorer

### LINKED ISSUE #356: [file share] Opening a read only file from file explorer

monalgarg:
Fixes #356

### DIFF
--- a/src/azureStorageExplorer/fileShares/FileShareFS.ts
+++ b/src/azureStorageExplorer/fileShares/FileShareFS.ts
@@ -88,28 +88,21 @@ export class FileShareFS implements vscode.FileSystemProvider {
         let treeItem: FileTreeItem = await this.lookupAsFile(uri, false);
 
         let fileService = treeItem.root.createFileService();
-        return await new Promise<Uint8Array>((resolve, reject) => {
-            fileService.getFileToText(treeItem.share.name, treeItem.directoryPath, treeItem.file.name, (error?: Error, _text?: string, _result?: azureStorage.FileService.FileResult, _response?: azureStorage.ServiceResponse) => {
-                if (!!error) {
-                    reject(error);
-                } else {
-                    resolve(this.strToUint8Array(_text));
-                }
-            });
-        });
+        return this.strToUint8Array(
+            await new Promise<string | undefined>((resolve, reject) => {
+                fileService.getFileToText(treeItem.share.name, treeItem.directoryPath, treeItem.file.name, (error?: Error, text?: string, _result?: azureStorage.FileService.FileResult, _response?: azureStorage.ServiceResponse) => {
+                    if (!!error) {
+                        reject(error);
+                    } else {
+                        resolve(text);
+                    }
+                });
+            }));
     }
 
     private strToUint8Array(str: string | undefined): Uint8Array {
-        if (str === undefined) {
-            return new Uint8Array(0);
-        }
-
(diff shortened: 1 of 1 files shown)


src/azureStorageExplorer/fileShares/FileShareFS.ts
```diff
@@ -84,8 +84,32 @@ export class FileShareFS implements vscode.FileSystemProvider {
         throw new Error("Method not implemented.");
     }
 
-    readFile(_uri: vscode.Uri): Uint8Array {
-        throw new Error("Method not implemented.");
+    async readFile(uri: vscode.Uri): Promise<Uint8Array> {
+        let treeItem: FileTreeItem = await this.lookupAsFile(uri, false);
+
+        let fileService = treeItem.root.createFileService();
+        return await new Promise<Uint8Array>((resolve, reject) => {
```
EricJizbaMSFT: You should do `this.strToUint8Array` outside of `await new Promise` (aka just do `resolve(text)`

In general, you should add as little code as possible inside `await new Promise`. The promise never finishes unless you call resolve or reject, which means if `this.strToUint8Array` throws an exception, this promise would never finish and the user would never get an error message.

src/azureStorageExplorer/fileShares/FileShareFS.ts
```diff
@@ -84,8 +84,32 @@ export class FileShareFS implements vscode.FileSystemProvider {
...
+            fileService.getFileToText(treeItem.share.name, treeItem.directoryPath, treeItem.file.name, (error?: Error, _text?: string, _result?: azureStorage.FileService.FileResult, _response?: azureStorage.ServiceResponse) => {
```
EricJizbaMSFT: nit: Rename `_text` to `text`. We only use the underscore for unused variables

src/azureStorageExplorer/fileShares/FileShareFS.ts
```diff
@@ -84,8 +84,32 @@ export class FileShareFS implements vscode.FileSystemProvider {
...
+                if (!!error) {
+                    reject(error);
+                } else {
+                    resolve(this.strToUint8Array(_text));
+                }
+            });
+        });
+    }
+
+    private strToUint8Array(str: string | undefined): Uint8Array {
```
EricJizbaMSFT: This is the kind of thing that I would google how to do rather than try to implement myself. In fact, I think this will work:

```
Buffer.from(text || '');
```

NOTE: `text || ''` is shorthand for `text === undefined ? '' : text`

src/azureStorageExplorer/fileShares/FileShareFS.ts
```diff
@@ -84,8 +84,25 @@ export class FileShareFS implements vscode.FileSystemProvider {
         throw new Error("Method not implemented.");
     }
 
-    readFile(_uri: vscode.Uri): Uint8Array {
-        throw new Error("Method not implemented.");
+    async readFile(uri: vscode.Uri): Promise<Uint8Array> {
+        let treeItem: FileTreeItem = await this.lookupAsFile(uri, false);
+
+        let fileService = treeItem.root.createFileService();
+        return this.strToUint8Array(
```
EricJizbaMSFT: It's kinda hard to read a bunch of nested calls like this. Also not a fan of one-line methods that are only used once (aka strToUint8Array). Would prefer something like:
```
const result = await new Promise...

return Buffer.from(result || '');
```

PR #367, opened 2019-06-13
writing to files through file explorer

### LINKED ISSUE #357: [file share] Open writeable file from file explorer

monalgarg:
Fixes #357


src/azureStorageExplorer/fileShares/FileShareFS.ts
```diff
@@ -88,8 +88,28 @@ export class FileShareFS implements vscode.FileSystemProvider {
...
-        throw new Error("Method not implemented.");
+    async writeFile(_uri: vscode.Uri, _content: Uint8Array, _options: { create: boolean; overwrite: boolean; }): Promise<void> {
+        let treeItem: FileTreeItem = await this.lookupAsFile(_uri, false);
+
+        let fileService = treeItem.root.createFileService();
+
+        let text: string = this.uint8ArrayToStr(_content);
+
+        // tslint:disable-next-line: no-void-expression
+        return await new Promise<void>((resolve, reject) => {
```
EricJizbaMSFT: Just do `await new ...` instead of `return await new` and you don't have to disable "no-void-expression"

src/azureStorageExplorer/fileShares/FileShareFS.ts
```diff
@@ -88,8 +88,28 @@ export class FileShareFS implements vscode.FileSystemProvider {
...
+            fileService.createFileFromText(treeItem.share.name, treeItem.directoryPath, treeItem.file.name, text, (error?: Error, _result?: azureStorage.FileService.FileResult, _response?: azureStorage.ServiceResponse) => {
```
EricJizbaMSFT: You can just remove the parameters you're not using (assuming they're after a parameter you _are_ using)

So instead of:
```
(error?: Error, _result?: azureStorage.FileService.FileResult, _response?: azureStorage.ServiceResponse)
```

Just do:
```
(error?: Error)
```

src/azureStorageExplorer/fileShares/FileShareFS.ts
```diff
@@ -88,8 +88,28 @@ export class FileShareFS implements vscode.FileSystemProvider {
...
+                if (!!error) {
+                    reject(error);
+                } else {
+                    resolve();
```
EricJizbaMSFT: What's the value of `_result` and should we be ignoring it?

src/azureStorageExplorer/fileShares/FileShareFS.ts
```diff
@@ -88,8 +88,28 @@ export class FileShareFS implements vscode.FileSystemProvider {
...
     }
 
-    writeFile(_uri: vscode.Uri, _content: Uint8Array, _options: { create: boolean; overwrite: boolean; }): void | Thenable<void> {
-        throw new Error("Method not implemented.");
+    async writeFile(_uri: vscode.Uri, _content: Uint8Array, _options: { create: boolean; overwrite: boolean; }): Promise<void> {
+        let treeItem: FileTreeItem = await this.lookupAsFile(_uri, false);
+
+        let fileService = treeItem.root.createFileService();
+
+        let text: string = this.uint8ArrayToStr(_content);
```
EricJizbaMSFT: Just do `content.toString()`

src/azureStorageExplorer/fileShares/FileShareFS.ts
```diff
@@ -88,8 +88,22 @@ export class FileShareFS implements vscode.FileSystemProvider {
         throw new Error("Method not implemented.");
     }
 
-    writeFile(_uri: vscode.Uri, _content: Uint8Array, _options: { create: boolean; overwrite: boolean; }): void | Thenable<void> {
-        throw new Error("Method not implemented.");
+    async writeFile(uri: vscode.Uri, content: Uint8Array, _options: { create: boolean; overwrite: boolean; }): Promise<void> {
```
EricJizbaMSFT: Feels like you should be using `options.create` and `options.overwrite` somehow

src/azureStorageExplorer/fileShares/FileShareFS.ts
```diff
@@ -103,8 +103,73 @@ export class FileShareFS implements vscode.FileSystemProvider {
...
+        let directoryPath: string = '';
+        let fileName: string;
+
+        if (treeItem instanceof FileShareGroupTreeItem) {
+            throw RangeError("Cannot create a FileShare folder through File Explorer.");
+        } else if (treeItem instanceof FileShareTreeItem) {
+            fileName = treeItem.share.name;
+        } else if (treeItem instanceof DirectoryTreeItem) {
+            directoryPath = treeItem.parentPath + treeItem.directory.name;
+            let arr: string[] = uri.path.split('/');
```
EricJizbaMSFT: Instead of 
```
let arr: string[] = uri.path.split('/');
fileName = arr[arr.length - 1];
```

I think you want:
```
fileName = path.basename(uri.path);
```

You will also need this at the top of the file:
```
import * as path from 'path';
```

src/azureStorageExplorer/fileShares/FileShareFS.ts
```diff
@@ -103,8 +103,73 @@ export class FileShareFS implements vscode.FileSystemProvider {
...
-    writeFile(_uri: vscode.Uri, _content: Uint8Array, _options: { create: boolean; overwrite: boolean; }): void | Thenable<void> {
-        throw new Error("Method not implemented.");
+    async writeFile(uri: vscode.Uri, content: Uint8Array, options: { create: boolean; overwrite: boolean; }): Promise<void> {
+        let treeItem: EntryTreeItem | undefined = await this.lookup(uri, false, true);
+        let fileTreeItem: FileTreeItem;
+
+        if (!!treeItem) {
+            if (uri.path !== treeItem.fullId) {
+                if (options.create) {
+                    fileTreeItem = await this.createFile(uri, content.length, treeItem);
```
EricJizbaMSFT: It's a little confusing to me that you're passing in `content.length`. Shouldn't it always be `0` when you're creating a new file?

src/azureStorageExplorer/fileShares/FileShareFS.ts
```diff
@@ -103,8 +103,73 @@ export class FileShareFS implements vscode.FileSystemProvider {
...
+                } else {
+                    throw vscode.FileSystemError.FileNotFound(uri);
+                }
+            } else {
+                if (options.create && !options.overwrite) { // not given permission to overwrite and can only create a new file
```
EricJizbaMSFT: What should happen if the file already exists, `options.create` is true, _and_ `options.overwrite` is true? Is that possible? I feel like it should overwrite the file with a clean new file, but I'm not sure if we should warn the user first.

src/azureStorageExplorer/fileShares/FileShareFS.ts
```diff
@@ -103,8 +103,73 @@ export class FileShareFS implements vscode.FileSystemProvider {
...
+                });
+            });
+
+            // tslint:disable-next-line: strict-boolean-expressions
+            if (!!fileResult) {
+                return;
+            }
+        }
+
+        throw vscode.FileSystemError.FileNotFound(uri);
```
EricJizbaMSFT: I don't like having `FileNotFound` as the "catch-all" error at the end here. I would prefer if we were more deliberate in each case

src/azureStorageExplorer/fileShares/FileShareFS.ts
```diff
@@ -141,49 +206,49 @@ export class FileShareFS implements vscode.FileSystemProvider {
...
                         parentPath = parentPath + part + '/';
-                    }
-                    if (fileResultChild) {
-                        entry = new FileTreeItem(entry, fileResultChild, parentPath, <azureStorage.FileService.ShareResult>shareResult);
+                    } else {
+                        let fileResultChild = entries.files.find(element => element.name === part);
+                        if (fileResultChild) {
+                            entry = new FileTreeItem(entry, fileResultChild, parentPath, <azureStorage.FileService.ShareResult>entry.share);
+                        } else {
+                            if (terminateEarly && part === parts[parts.length - 1]) {
```
EricJizbaMSFT: Can you explain the purpose of `part === parts[parts.length - 1]`?

src/azureStorageExplorer/fileShares/FileShareFS.ts
```diff
@@ -116,23 +181,23 @@ export class FileShareFS implements vscode.FileSystemProvider {
         throw new Error("Method not implemented.");
     }
 
-    private async lookupAsDirectory(uri: vscode.Uri, silent: boolean): Promise<DirectoryTreeItem | FileShareTreeItem> {
+    private async lookupAsFile(uri: vscode.Uri, silent: boolean): Promise<FileTreeItem> {
```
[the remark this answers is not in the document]
monalgarg: I hand lookupAsDirectory before lookupAsFile and then switched the order.

src/azureStorageExplorer/fileShares/FileShareFS.ts
monalgarg: In cases where the file exists and you try to save it, both create and overwrite are set to true. So, I don't think we should create a new file when both are set to true.

src/azureStorageExplorer/fileShares/FileShareFS.ts
monalgarg: This is to ensure that you can create a file that is strictly only one level below an existing directory. I can change it if we don't want to keep that restriction.

EricJizbaMSFT: What if I have a path like this, though:
```
/a/a/a/a
```

src/azureStorageExplorer/fileShares/FileShareFS.ts
monalgarg: If I pass in 0 for the number of bytes, a new file is not created. I could pass in 1?

src/azureStorageExplorer/fileShares/FileShareFS.ts
```diff
@@ -37,6 +38,8 @@ class FileStatImpl implements vscode.FileStat {
 
 export class FileShareFS implements vscode.FileSystemProvider {
 
+    root: EntryTreeItem;
```
EricJizbaMSFT: Won't this fail as soon as I open more than one file share? I suggest doing something like:
```
private _rootMap: Map<string, EntryTreeItem> = new Map<string, EntryTreeItem>();
```

instead of 
```
root: EntryTreeItem;
```

src/azureStorageExplorer/fileShares/FileShareFS.ts
```diff
@@ -103,8 +105,65 @@ export class FileShareFS implements vscode.FileSystemProvider {
...
-        throw new Error("Method not implemented.");
+    async writeFile(uri: vscode.Uri, content: Uint8Array, options: { create: boolean; overwrite: boolean; }): Promise<void> {
+        let fileTreeItem: FileTreeItem;
+
+        if (options.create) {
+            let dirUri = vscode.Uri.file(path.dirname(uri.path));
+            let dirTreeItem: FileShareTreeItem | DirectoryTreeItem = await this.lookupAsDirectory(dirUri);
+
+            // tslint:disable-next-line: no-non-null-assertion
+            let listFilesAndDirectoriesResult = await dirTreeItem.listFiles(<azureStorage.common.ContinuationToken>undefined!);
```
EricJizbaMSFT: There should be an api like `fileService.doesFileExist` that you can call instead of listing out all files (which will be slower performance-wise)

src/azureStorageExplorer/fileShares/FileShareFS.ts
```diff
@@ -103,8 +105,65 @@ export class FileShareFS implements vscode.FileSystemProvider {
...
+
+            if (fileResultChild) {
+                if (!options.overwrite) {
+                    throw vscode.FileSystemError.FileExists(uri);
+                }
+                // tslint:disable-next-line: restrict-plus-operands
+                let parentPath: string = dirTreeItem instanceof DirectoryTreeItem ? dirTreeItem.parentPath + dirTreeItem.directory : '';
+                fileTreeItem = new FileTreeItem(dirTreeItem, fileResultChild, parentPath, <azureStorage.FileService.ShareResult>dirTreeItem.share);
+            } else {
+                if (!options.create) {
```
EricJizbaMSFT: This code is not reachable since you're already inside `if (options.create) {` above

src/azureStorageExplorer/fileShares/FileShareFS.ts
```diff
@@ -103,8 +105,65 @@ export class FileShareFS implements vscode.FileSystemProvider {
...
+            let listFilesAndDirectoriesResult = await dirTreeItem.listFiles(<azureStorage.common.ContinuationToken>undefined!);
+            let fileResultChild = listFilesAndDirectoriesResult.entries.files.find(element => element.name === path.basename(uri.path));
+
+            if (fileResultChild) {
+                if (!options.overwrite) {
+                    throw vscode.FileSystemError.FileExists(uri);
+                }
+                // tslint:disable-next-line: restrict-plus-operands
+                let parentPath: string = dirTreeItem instanceof DirectoryTreeItem ? dirTreeItem.parentPath + dirTreeItem.directory : '';
+                fileTreeItem = new FileTreeItem(dirTreeItem, fileResultChild, parentPath, <azureStorage.FileService.ShareResult>dirTreeItem.share);
```
EricJizbaMSFT: If overwrite is true, it should still fall through to `this.createFile` so that you overwrite it with a blank new file.

I believe your current logic isn't actually overwriting anything.

src/azureStorageExplorer/fileShares/FileShareFS.ts
```diff
@@ -103,8 +105,65 @@ export class FileShareFS implements vscode.FileSystemProvider {
...
+                fileTreeItem = await this.createFile(uri, dirTreeItem);
+                return;
+            }
+        } else {
+            fileTreeItem = await this.lookupAsFile(uri);
+        }
+
+        let fileService = fileTreeItem.root.createFileService();
+        // tslint:disable-next-line: no-void-expression
+        return await new Promise<void>((resolve, reject) => {
```
EricJizbaMSFT: nit: Move all of this into the else case above to make it clear it only happens when `(!options.create)`

src/azureStorageExplorer/fileShares/FileShareFS.ts
```diff
@@ -103,8 +105,65 @@ export class FileShareFS implements vscode.FileSystemProvider {
...
+                if (!!error) {
+                    reject(error);
+                } else {
+                    resolve();
+                }
+            });
+        });
+    }
+
+    private async createFile(uri: vscode.Uri, treeItem: EntryTreeItem): Promise<FileTreeItem> {
```
EricJizbaMSFT: change the type of `treeItem` to `FileShareTreeItem | DirectoryTreeItem` since you will only call it after calling `lookupAsDirectory` above. That should help simplify this method.

If the function ends up being really simple after you make that change, you might want to remove this method completely and move the code up into `writeFile`

src/azureStorageExplorer/fileShares/FileShareFS.ts
```diff
@@ -116,85 +175,103 @@ export class FileShareFS implements vscode.FileSystemProvider {
...
 
-            let shareResult: azureStorage.FileService.ShareResult | undefined;
-
-            for (const part of parts) {
-                if (entry instanceof FileShareGroupTreeItem) {
-
-                    // Intentionally passing undefined for token - only supports listing first batch of files for now
-                    // tslint:disable-next-line:no-non-null-assertion // currentToken argument typed incorrectly in SDK
-                    let listShareResult = await entry.listFileShares(<azureStorage.common.ContinuationToken>undefined!);
+            if (this.rootMap.get(parts[0]) === undefined) {
```
EricJizbaMSFT:
```
if (!this.rootMap.has(parts[0])) {
```

src/azureStorageExplorer/fileShares/FileShareFS.ts
monalgarg: The createFileFromText replaces the entire file's previous content with the new content. So, I think it is overwriting the file.

src/azureStorageExplorer/fileShares/FileShareFS.ts
nturinski: This one hasn't been addressed yet.  

It's a little nit-picky, but if you just remove the `return` statement and leave the rest, the function will implicitly return `void` since there's no return statement.

And then you can remove the `//tslint:disable` line.

src/azureStorageExplorer/fileShares/FileShareFS.ts
```diff
@@ -37,6 +39,8 @@ class FileStatImpl implements vscode.FileStat {
 
 export class FileShareFS implements vscode.FileSystemProvider {
 
+    private rootMap: Map<string, FileShareTreeItem> = new Map<string, FileShareTreeItem>();
```
nturinski: nit: `private` variables usually are prefixed with an `_`

src/azureStorageExplorer/fileShares/FileShareFS.ts
```diff
@@ -103,8 +105,59 @@ export class FileShareFS implements vscode.FileSystemProvider {
...
+    async writeFile(uri: vscode.Uri, content: Uint8Array, options: { create: boolean; overwrite: boolean; }): Promise<void> {
+        if (!options.create && !options.overwrite) {
+            throw vscode.FileSystemError.NoPermissions(uri);
+        }
+
+        let dirUri = vscode.Uri.file(path.dirname(uri.path));
+        let dirTreeItem: FileShareTreeItem | DirectoryTreeItem = await this.lookupAsDirectory(dirUri);
+
+        // tslint:disable-next-line: restrict-plus-operands
+        const parentPath: string = dirTreeItem instanceof DirectoryTreeItem ? dirTreeItem.parentPath + dirTreeItem.directory.name : '';
```
nturinski: Now that you have path in here, I would just use `path.join` to create this `parentPath`.  Then you can remove the tslint rule line above

Though, does `parentPath` automatically include a `/` in it?  That might screw up the join

src/azureStorageExplorer/fileShares/FileShareFS.ts
```diff
@@ -103,8 +105,59 @@ export class FileShareFS implements vscode.FileSystemProvider {
...
+        }
+
+        if (!fileResultChild) {
+            const fileName: string = path.basename(uri.path);
+            fileResultChild = await createFile(parentPath, fileName, dirTreeItem.share, dirTreeItem.root);
+        }
+
+        let fileTreeItem: FileTreeItem = new FileTreeItem(dirTreeItem, fileResultChild, parentPath, <azureStorage.FileService.ShareResult>dirTreeItem.share);
+
+        if (!options.overwrite) { return; }
```
nturinski: Why does this just return?   It shouldn't throw an error like the other conditional statements do?

src/azureStorageExplorer/fileShares/FileShareFS.ts
```diff
@@ -116,7 +169,7 @@ export class FileShareFS implements vscode.FileSystemProvider {
                 throw new Error("Azure storage does not support nonrecursive deletion of folders.");
             }
 
-            let fileFound: EntryTreeItem | undefined = await this.lookup(uri, false);
+            let fileFound: EntryTreeItem | undefined = await this.lookup(uri);
```
nturinski: I know this isn't your code, but I think that the check below is unnecessary because `lookup` shouldn't return `undefined` and throws the same error on `L250`.  Since you're refactoring anyway, I think we can remove this 😃

src/azureStorageExplorer/fileShares/FileShareFS.ts
```diff
@@ -134,85 +187,103 @@ export class FileShareFS implements vscode.FileSystemProvider {
...
-                    let fileShareResultChild = entries.find(element => element.name === part);
+            let entry: EntryTreeItem;
+            let root = this._rootMap.get(parts[0]);
 
-                    if (fileShareResultChild) {
-                        shareResult = fileShareResultChild;
-                        entry = new FileShareTreeItem(entry, shareResult);
-                    }
-                } else if (entry instanceof FileShareTreeItem || entry instanceof DirectoryTreeItem) {
+            if (root === undefined) {
```
nturinski: `root` should never be `undefined` I think.  When `this.findRoot` is called, it sets the _rootMap `(this._rootMap.set(fileShareName, <FileShareTreeItem>rootFound);)` otherwise it'll throw an error.  I think that this check is unneccessary.

src/azureStorageExplorer/fileShares/FileShareFS.ts
```diff
@@ -134,85 +187,103 @@ export class FileShareFS implements vscode.FileSystemProvider {
...
+                throw new RangeError('Could not find File Share.');
+            } else {
+                entry = root;
+            }
 
+            for (let part of parts.slice(1)) {
```
nturinski: Sorry, but could you explain why your do `parts.slice(1)` here and on line 217?  What are you slicing?

src/azureStorageExplorer/fileShares/FileShareFS.ts
```diff
@@ -134,85 +187,103 @@ export class FileShareFS implements vscode.FileSystemProvider {
...
+                rootPath = uri.path;
+            } else {
+                rootPath = uri.path.substring(0, endOfFileShareName);
+            }
+
+            let rootFound = await ext.tree.findTreeItem(rootPath, context);
+
+            let fileShareName = uri.path.substring(endOfFileShareIndx, endOfFileShareName);
+
+            if (rootFound && rootFound instanceof FileShareTreeItem) {
```
nturinski: I don't think the rootFound check is required since `undefined` wouldn't be an instance of `FileShareTreeItem` either

src/azureStorageExplorer/fileShares/FileShareFS.ts
```diff
@@ -134,85 +187,103 @@ export class FileShareFS implements vscode.FileSystemProvider {
...
+
+            if (endOfFileShareName === -1) {
+                rootPath = uri.path;
+            } else {
+                rootPath = uri.path.substring(0, endOfFileShareName);
+            }
+
+            let rootFound = await ext.tree.findTreeItem(rootPath, context);
+
+            let fileShareName = uri.path.substring(endOfFileShareIndx, endOfFileShareName);
```
nturinski: `fileShareName` can be in the if statement since it's not called unless `rootFound` is a `FileShareTreeItem`

src/azureStorageExplorer/fileShares/FileShareFS.ts
```diff
@@ -241,7 +241,7 @@ export class FileShareFS implements vscode.FileSystemProvider {
                     if (directoryResultChild) {
                         entry = new DirectoryTreeItem(entry, parentPath, directoryResultChild, <azureStorage.FileService.ShareResult>entry.share);
                         // tslint:disable-next-line: prefer-template
-                        parentPath = parentPath + part + '/';
+                        parentPath = path.join(parentPath, part);
```
nturinski: So now you can remove the `// tslint:disable-next-line: prefer-template` above it 😄  (same for line 116)

src/azureStorageExplorer/fileShares/FileShareFS.ts
monalgarg: So, because I am storing the File Share roots, my entry already starts at the File Share level. But, the parts array includes the FIle Share name. So, because I am already starting at the level after the File Share, I don't need that first element in the parts array.

nturinski: So just want to make sure, but you are intending to slice the array twice, losing the first two elements in the parts array?

src/azureStorageExplorer/fileShares/FileShareFS.ts
```diff
@@ -103,8 +105,58 @@ export class FileShareFS implements vscode.FileSystemProvider {
...
-        throw new Error("Method not implemented.");
+    async writeFile(uri: vscode.Uri, content: Uint8Array, options: { create: boolean; overwrite: boolean; }): Promise<void> {
+        if (!options.create && !options.overwrite) {
+            throw vscode.FileSystemError.NoPermissions(uri);
+        }
+
+        let dirUri = vscode.Uri.file(path.dirname(uri.path));
+        let dirTreeItem: FileShareTreeItem | DirectoryTreeItem = await this.lookupAsDirectory(dirUri);
+
+        // tslint:disable-next-line: restrict-plus-operands
```
nturinski: nit: This can go now

PR #368, opened 2019-06-13
Delete files and directories through file explorer

monalgarg:


src/azureStorageExplorer/fileShares/FileShareFS.ts
```diff
@@ -93,8 +93,14 @@ export class FileShareFS implements vscode.FileSystemProvider {
     }
 
     // tslint:disable-next-line: no-reserved-keywords
-    delete(_uri: vscode.Uri, _options: { recursive: boolean; }): void {
-        throw new Error("Method not implemented.");
+    async delete(uri: vscode.Uri, _options: { recursive: boolean; }): Promise<void> {
+        return <void>await callWithTelemetryAndErrorHandling('fs.delete', async (context) => {
+            let fileFound: EntryTreeItem | undefined = await this.lookup(uri, false);
```
EricJizbaMSFT: What about if you don't find the file? Shouldn't that throw a FileNotFound error or something?

src/azureStorageExplorer/fileShares/FileShareFS.ts
```diff
@@ -93,8 +93,14 @@ export class FileShareFS implements vscode.FileSystemProvider {
     }
 
     // tslint:disable-next-line: no-reserved-keywords
-    delete(_uri: vscode.Uri, _options: { recursive: boolean; }): void {
-        throw new Error("Method not implemented.");
+    async delete(uri: vscode.Uri, _options: { recursive: boolean; }): Promise<void> {
```
EricJizbaMSFT: Should we be ignoring options.recursive? Is it always passed as true?

src/azureStorageExplorer/fileShares/FileShareFS.ts
```diff
@@ -93,8 +93,24 @@ export class FileShareFS implements vscode.FileSystemProvider {
     }
 
     // tslint:disable-next-line: no-reserved-keywords
-    delete(_uri: vscode.Uri, _options: { recursive: boolean; }): void {
-        throw new Error("Method not implemented.");
+    async delete(uri: vscode.Uri, options: { recursive: boolean; }): Promise<void> {
+        return <void>await callWithTelemetryAndErrorHandling('fs.delete', async (context) => {
+            if (!options.recursive) {
+                throw new Error("Azure storage does not support nonrecusive deletion of folders.");
```
EricJizbaMSFT: Spelling "nonrecusive"

src/azureStorageExplorer/fileShares/FileShareFS.ts
```diff
@@ -93,8 +93,24 @@ export class FileShareFS implements vscode.FileSystemProvider {
     }
 
     // tslint:disable-next-line: no-reserved-keywords
-    delete(_uri: vscode.Uri, _options: { recursive: boolean; }): void {
-        throw new Error("Method not implemented.");
+    async delete(uri: vscode.Uri, options: { recursive: boolean; }): Promise<void> {
+        return <void>await callWithTelemetryAndErrorHandling('fs.delete', async (context) => {
```
EricJizbaMSFT: Just do `await ...` instead of `return <void>await...`

src/azureStorageExplorer/fileShares/FileShareFS.ts
```diff
@@ -93,8 +93,24 @@ export class FileShareFS implements vscode.FileSystemProvider {
...
+            if (!options.recursive) {
+                throw new Error("Azure storage does not support nonrecusive deletion of folders.");
+            }
+
+            let fileFound: EntryTreeItem | undefined = await this.lookup(uri, false);
+
+            if (fileFound === undefined) {
+                throw vscode.FileSystemError.FileNotFound(uri);
```
EricJizbaMSFT: This error and all other errors will be swallowed by `callWithTelemetryAndErrorHandling`. Is that the correct behavior? Or should you be using one or both of these properties:
```
context.errorHandling.suppressDisplay
context.errorHandling.rethrow
```

If you don't know how to make the error happen, just fake it by temporarily flipping the `if` condition and make sure it's displayed to the user in a way that makes sense and nothing else weird happens.

src/azureStorageExplorer/fileShares/FileShareFS.ts
```diff
@@ -93,8 +93,26 @@ export class FileShareFS implements vscode.FileSystemProvider {
     }
 
     // tslint:disable-next-line: no-reserved-keywords
-    delete(_uri: vscode.Uri, _options: { recursive: boolean; }): void {
-        throw new Error("Method not implemented.");
+    async delete(uri: vscode.Uri, options: { recursive: boolean; }): Promise<void> {
+        await callWithTelemetryAndErrorHandling('fs.delete', async (context) => {
+            context.errorHandling.rethrow = true;
```
EricJizbaMSFT: I'm trusting that you tested the error handling part and assuming that there was no weird behavior :)

PR #374, opened 2019-06-24
View virtual blob folder structure in file explorer and read blobs in file explorer

### LINKED ISSUE #369: [blob] View blob container folder structure through file explorer

### LINKED ISSUE #372: [blob] Read blobs through File Explorer

monalgarg:
Fixes #369 

Fixes #372


src/azureStorageExplorer/blobContainers/blobContainerFS.ts
```diff
@@ -0,0 +1,245 @@
...
+import { BlobTreeItem } from './blobNode';
+
+export type EntryTreeItem = BlobContainerGroupTreeItem | BlobContainerTreeItem | BlobDirectoryTreeItem | BlobTreeItem;
+
+export class BlobContainerFS implements vscode.FileSystemProvider {
+
+    private rootMap: Map<string, BlobContainerTreeItem> = new Map<string, BlobContainerTreeItem>();
+
+    // tslint:disable-next-line: typedef
+    private _emitter = new vscode.EventEmitter<vscode.FileChangeEvent[]>();
```
EricJizbaMSFT: Just set the type rather than disabling the tslint rule. You're welcome to advocate for turning off the rule if you want, but otherwise it should be followed

src/azureStorageExplorer/blobContainers/blobContainerFS.ts
```diff
@@ -0,0 +1,222 @@
...
+        } else if (entry instanceof BlobTreeItem) {
+            // creation and modification times as well as size of tree item are intentionally set to 0 for now
+            return new FileStatImpl(vscode.FileType.File, 0, 0, 0);
+        }
+
+        throw vscode.FileSystemError.FileNotFound(uri);
+    }
+
+    async readDirectory(uri: vscode.Uri): Promise<[string, vscode.FileType][]> {
+        let entry: EntryTreeItem = await this.lookup(uri);
```
EricJizbaMSFT: Seems like this should use a method called `lookupAsDirectory` to be consistent with `lookupAsBlob` and `FileShareFS`

src/azureStorageExplorer/fileShares/FileShareFS.ts
```diff
@@ -247,31 +233,52 @@ export class FileShareFS implements vscode.FileSystemProvider {
...
             context.errorHandling.rethrow = true;
             context.errorHandling.suppressDisplay = true;
 
-            const fileShareString = 'File Shares';
-            const endOfFileShareIndx = uri.path.indexOf(fileShareString) + fileShareString.length + 1;
-            const endOfFileShareName = uri.path.indexOf('/', endOfFileShareIndx);
-            let rootPath: string;
+            let parsedUri = FileShareFS.parseUri(uri, fileTypeString);
+            let rootPath = path.join(parsedUri.rootPath, parsedUri.groupTreeItemName);
+            return await ext.tree.findTreeItem(rootPath, context);
```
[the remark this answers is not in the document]
monalgarg: When I add to this root map, I have to do checks on the type of the root found , which is specific to either Blob or File Share. Because of this, I decided to split up the functionality into updateRootMap (which both fileshare and blob have their own version of) and the generic findRoot. Should I put those checks into the generic findRoot method?

src/azureStorageExplorer/fsp.ts
```diff
@@ -0,0 +1,43 @@
...
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as path from 'path';
+import * as vscode from 'vscode';
+import { AzExtTreeItem, callWithTelemetryAndErrorHandling } from 'vscode-azureextensionui';
+import { ext } from '../extensionVariables';
+import { FileShareTreeItem } from './fileShares/fileShareNode';
+
+export function parseUri(uri: vscode.Uri, fileType: string): { accountName: string, rootPath: string, groupTreeItemName: string, parentPath: string, baseName: string } {
```
EricJizbaMSFT: This is better, but `parseUri` can still be simpler and more effective. I think the goal of this function should be to completely parse and provide all combinations of the path that you would need so that you never have to do anything like `path.join(parsedUri.rootPath, parsedUri.groupTreeItemName)` or 
```
parsedUri.parentPath === '' && parsedUri.baseName === '' ? '' : `${path.join(parsedUri.parentPath, parsedUri.baseName)}/`;
```

Try this out for size and see if it provides all the paths you might need:
```typescript
export function parseUri(uri: vscode.Uri, fileType: string): IParsedUri {
    const matches: RegExpMatchArray | null = uri.path.match(`^(\/subscriptions\/[^\/]+\/resourceGroups\/[^\/]+\/providers\/Microsoft\.Storage\/storageAccounts\/[^\/]+\/${fileType}\/([^\/]+))\/?((.*?\/?)([^\/]*))$`);
    if (!matches) {
        throw new Error(`Invalid ${fileType} uri`);
    } else {
        return {
            rootPath: matches[1],
            rootName: matches[2],
            filePath: matches[3],
            dirPath: matches[3] ? `${matches[3]}/` : '',
            parentDirPath: matches[4],
            baseName: matches[5]
        };
    }
}
```

Also the naming is pretty confusing, so it would be good to add documentation with examples to the return type:
```typescript
/**
 * Example uri: /subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/resourcegroup1/providers/Microsoft.Storage/storageAccounts/storageaccount1/Blob Containers/container1/parentdir1/subdir
 */
export interface IParsedUri {
    /**
     * Path to container or file share
     * e.g. /subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/resourcegroup1/providers/Microsoft.Storage/storageAccounts/storageaccount1/Blob Containers/container1
     */
    rootPath: string;

    /**
     * Name of container or file share
     * e.g. container1
     */
    rootName: string;

    /**
     * Full path within container or file share
     * e.g. parentdir1/subdir
     */
    filePath: string;

    /**
     * Same as `filePath`, just with a delimiter at the end in case this is a directory
     * e.g. parentdir1/subdir/
     */
    dirPath: string;

    /**
     * Path of parent directory within container or file share
     * e.g. parentdir1/
     */
    parentDirPath: string;

    /**
     * Name of file or directory
     * e.g. subdir
     */
    baseName: string;
}
```

src/azureStorageExplorer/fsp.ts
```diff
@@ -0,0 +1,43 @@
...
+
+    const matchesTemp: RegExpMatchArray | null = uri.path.match(`^(.*)\/${fileType}\/(.*)$`);
+    if (!matchesTemp) {
+        throw new RangeError(`Not valid ${fileType} uri`);
+    }
+
+    return { accountName: matches[1], rootPath: `${matchesTemp[1]}/${fileType}`, groupTreeItemName: matches[2], parentPath: matches[3], baseName: matches[4] };
+}
+
+export async function findRoot(uri: vscode.Uri, fileTypeString: string): Promise<AzExtTreeItem | undefined> {
```
EricJizbaMSFT: > When I add to this root map, I have to do checks on the type of the root found , which is specific to either Blob or File Share. Because of this, I decided to split up the functionality into updateRootMap (which both fileshare and blob have their own version of) and the generic findRoot. Should I put those checks into the generic findRoot method?

Here's how I would split it up:
```typescript
const rootMap: Map<string, AzExtTreeItem> = new Map<string, AzExtTreeItem>();

export async function findRoot(context: IActionContext, uri: vscode.Uri, fileTypeString: string): Promise<AzExtTreeItem> {
    let parsedUri = parseUri(uri, fileTypeString);

    let root = rootMap.get(parsedUri.rootPath);
    if (root) {
        return root;
    } else {
        root = await ext.tree.findTreeItem(parsedUri.rootPath, context);
        if (!root) {
            throw vscode.FileSystemError.FileNotFound(uri);
        } else {
            rootMap.set(parsedUri.rootPath, root);
            return root;
        }
    }
}
```

And then you could remove `updateRootMap` and `_rootMap` from both classes and just do something like this:
```typescript
let entry = await findRoot(context, uri, this.blobContainerString);
if (!(entry instanceof BlobContainerTreeItem)) {
    throw new Error('todo');
}
```

src/azureStorageExplorer/fsp.ts
```diff
@@ -0,0 +1,43 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as path from 'path';
```
EricJizbaMSFT: `fsp.ts` doesn't mean much to me. I'd split this file up into two files - one per function with the function name as the file name

src/azureStorageExplorer/blobContainers/BlobContainerFS.ts
```diff
@@ -0,0 +1,188 @@
...
+
+    private async lookup(uri: vscode.Uri): Promise<EntryTreeItem> {
+        return <EntryTreeItem>await callWithTelemetryAndErrorHandling('blob.lookup', async (context) => {
+            context.errorHandling.rethrow = true;
+            context.errorHandling.suppressDisplay = true;
+
+            let parsedUri = parseUri(uri, this._blobContainerString);
+
+            let entry = await this.getRoot(uri);
+            // let entry: BlobContainerTreeItem = !!foundRoot ? foundRoot : await this.updateRootMap(uri);
```
EricJizbaMSFT: remove this comment

src/azureStorageExplorer/blobContainers/BlobContainerFS.ts
```diff
@@ -0,0 +1,188 @@
...
+                if (!!blobResultChild) {
+                    return new BlobTreeItem(entry, blobResultChild, entry.container);
+                }
+                throw vscode.FileSystemError.FileNotFound(uri);
+            }
+        });
+    }
+
+    private async getRoot(uri: vscode.Uri): Promise<BlobContainerTreeItem> {
+        return <BlobContainerTreeItem>await callWithTelemetryAndErrorHandling('blob.getRoot', async (context) => {
```
EricJizbaMSFT: Why does this have separate telemetry and error handling? It seems like you should just use the context from `lookup`, which is the only place this is called

src/azureStorageExplorer/fileShares/FileShareFS.ts
```diff
@@ -52,34 +31,42 @@ export class FileShareFS implements vscode.FileSystemProvider {
...
-            return new FileStatImpl(vscode.FileType.Directory, 0, 0, 0);
+            return { type: vscode.FileType.File, ctime: 0, mtime: 0, size: 0 };
         }
 
         throw vscode.FileSystemError.FileNotFound(uri);
     }
 
     async readDirectory(uri: vscode.Uri): Promise<[string, vscode.FileType][]> {
-        let entry: DirectoryTreeItem | FileShareTreeItem = await this.lookupAsDirectory(uri);
+        return <[string, vscode.FileType][]>await callWithTelemetryAndErrorHandling('fs.lookup', async (context) => {
```
EricJizbaMSFT: Change 'fs.lookup'

src/azureStorageExplorer/fileShares/FileShareFS.ts
```diff
@@ -110,14 +96,16 @@ export class FileShareFS implements vscode.FileSystemProvider {
             throw vscode.FileSystemError.NoPermissions(uri);
         }
 
-        let dirUri = vscode.Uri.file(path.dirname(uri.path));
-        let dirTreeItem: FileShareTreeItem | DirectoryTreeItem = await this.lookupAsDirectory(dirUri);
+        let parsedUri = parseUri(uri, this._fileShareString);
+        let dirTreeItem: DirectoryTreeItem | FileShareTreeItem | FileShareGroupTreeItem = await this.lookupAsDirectory(vscode.Uri.file(parsedUri.dirPath));
```
EricJizbaMSFT: I don't think you should be using `lookupAsDirectory` in this method because it will degrade performance. In the below `fileResultChild` can you use that to check if this already exists as a directory?

src/azureStorageExplorer/fileShares/FileShareFS.ts
```diff
@@ -52,34 +31,42 @@ export class FileShareFS implements vscode.FileSystemProvider {
     async stat(uri: vscode.Uri): Promise<vscode.FileStat | Thenable<vscode.FileStat>> {
         let treeItem: EntryTreeItem = await this.lookup(uri);
 
-        if (treeItem instanceof FileTreeItem) {
+        if (treeItem instanceof DirectoryTreeItem || treeItem instanceof FileShareTreeItem || treeItem instanceof FileShareGroupTreeItem) {
```
EricJizbaMSFT: Seems like you added `FileShareGroupTreeItem` a few places in this file, but again I think we decided just to support opening a single file share so that shouldn't be possible

PR #378, opened 2019-06-26
Don't prompt again during static website deployment if user chooses to create account 

PrashanthCorp:


src/azureStorageExplorer/storageAccounts/storageAccountActionHandlers.ts
```diff
@@ -75,9 +75,9 @@ async function deployStaticWebsite(context: IActionContext, target?: vscode.Uri
         context,
         {
             mustBeWebsiteCapable: true,
-            askToConfigureWebsite: true
+            askToConfigureWebsite: false
         });
-
+    await destAccountTreeItem.configureStaticWebsite();
```
EricJizbaMSFT: Your PR title says "if user chooses to create account", but as far as I can tell you changed the behavior even if user picks an _existing_ account.

Check out this code to see how you can detect a "new node" and differentiate that from an existing account:
https://github.com/microsoft/vscode-azureappservice/blob/master/src/commands/deploy.ts#L59

EricJizbaMSFT:
Closing due to lack of activity. Feel free to re-open once you're actively working on this

PR #379, opened 2019-06-27
Create directories in file explorer for file share

### LINKED ISSUE #371: [file share] Create new directory through file explorer

monalgarg:
Fixes #371


src/azureStorageExplorer/fileShares/FileShareFS.ts
```diff
@@ -66,8 +68,32 @@ export class FileShareFS implements vscode.FileSystemProvider {
...
-    createDirectory(_uri: vscode.Uri): void | Thenable<void> {
-        throw new Error("Method not implemented.");
+    async createDirectory(uri: vscode.Uri): Promise<void> {
+        return await callWithTelemetryAndErrorHandling('fs.createDirectory', async (context) => {
+            context.errorHandling.rethrow = true;
+
+            let parsedUri = parseUri(uri, this._fileShareString);
+            let root = await this.getRoot(context, uri);
+
+            if (root instanceof FileShareGroupTreeItem) {
```
EricJizbaMSFT: Can you modify `getRoot` to always return `FileShareTreeItem`? Seems unnecessary to check `instanceof FileShareGroupTreeItem` every time you call `getRoot` when it's never supported (right?)

src/azureStorageExplorer/fileShares/FileShareFS.ts
```diff
@@ -66,8 +68,32 @@ export class FileShareFS implements vscode.FileSystemProvider {
...
+                throw new Error('Cannot create a File Share directory.');
+            }
+
+            let response: string | undefined | null = validateDirectoryName(parsedUri.baseName);
+            if (response) {
+                throw new Error(response);
+            }
+
+            let fileShareTreeItem = <FileShareTreeItem>root;
+            await vscode.window.withProgress({ location: vscode.ProgressLocation.Window }, async (progress) => {
```
EricJizbaMSFT: We prefer `ProgressLocation.Notification` instead of `ProgressLocation.Window` (the `notification` option is newer and we just haven't moved all the old `window` stuff over yet)

src/azureStorageExplorer/fileShares/FileShareFS.ts
```diff
@@ -66,8 +68,32 @@ export class FileShareFS implements vscode.FileSystemProvider {
...
+                progress.report({ message: `Azure Storage: Creating directory ${parsedUri.filePath}` });
+                let dir = await createDirectory(fileShareTreeItem.share, root.root, parsedUri.parentDirPath, parsedUri.baseName);
```
EricJizbaMSFT: Assuming you move over to `ProgressLocation.Notification`, you can remove "Azure Storage: " from the message since it's already displayed as the "Source" in the notification

src/azureStorageExplorer/fileShares/FileShareFS.ts
```diff
@@ -66,8 +68,32 @@ export class FileShareFS implements vscode.FileSystemProvider {
...
+
+                // DirectoryResult.name contains the parent path in this call, but doesn't in other places such as listing directories.
+                // Remove it here to be consistent.
+                dir.name = parsedUri.baseName;
```
EricJizbaMSFT: I don't see you using `dir` anywhere, so you should be able to do just `await createDirectory(` instead of `let dir = await createDirectory(`

src/azureStorageExplorer/fileShares/FileShareFS.ts
EricJizbaMSFT: Also how long does it take to create a directory normally? If it's really fast then we might not need this

src/azureStorageExplorer/fileShares/FileShareFS.ts
```diff
@@ -208,10 +218,10 @@ export class FileShareFS implements vscode.FileSystemProvider {
...
 
-    private async getRoot(context: IActionContext, uri: vscode.Uri): Promise<FileShareTreeItem | FileShareGroupTreeItem> {
+    private async getRoot(context: IActionContext, uri: vscode.Uri): Promise<FileShareTreeItem> {
         let root = await findRoot(context, uri, this._fileShareString);
 
-        if (root instanceof FileShareTreeItem || root instanceof FileShareGroupTreeItem) {
+        if (root instanceof FileShareTreeItem) {
             return root;
         } else {
             throw new RangeError('The root found must be either a FileShareTreeItem or FileShareGroupTreeItem.');
```
EricJizbaMSFT: nit: Remove "or FileShareGroupTreeItem" from the error

PR #384, opened 2019-07-08
[blob] deleting virtual folders and files recursively

### LINKED ISSUE #375: [blob] Delete blobs through file explorer

monalgarg:
Fixes #375

### DIFF
--- a/src/azureStorageExplorer/blobContainers/BlobContainerFS.ts
+++ b/src/azureStorageExplorer/blobContainers/BlobContainerFS.ts
@@ -91,8 +91,54 @@ export class BlobContainerFS implements vscode.FileSystemProvider {
     }
 
     // tslint:disable-next-line: no-reserved-keywords
-    delete(_uri: vscode.Uri, _options: { recursive: boolean; }): void | Thenable<void> {
-        throw new Error("Method not implemented.");
+    async delete(uri: vscode.Uri, options: { recursive: boolean; }): Promise<void> {
+        return await callWithTelemetryAndErrorHandling('blob.delete', async (context) => {
+            context.errorHandling.rethrow = true;
+            if (!options.recursive) {
+                throw new Error('Do not support non recursive deletion of folders or files.');
+            }
+
+            let entry: EntryTreeItem = await this.lookup(uri);
+            let parsedUri = parseUri(uri, this._blobContainerString);
+
+            const blobService = entry.root.createBlobService();
+            if (entry instanceof BlobTreeItem) {
+                await this.deleteBlob(parsedUri.rootName, path.join(parsedUri.parentDirPath, parsedUri.baseName), blobService);
+            } else if (entry instanceof BlobDirectoryTreeItem) {
+                await this.deleteFolder(entry, blobService);
+            } else if (entry instanceof BlobContainerTreeItem) {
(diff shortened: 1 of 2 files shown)


src/azureStorageExplorer/blobContainers/BlobContainerFS.ts
```diff
@@ -91,8 +91,54 @@ export class BlobContainerFS implements vscode.FileSystemProvider {
...
+            } else if (entry instanceof BlobDirectoryTreeItem) {
+                await this.deleteFolder(entry, blobService);
+            } else if (entry instanceof BlobContainerTreeItem) {
+                throw new Error('Cannot delete a Blob Container.');
+            }
+        });
+    }
+
+    private async deleteFolder(entry: BlobDirectoryTreeItem, blobService: azureStorage.BlobService): Promise<void> {
+        let rootName = entry.container.name;
```
EricJizbaMSFT: It's usually better to do a loop instead of recursion to be more memory-effecient. For example, you could do something like this:
```
let dirPaths: string[] = [];
while (dirPath) {
    for (...) {
        await this.deleteBlob(<child blob>);
    }

    for (...) {
        dirPaths.push(<child dir>);
    }

    dirPath = dirPaths.pop();
}
```

src/azureStorageExplorer/blobContainers/BlobContainerFS.ts
```diff
@@ -91,8 +91,54 @@ export class BlobContainerFS implements vscode.FileSystemProvider {
     }
 
     // tslint:disable-next-line: no-reserved-keywords
-    delete(_uri: vscode.Uri, _options: { recursive: boolean; }): void | Thenable<void> {
-        throw new Error("Method not implemented.");
+    async delete(uri: vscode.Uri, options: { recursive: boolean; }): Promise<void> {
```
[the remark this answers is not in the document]
monalgarg: A popup comes us asking if the user is sure they want to delete the blob. So, I think the user experience is okay for blobs but I discovered it asks the user twice for deleting file share (so working on that).

src/azureStorageExplorer/blobContainers/BlobContainerFS.ts
EricJizbaMSFT: Your `Set` and `dirPaths.delete` pattern is pretty unconventional. Any reason you didn't use the suggested while loop pattern from above? Here it is more explicit (I coded it up real quick, but didn't fully test it so take it as you will):
```
private async deleteFolder(parsedUri: IParsedUri, blobService: azureStorage.BlobService): Promise<void> {
    let dirPaths: string[] = [];
    while (dirPath) {
        for (const blob of childBlob.entries) {
            await this.deleteBlob(parsedUri.rootName, blob.name, blobService);
        }

        let childDir = await this.listAllChildDirectory(blobService, parsedUri.rootName, dirPath);
        for (const dir of childDir.entries) {
            dirPaths.push(dir.name);
        }

        dirPath = dirPaths.pop();
    }
}
```

monalgarg: Oh oops, I thought your suggestion was something else. Fixing it now.

Issue #390, opened 2019-07-11
Saving blobs changes their content type to to octet-stream


fiveisprime:
Repro steps:

1. Deploy a static site (you can deploy just a basic HTML file with a <p> tag and some text)
1. Open the `$web` container in the file explorer
1. Edit the HTML file and save
1. Browse to the site (right-click container, browse to static website)

Expected:
The index.html renders with the edits

Actual:
The file is downloaded rather than rendered because the file's content type is set to `application/octet-stream`

Verified with Storage Explorer

![image](https://user-images.githubusercontent.com/1186948/61014922-89666600-a33e-11e9-9a4b-89c476c37074.png)

Issue #386, opened 2019-07-09
"Open in File Explorer" should be at the top


fiveisprime:
It's currently sorted alphabetically with the files so it's typically at the bottom and sometimes in the middle of the list. It should be the first option.

v-dantch:
Verified on the build [20200114.14](https://dev.azure.com/ms-azuretools/AzCode/_build/results?buildId=13766&view=artifacts&type=publishedArtifacts). The current context menu is like below:
![image](https://user-images.githubusercontent.com/34729092/72499780-c40d0600-386d-11ea-9fed-962309091b79.png)

PR #392, opened 2019-07-11
Create and delete virtual folders

### LINKED ISSUE #385: [blob] Create virtual folders for blobs

monalgarg:
Fixes #385 #390 #386

### DIFF
--- a/src/azureStorageExplorer/blobContainers/BlobContainerFS.ts
+++ b/src/azureStorageExplorer/blobContainers/BlobContainerFS.ts
@@ -18,7 +18,7 @@ export type EntryTreeItem = BlobTreeItem | BlobDirectoryTreeItem | BlobContainer
 export class BlobContainerFS implements vscode.FileSystemProvider {
 
     private _blobContainerString: string = 'Blob Containers';
-    private _virtualDirCreatedUri: vscode.Uri[] = [];
+    private _virtualDirCreatedUri: string[] = [];
 
     private _emitter: vscode.EventEmitter<vscode.FileChangeEvent[]> = new vscode.EventEmitter<vscode.FileChangeEvent[]>();
     readonly onDidChangeFile: vscode.Event<vscode.FileChangeEvent[]> = this._emitter.event;
@@ -28,7 +28,15 @@ export class BlobContainerFS implements vscode.FileSystemProvider {
     }
 
     async stat(uri: vscode.Uri): Promise<vscode.FileStat> {
+        if (this._virtualDirCreatedUri.indexOf(uri.path) > -1) {
+            return { type: vscode.FileType.Directory, ctime: 0, mtime: 0, size: 0 };
+        }
+
         let entry: EntryTreeItem = await this.lookup(uri);
+
+        // if (this._virtualDirCreatedUri.indexOf(uri) > -1) {
+        //     return { type: vscode.FileType.Directory, ctime: 0, mtime: 0, size: 0 };
+        // }
 
         if (entry instanceof BlobDirectoryTreeItem || entry instanceof BlobContainerTreeItem) {
             // creation and modification times as well as size of tree item are intentionally set to 0 for now
@@ -61,8 +69,8 @@ export class BlobContainerFS implements vscode.FileSystemProvider {
             directoryChildren.push([dirName, vscode.FileType.Directory]);
         }
         for (let dirCreated of this._virtualDirCreatedUri) {
-            if (dirCreated.path.includes(uri.path)) {
-                let dirName = path.basename(dirCreated.path);
+            if (dirCreated.includes(uri.path)) {
+                let dirName = path.basename(dirCreated);
                 directoryChildren.push([dirName, vscode.FileType.Directory]);
             }
         }
@@ -71,7 +79,7 @@ export class BlobContainerFS implements vscode.FileSystemProvider {
     }
 
     createDirectory(uri: vscode.Uri): void {
-        this._virtualDirCreatedUri.push(uri);
+        this._virtualDirCreatedUri.push(uri.path);
     }
 
     async readFile(uri: vscode.Uri): Promise<Uint8Array> {
--- a/src/azureStorageExplorer/blobContainers/BlobContainerFS.ts
+++ b/src/azureStorageExplorer/blobContainers/BlobContainerFS.ts
@@ -47,7 +47,6 @@ export class BlobContainerFS implements vscode.FileSystemProvider {
 
     async readDirectory(uri: vscode.Uri): Promise<[string, vscode.FileType][]> {
         return <[string, vscode.FileType][]>await callWithTelemetryAndErrorHandling('blob.readDirectory', async (context) => {
-            // let entry: BlobDirectoryTreeItem | BlobContainerTreeItem = await this.lookupAsDirectory(uri);
             let root: BlobContainerTreeItem = await this.getRoot(context, uri);
             let parsedUri = parseUri(uri, this._blobContainerString);
 
@@ -81,23 +80,25 @@ export class BlobContainerFS implements vscode.FileSystemProvider {
     }
 
     async readFile(uri: vscode.Uri): Promise<Uint8Array> {
-        let treeItem: BlobTreeItem = await this.lookupAsBlob(uri);
+        return <Uint8Array>await callWithTelemetryAndErrorHandling('blob.readFile', async (context) => {
+            let root: BlobContainerTreeItem = await this.getRoot(context, uri);
 
-        let parsedUri = parseUri(uri, this._blobContainerString);
+            let parsedUri = parseUri(uri, this._blobContainerString);
 
-        let blobSerivce: azureStorage.BlobService = treeItem.root.createBlobService();
-        const result = await new Promise<string | undefined>((resolve, reject) => {
-            blobSerivce.getBlobToText(parsedUri.rootName, parsedUri.filePath, (error?: Error, text?: string) => {
-                if (!!error) {
-                    reject(error);
-                } else {
-                    resolve(text);
-                }
+            let blobSerivce: azureStorage.BlobService = root.root.createBlobService();
+            const result = await new Promise<string | undefined>((resolve, reject) => {
+                blobSerivce.getBlobToText(parsedUri.rootName, parsedUri.filePath, (error?: Error, text?: string) => {
+                    if (!!error) {
+                        reject(error);
+                    } else {
+                        resolve(text);
+                    }
+                });
             });
-        });
 
-        // tslint:disable-next-line: strict-boolean-expressions
-        return Buffer.from(result || '');
+            // tslint:disable-next-line: strict-boolean-expressions
+            return Buffer.from(result || '');
+        });
     }
 
     async writeFile(uri: vscode.Uri, content: Uint8Array, options: { create: boolean; overwrite: boolean; }): Promise<void> {
@@ -151,21 +152,21 @@ export class BlobContainerFS implements vscode.FileSystemProvider {
         throw new Error("Method not implemented.");
     }
 
-    private async lookupAsBlob(uri: vscode.Uri): Promise<BlobTreeItem> {
-        const entry = await this.lookup(uri);
-        if (entry instanceof BlobTreeItem) {
-            return entry;
-        }
-        throw vscode.FileSystemError.FileIsADirectory(uri);
-    }
+    // private async lookupAsBlob(uri: vscode.Uri): Promise<BlobTreeItem> {
+    //     const entry = await this.lookup(uri);
+    //     if (entry instanceof BlobTreeItem) {
+    //         return entry;
+    //     }
+    //     throw vscode.FileSystemError.FileIsADirectory(uri);
+    // }
 
-    private async lookupAsDirectory(uri: vscode.Uri): Promise<BlobDirectoryTreeItem | BlobContainerTreeItem> {
-        let entry = await this.lookup(uri);
-        if (entry instanceof BlobDirectoryTreeItem || entry instanceof BlobContainerTreeItem) {
-            return entry;
-        }
-        throw vscode.FileSystemError.FileNotADirectory(uri);
-    }
+    // private async lookupAsDirectory(uri: vscode.Uri): Promise<BlobDirectoryTreeItem | BlobContainerTreeItem> {
+    //     let entry = await this.lookup(uri);
+    //     if (entry instanceof BlobDirectoryTreeItem || entry instanceof BlobContainerTreeItem) {
+    //         return entry;
+    //     }
+    //     throw vscode.FileSystemError.FileNotADirectory(uri);
+    // }
 
     private async lookup(uri: vscode.Uri): Promise<EntryTreeItem> {
         return <EntryTreeItem>await callWithTelemetryAndErrorHandling('blob.lookup', async (context) => {
--- a/src/azureStorageExplorer/blobContainers/BlobContainerFS.ts
+++ b/src/azureStorageExplorer/blobContainers/BlobContainerFS.ts
@@ -23,6 +23,9 @@ export class BlobContainerFS implements vscode.FileSystemProvider {
     private _emitter: vscode.EventEmitter<vscode.FileChangeEvent[]> = new vscode.EventEmitter<vscode.FileChangeEvent[]>();
     readonly onDidChangeFile: vscode.Event<vscode.FileChangeEvent[]> = this._emitter.event;
 
+    private _configUri: string[] = ['pom.xml', 'node_modules', '.vscode', '.vscode/settings.json', '.vscode/tasks.json', '.vscode/launch.json', '.git/config'];
+    private _configRootNames: string[] = ['pom.xml', 'node_modules', '.git'];
+
     watch(_uri: vscode.Uri, _options: { recursive: boolean; excludes: string[]; }): vscode.Disposable {
         throw new Error("Method not implemented.");
     }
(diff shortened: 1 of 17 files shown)


src/azureStorageExplorer/blobContainers/BlobContainerFS.ts
```diff
@@ -27,63 +28,92 @@ export class BlobContainerFS implements vscode.FileSystemProvider {
...
-        let entry: EntryTreeItem = await this.lookup(uri);
-
-        if (entry instanceof BlobDirectoryTreeItem || entry instanceof BlobContainerTreeItem) {
-            // creation and modification times as well as size of tree item are intentionally set to 0 for now
-            return { type: vscode.FileType.Directory, ctime: 0, mtime: 0, size: 0 };
-        } else if (entry instanceof BlobTreeItem) {
-            // creation and modification times as well as size of tree item are intentionally set to 0 for now
-            return { type: vscode.FileType.File, ctime: 0, mtime: 0, size: 0 };
-        }
+        return <vscode.FileStat>await callWithTelemetryAndErrorHandling('blob.readDirectory', async (context) => {
```
nturinski: Change the name of the event `blob.readDirectory`.

src/azureStorageExplorer/blobContainers/BlobContainerFS.ts
```diff
@@ -27,63 +28,92 @@ export class BlobContainerFS implements vscode.FileSystemProvider {
...
 
     async readDirectory(uri: vscode.Uri): Promise<[string, vscode.FileType][]> {
-        let entry: BlobDirectoryTreeItem | BlobContainerTreeItem = await this.lookupAsDirectory(uri);
+        return <[string, vscode.FileType][]>await callWithTelemetryAndErrorHandling('blob.readDirectory', async (context) => {
+            let root: BlobContainerTreeItem = await this.getRoot(context, uri);
+            let parsedUri = parseUri(uri, this._blobContainerString);
 
-        let parsedUri = parseUri(uri, this._blobContainerString);
+            if (parsedUri.filePath === '') {
+                console.log('hi');
```
nturinski: 🤣

src/azureStorageExplorer/blobContainers/BlobContainerFS.ts
```diff
@@ -27,63 +28,81 @@ export class BlobContainerFS implements vscode.FileSystemProvider {
...
     }
 
     async readDirectory(uri: vscode.Uri): Promise<[string, vscode.FileType][]> {
-        let entry: BlobDirectoryTreeItem | BlobContainerTreeItem = await this.lookupAsDirectory(uri);
+        return <[string, vscode.FileType][]>await callWithTelemetryAndErrorHandling('blob.readDirectory', async (context) => {
+            let root: BlobContainerTreeItem = await this.getRoot(context, uri);
+            let parsedUri = parseUri(uri, this._blobContainerString);
 
-        let parsedUri = parseUri(uri, this._blobContainerString);
+            const blobSerivce = root.root.createBlobService();
```
nturinski: Do you need to do root.root?  Isn't `this.getRoot` getting the root?

PrashanthCorp: blobSerivce -> blobService

src/azureStorageExplorer/blobContainers/BlobContainerFS.ts
```diff
@@ -27,63 +28,81 @@ export class BlobContainerFS implements vscode.FileSystemProvider {
...
-    createDirectory(_uri: vscode.Uri): void | Thenable<void> {
-        throw new Error("Method not implemented.");
+    createDirectory(uri: vscode.Uri): void {
+        this._virtualDirCreatedUri.add(uri.path);
     }
 
     async readFile(uri: vscode.Uri): Promise<Uint8Array> {
-        let treeItem: BlobTreeItem = await this.lookupAsBlob(uri);
+        return <Uint8Array>await callWithTelemetryAndErrorHandling('blob.readFile', async (context) => {
+            context.errorHandling.rethrow = true;
```
nturinski: Why do you want to rethrow and surpress?

src/azureStorageExplorer/blobContainers/BlobContainerFS.ts
```diff
@@ -27,63 +28,81 @@ export class BlobContainerFS implements vscode.FileSystemProvider {
...
 
-        let blobSerivce: azureStorage.BlobService = treeItem.root.createBlobService();
-        const result = await new Promise<string | undefined>((resolve, reject) => {
-            blobSerivce.getBlobToText(parsedUri.rootName, parsedUri.filePath, (error?: Error, text?: string) => {
-                if (!!error) {
-                    reject(error);
-                } else {
-                    resolve(text);
-                }
+            let blobSerivce: azureStorage.BlobService = root.root.createBlobService();
```
nturinski: Same with the root.root here

src/azureStorageExplorer/blobContainers/BlobContainerFS.ts
```diff
@@ -27,63 +28,81 @@ export class BlobContainerFS implements vscode.FileSystemProvider {
...
-            let dirName = path.basename(dirRes.name);
-            directoryChildren.push([dirName, vscode.FileType.Directory]);
-        }
-        return directoryChildren;
+                let parentPath = path.posix.join(dirCreatedParsedUri.rootPath, dirCreatedParsedUri.parentDirPath);
+                if (parentPath.endsWith("/")) {
+                    parentPath = parentPath.substring(0, parentPath.length - 1);
+                }
+
+                if (uri.path === parentPath) {
```
PrashanthCorp: For my understanding, is this condition expected to be satisfied only once? Given that `_virtualCreatedDirUri` is a set, is there a case where multiple `dirCreated`s can trigger this condition?

src/azureStorageExplorer/blobContainers/BlobContainerFS.ts
PrashanthCorp: blobSerivce -> blobService :D

src/azureStorageExplorer/blobContainers/BlobContainerFS.ts
```diff
@@ -27,63 +28,81 @@ export class BlobContainerFS implements vscode.FileSystemProvider {
...
-                }
+            let blobSerivce: azureStorage.BlobService = root.root.createBlobService();
+            let result = await new Promise<string | undefined>((resolve, reject) => {
+                blobSerivce.getBlobToText(parsedUri.rootName, parsedUri.filePath, (error?: Error, text?: string) => {
+                    if (!!error) {
+                        reject(error);
+                    } else {
+                        resolve(text);
+                    }
+                });
```
PrashanthCorp: Is it just git diff being weird, or has something changed here that I'm missing?

src/azureStorageExplorer/blobContainers/BlobContainerFS.ts
```diff
@@ -112,14 +131,32 @@ export class BlobContainerFS implements vscode.FileSystemProvider {
                 throw vscode.FileSystemError.FileExists(uri);
             } else {
                 await new Promise<void>((resolve, reject) => {
-                    blobSerivce.createBlockBlobFromText(parsedUri.rootName, parsedUri.filePath, content.toString(), (error?: Error) => {
+                    let contentType: string | null = mime.getType(parsedUri.filePath);
+                    let temp: string | undefined = contentType === null ? undefined : contentType;
+                    blobSerivce.createBlockBlobFromText(parsedUri.rootName, parsedUri.filePath, content.toString(), { contentSettings: { contentType: temp } }, (error?: Error) => {
```
PrashanthCorp: Are these 3 lines the fix to #390? I didn't see a specific commit referencing the fix.

src/azureStorageExplorer/blobContainers/BlobContainerFS.ts
```diff
@@ -183,49 +231,31 @@ export class BlobContainerFS implements vscode.FileSystemProvider {
...
-        throw vscode.FileSystemError.FileNotADirectory(uri);
-    }
 
-    private async lookup(uri: vscode.Uri): Promise<EntryTreeItem> {
-        return <EntryTreeItem>await callWithTelemetryAndErrorHandling('blob.lookup', async (context) => {
-            context.errorHandling.rethrow = true;
-            context.errorHandling.suppressDisplay = true;
-
-            let parsedUri = parseUri(uri, this._blobContainerString);
+        let blobSerivce = entry.root.createBlobService();
```
PrashanthCorp: Same typo.

src/azureStorageExplorer/blobContainers/BlobContainerFS.ts
```diff
@@ -228,15 +239,18 @@ export class BlobContainerFS implements vscode.FileSystemProvider {
     }
 
     rename(_oldUri: vscode.Uri, _newUri: vscode.Uri, _options: { overwrite: boolean; }): void | Thenable<void> {
-        throw new Error("Method not implemented.");
+        throw new Error("Do not support renaming/moving folders or files.");
```
PrashanthCorp: Nit: "Renaming/moving folders or files not supported"

The current wording sounds like we're asking the user not to support it. :)

src/azureStorageExplorer/blobContainers/BlobContainerFS.ts
```diff
@@ -132,19 +169,30 @@ export class BlobContainerFS implements vscode.FileSystemProvider {
                 throw new Error('Do not support non recursive deletion of folders or files.');
             }
 
-            let entry: EntryTreeItem = await this.lookup(uri);
             let parsedUri = parseUri(uri, this._blobContainerString);
+            try {
```
nturinski: With try/catch, you normally want to only put the command that you expect to fail in the try block.  For instance, right now, this line `throw new Error('Cannot delete a Blob Container.');` will never get displayed to the user because the catch block doesn't rethrow the error.  I'm assuming you only expect the lookup to fail if the uri is virtual?

src/azureStorageExplorer/blobContainers/BlobContainerFS.ts
```diff
@@ -255,7 +285,7 @@ export class BlobContainerFS implements vscode.FileSystemProvider {
         return await new Promise<azureStorage.BlobService.ListBlobsResult>((resolve, reject) => {
             // Intentionally passing undefined for token - only supports listing first batch of files for now
             // tslint:disable-next-line: no-non-null-assertion
-            let options = { delimiter: '/' };
+            let options = { maxResults: 500, delimiter: '/' };
```
nturinski: Nice, so I assume that this limits the results to only have 500 blobs?  Though what if there are more?  Is there a load more that's implemented already?

src/azureStorageExplorer/blobContainers/BlobContainerFS.ts
```diff
@@ -183,49 +231,31 @@ export class BlobContainerFS implements vscode.FileSystemProvider {
         throw new Error("Method not implemented.");
```
nturinski: This should probably get deleted now that it's implemented, right?

src/azureStorageExplorer/blobContainers/BlobContainerFS.ts
monalgarg: oh i think this is probably confusing because of the naming. But, get root just gets the blobcontainertreeitem but I need an IStorageRoot object which is stored in the root property of the blobcontainertreeitem.

src/azureStorageExplorer/blobContainers/BlobContainerFS.ts
monalgarg: This can be fulfilled multiple times. For example, suppose I have a folder a. Then, I create a virtual a/b, a/c. So, this statement would be satisfied for both those uri strings.

src/azureStorageExplorer/blobContainers/BlobContainerFS.ts
monalgarg: I think it shows up like this because I surrounded these statements with a callWithTelemetryAndErrorHandling

src/azureStorageExplorer/blobContainers/BlobContainerFS.ts
monalgarg: yes they are for #390

src/azureStorageExplorer/blobContainers/BlobContainerFS.ts
monalgarg: Oops, I accidentally left that in while I was trying something. I haven't actually implemented it fully yet.

src/azureStorageExplorer/blobContainers/BlobContainerFS.ts
nturinski: Ahh, gotcha.  Yeah, I think the naming convention is kind of confusing.  I don't think you should change `getRoot`, but maybe you should change the variable name, `root`?

src/azureStorageExplorer/blobContainers/BlobContainerFS.ts
monalgarg: These changes are under the commit "content type being updated" 74404008eb721a338f53643ea2cfe720784f976f.

src/azureStorageExplorer/blobContainers/BlobContainerFS.ts
PrashanthCorp: Ah.

src/azureStorageExplorer/fileShares/FileShareFS.ts
```diff
@@ -27,6 +26,9 @@ export class FileShareFS implements vscode.FileSystemProvider {
     private _emitter: vscode.EventEmitter<vscode.FileChangeEvent[]> = new vscode.EventEmitter<vscode.FileChangeEvent[]>();
     readonly onDidChangeFile: vscode.Event<vscode.FileChangeEvent[]> = this._emitter.event;
 
+    private _configUri: string[] = ['pom.xml', 'node_modules', '.vscode', '.vscode/settings.json', '.vscode/tasks.json', '.vscode/launch.json', '.git/config'];
```
nturinski: Write a comment clarifying what these are for.

src/azureStorageExplorer/blobContainers/BlobContainerFS.ts
```diff
@@ -22,6 +22,9 @@ export class BlobContainerFS implements vscode.FileSystemProvider {
 
     private _emitter: vscode.EventEmitter<vscode.FileChangeEvent[]> = new vscode.EventEmitter<vscode.FileChangeEvent[]>();
     readonly onDidChangeFile: vscode.Event<vscode.FileChangeEvent[]> = this._emitter.event;
+    // private _listener: vscode.Disposable = this.onDidChangeFile((e => {
```
nturinski: Haha, you should probably get rid of this. (Sorry about writing this 😓)

src/extension.ts
```diff
@@ -72,12 +73,18 @@ export async function activateInternal(context: vscode.ExtensionContext, perfSta
             context.subscriptions.push(vscode.workspace.registerFileSystemProvider('azurestorageblob', new BlobContainerFS(), { isCaseSensitive: true }));
         }
         registerCommand('azureStorage.openFileShareInFileExplorer', async (_actionContext: IActionContext, treeItem: FileShareTreeItem) => {
-            // tslint:disable-next-line: prefer-template
-            await commands.executeCommand('vscode.openFolder', vscode.Uri.parse('azurestoragefile://' + treeItem.fullId));
+            await callWithTelemetryAndErrorHandling('blob.openInFileExplorer', async () => {
```
nturinski: Should the command be `file.openInFileExplorer`?

src/azureStorageExplorer/fileShares/fileShareNode.ts
```diff
@@ -56,19 +58,34 @@ export class FileShareTreeItem extends AzureParentTreeItem<IStorageRoot> impleme
...
+            return -1;
+        } else if (ti2.label === this._openInFileExplorerString) {
+            return 1;
+        }
+
+        return ti1.label.localeCompare(ti2.label);
+    }
+
+    public async refreshImpl(): Promise<void> {
+        this._openedInFileExplorer = false;
```
nturinski: Why are you always setting this to false when you refresh the FileShareTreeItem?

src/azureStorageExplorer/blobContainers/blobContainerNode.ts
```diff
@@ -79,24 +81,36 @@ export class BlobContainerTreeItem extends AzureParentTreeItem<IStorageRoot> imp
...
+        }
+
+        return ti1.label.localeCompare(ti2.label);
+    }
+
     public async refreshImpl(): Promise<void> {
         //tslint:disable-next-line:no-non-null-assertion
         const hostingStatus = await (<StorageAccountTreeItem>this!.parent!.parent).getActualWebsiteHostingStatus();
         this._websiteHostingEnabled = hostingStatus.enabled;
+        this._openedInFileExplorer = false;
```
nturinski: Same here.

src/azureStorageExplorer/blobContainers/blobContainerNode.ts
```diff
@@ -79,24 +81,36 @@ export class BlobContainerTreeItem extends AzureParentTreeItem<IStorageRoot> imp
...
             const ti = new GenericTreeItem(this, {
-                label: 'Open in File Explorer...',
+                label: this._openInFileExplorerString,
                 commandId: 'azureStorage.openBlobContainerInFileExplorer',
                 contextValue: 'openBlobContainerInFileExplorer'
             });
 
             ti.commandArgs = [this];
             result.push(ti);
+            this._openedInFileExplorer = true;
```
nturinski: I'm a little confused.  Shouldn't this only be set to true if the user clicked on it?

monalgarg: Oops, the name may be misleading. I want this boolean basically to indicate whether or not this file has been opened in azure storage and if the Open in File Explorer option has been already displayed.

Issue #344, opened 2019-05-17
Deployment experience improvements


fiveisprime:
- [ ] 	Automatically configure predeploy task (or clearly document this)
- [ ] Add support for deploy subpath
- [ ] Improve feedback when creating a storage account (long running notifications for creation, I'll spec this out)
- [ ] Add support for `defaultWebAppToDeploy` similar to how App Service does this today
- [x] Basic creation mode
    - [x] Create resource group in West US
    - [x] Create set index.html for index and error documents when user chooses to enable static hosting

burkeholland:
Suggestion: Add commands (both to the palette and right-click) for "Create Static Website".

Right now the user has to know to pick "Configure Static Website", which seems like a weird choice if you don't already have one. Otherwise, they have to right-click and create a new Storage Account, know to select Blob and then know to do the Configure step.

>   * [ ]  Create set index.html for index and error documents when user chooses to enable static hosting

What about changing the placeholder text to "Set to index.html if you have a single page application".

PrashanthCorp:
> - [ ] 	Automatically configure predeploy task (or clearly document this)
We could leverage the task discovery that vscode uses in `Tasks: Config default build task`.

johnpapa:
Here are some thoughts on the reduction of prompts, based on my chat with Matt recently

The experience with Azure Storage was this:

- I started with my account, so login was already there. 
- in vscode extension I right clicked my dist folder, chose "push to storage"
- was prompted for a storage account (created new one)
- prompted for a resource group (created new one)
- was asked if I wanted to enable web site (chose yes)
- was prompted for default page (typed index.html )
- was prompted for error page (typed 404.html )
- was prompted to browse to the site (took about 15 seconds before I got this prompt)

https://papapeacockstorage.z13.web.core.windows.net/

Overall fast. But some prompts feel extraneous (index.html could be assumed with fallbacks of the other common ones, IMO)

Total time was about 2 minutes.

pros - the site is fast, it didnt take long to get it running at all

cons - seems like we could default even more of these prompts for storage, resource group, default page, error page. I also already had a login ... so I imagine the time to get this working is much longer if I had to sign up from scratch.

Super nice to have ... I think it would be great if the name of the url is cleaner and more fun with such as dancing-penguins-xxxx.azurewebsites.net

PR #393, opened 2019-07-19
 Storage account create : Add basic creation; set index and Error404DocPath

PrashanthCorp:
Addresses #344 's basic create mode.

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,6 +62,7 @@
 		"onCommand:azureStorage.deleteDirectory",
 		"onCommand:azureStorage.createTextFile",
 		"onCommand:azureStorage.createGpv2Account",
+		"onCommand:azureStorage.basicCreateGpv2Account",
 		"onCommand:azureStorage.deleteFile",
 		"onCommand:azureStorage.createDirectory",
 		"onCommand:azureStorage.createSubdirectory",
@@ -257,6 +258,11 @@
 			{
 				"command": "azureStorage.createGpv2Account",
 				"title": "Create Storage Account",
+				"category": "Azure Storage"
+			},
+			{
+				"command": "azureStorage.basicCreateGpv2Account",
+				"title": "Quick Create Storage Account",
 				"category": "Azure Storage"
 			},
 			{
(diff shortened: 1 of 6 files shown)


PrashanthCorp:
Deferring to @fiveisprime for whether we follow the trend.

PrashanthCorp:
IIRC 

old -> new

create -> advanced create
what I added -> create

package.json
```diff
@@ -643,11 +643,11 @@
 					"when": "never"
 				},
 				{
-					"command": "azureStorage.createGpv2Account",
+					"command": "azureStorage.advancedCreateGpv2Account",
 					"when": "never"
 				},
 				{
-					"command": "azureStorage.basicCreateGpv2Account",
+					"command": "azureStorage.createGpv2Account",
```
nturinski: I know it's not part of this PR, but is there any reason that these commands don't support command palette? It doesn't seem like it'd be very difficult to support.

package.json
```diff
@@ -469,12 +469,12 @@
 					"group": "2_create@1"
 				},
 				{
-					"command": "azureStorage.createGpv2Account",
+					"command": "azureStorage.advancedCreateGpv2Account",
```
nturinski: Change the ordering of the commands by altering the group label.  Advanced create should be below the normal create.

package.json
```diff
@@ -256,13 +256,13 @@
 				"category": "Azure Storage"
 			},
 			{
-				"command": "azureStorage.createGpv2Account",
-				"title": "Create Storage Account",
+				"command": "azureStorage.advancedCreateGpv2Account",
+				"title": "Advanced Create Storage Account",
```
nturinski: Follow the pattern of the other extensions for the title:
Create Storage Account... (Advanced)

In general, any title that requires the user to add more inputs should end with ...

Also, this is a nit, but Functions and App Service name the command createGpv2AccountAdvanced (has it as a suffix rather than a prefix)

EricJizbaMSFT:

PR #395, opened 2019-07-20
Try to configure pre-deploy task to 'build'.

PrashanthCorp:
Address #344

### DIFF
--- a/src/azureStorageExplorer/storageAccounts/storageAccountActionHandlers.ts
+++ b/src/azureStorageExplorer/storageAccounts/storageAccountActionHandlers.ts
@@ -89,10 +89,10 @@ async function deployStaticWebsite(context: IActionContext, target?: vscode.Uri 
         throw new Error(`Could not find $web blob container for storage account "${destAccountTreeItem.label}"`);
     }
 
-    const hasAsked = <boolean>vscode.workspace.getConfiguration(extensionPrefix).get("requestPredeploy", false);
+    const hasAsked = <boolean>vscode.workspace.getConfiguration(extensionPrefix).get("requestPreDeploy", false);
     if (!hasAsked) {
         await configurePredeployTask(sourcePath);
-        vscode.workspace.getConfiguration(extensionPrefix).update("requestPredeploy", true, vscode.ConfigurationTarget.Workspace);
+        vscode.workspace.getConfiguration(extensionPrefix).update("requestPreDeploy", true, vscode.ConfigurationTarget.Workspace);
     }
 
     await runPreDeployTask(sourcePath, context);
--- a/src/azureStorageExplorer/storageAccounts/storageAccountActionHandlers.ts
+++ b/src/azureStorageExplorer/storageAccounts/storageAccountActionHandlers.ts
@@ -89,9 +89,24 @@ async function deployStaticWebsite(context: IActionContext, target?: vscode.Uri 
         throw new Error(`Could not find $web blob container for storage account "${destAccountTreeItem.label}"`);
     }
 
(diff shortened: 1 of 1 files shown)


src/azureStorageExplorer/storageAccounts/storageAccountActionHandlers.ts
```diff
@@ -89,11 +89,26 @@ async function deployStaticWebsite(context: IActionContext, target?: vscode.Uri
         throw new Error(`Could not find $web blob container for storage account "${destAccountTreeItem.label}"`);
     }
 
+    const hasAsked = <boolean>vscode.workspace.getConfiguration(extensionPrefix).get("requestPredeploy", false);
```
nturinski: You need to define this setting in the package.json.

src/azureStorageExplorer/storageAccounts/storageAccountActionHandlers.ts
```diff
@@ -89,11 +89,26 @@ async function deployStaticWebsite(context: IActionContext, target?: vscode.Uri
...
+        await configurePredeployTask(sourcePath);
+        vscode.workspace.getConfiguration(extensionPrefix).update("requestPredeploy", true, vscode.ConfigurationTarget.Workspace);
+    }
+
     await runPreDeployTask(sourcePath, context);
 
     return destContainerTreeItem.deployStaticWebsite(context, sourcePath);
 }
 
+async function configurePredeployTask(deployFsPath: string): Promise<void> {
```
nturinski: nit: Casing on PreDeploy

src/azureStorageExplorer/storageAccounts/storageAccountActionHandlers.ts
```diff
@@ -89,11 +89,26 @@ async function deployStaticWebsite(context: IActionContext, target?: vscode.Uri
...
+    const tasks: vscode.Task[] = await vscode.tasks.fetchTasks();
+    if (tasks.find((task: vscode.Task) => isTaskEqual("build", deployFsPath, task))) {
+        vscode.workspace.getConfiguration(extensionPrefix, vscode.Uri.file(deployFsPath)).update(configurationSettingsKeys.preDeployTask, "build");
+    } else {
+        vscode.window.showInformationMessage(`Please set your predeploy task under the setting "${extensionPrefix}.${configurationSettingsKeys.preDeployTask}".`);
```
nturinski: This error is hard to understand.  

The issue here is that they need the "build" `task` for it to be set as the preDeployTask `setting`, right?  

The wording of this error sounds like the issue is that the `setting` isn't set rather than the `task`.

Maybe something like:

> Unable to find "build" in tasks.json.  Create a "build" task that will be set as...

src/azureStorageExplorer/storageAccounts/storageAccountActionHandlers.ts
```diff
@@ -89,11 +89,26 @@ async function deployStaticWebsite(context: IActionContext, target?: vscode.Uri
         throw new Error(`Could not find $web blob container for storage account "${destAccountTreeItem.label}"`);
     }
 
+    const hasAsked = <boolean>vscode.workspace.getConfiguration(extensionPrefix).get("requestPredeploy", false);
+    if (!hasAsked) {
```
nturinski: Why is this called `hasAsked`?  It doesn't seem like the user has to interact with the prompt at all.

It also seems weird to me that this is only shown once and then never again whether the preDeploy task is set or not.

EricJizbaMSFT:

PR #404, opened 2019-07-23
Checking for permissions when deleting blobs

### LINKED ISSUE #403: Handle deletion of blobs with snapshots

monalgarg:
Fixes #403 

Blobs with special permissions (such as those with snapshots or with leases) cannot be deleted from the file explorer feature. If a user tries to delete a single blob, BlobContainerFS will rethrow the error and the UI will not change. However, if a user tries to delete a folder, which may or may not contain blobs that cannot be deleted, BlobContainerFS handles this error. Since vscode is not handling this error, the file explorer is updated as if the whole folder was deleted. The user must then refresh to see the true state.

### DIFF
--- a/src/azureStorageExplorer/blobContainers/BlobContainerFS.ts
+++ b/src/azureStorageExplorer/blobContainers/BlobContainerFS.ts
@@ -4,6 +4,7 @@
  *--------------------------------------------------------------------------------------------*/
 
 import * as azureStorage from "azure-storage";
+import * as mime from "mime";
 import * as path from 'path';
 import * as vscode from 'vscode';
 import { callWithTelemetryAndErrorHandling, IActionContext } from "vscode-azureextensionui";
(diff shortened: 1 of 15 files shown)


src/azureStorageExplorer/blobContainers/BlobContainerFS.ts
```diff
@@ -228,18 +241,29 @@ export class BlobContainerFS implements vscode.FileSystemProvider {
...
+                    if (!!error) {
+                        reject(error);
+                    } else {
+                        resolve();
+                    }
+                });
             });
-        });
+        } catch (error) {
+            const pe = parseError(error);
```
EricJizbaMSFT: per discussion in standup - you should handle _all_ errors the same way instead of singling out snapshot errors

src/azureStorageExplorer/blobContainers/BlobContainerFS.ts
```diff
@@ -199,27 +200,50 @@ export class BlobContainerFS implements vscode.FileSystemProvider {
...
+            } else if (entry instanceof BlobDirectoryTreeItem) {
+                await vscode.window.withProgress({ location: vscode.ProgressLocation.Notification }, async (progress) => {
                     progress.report({ message: `Deleting directory ${parsedUri.filePath}` });
-                    await this.deleteFolder(parsedUri, blobService);
-                } else if (entry instanceof BlobContainerTreeItem) {
-                    throw new Error('Cannot delete a Blob Container.');
-                }
-            });
+                    let deletionErrors: blobDeletionErrors = await this.deleteFolder(parsedUri, blobService);
+                    if (deletionErrors.snapshotDetected && deletionErrors.leaseDetected) {
```
EricJizbaMSFT: I don't think the slightly different messages here are worth the added complexity. Just handle all errors the same way with a generic message

src/azureStorageExplorer/blobContainers/BlobContainerFS.ts
EricJizbaMSFT: You're special-casing two errors here and ignoring a whole bunch of other possible errors. Instead, we want you to add generic logic to handle _all_ errors

src/azureStorageExplorer/blobContainers/BlobContainerFS.ts
```diff
@@ -199,27 +200,50 @@ export class BlobContainerFS implements vscode.FileSystemProvider {
...
-            await vscode.window.withProgress({ location: vscode.ProgressLocation.Notification }, async (progress) => {
-                if (entry instanceof BlobTreeItem) {
+
+            if (entry instanceof BlobTreeItem) {
+                await vscode.window.withProgress({ location: vscode.ProgressLocation.Notification }, async (progress) => {
                     progress.report({ message: `Deleting blob ${parsedUri.filePath}` });
-                    await this.deleteBlob(parsedUri.rootName, parsedUri.filePath, blobService);
-                } else if (entry instanceof BlobDirectoryTreeItem) {
+                    let deletionErrors: blobDeletionErrors = await this.deleteBlob(parsedUri.rootName, parsedUri.filePath, blobService);
+                    if (deletionErrors.snapshotDetected) {
```
EricJizbaMSFT: I thought you were going to give users an option to delete snapshots if you ran into a snapshot error for a single blob? If not, I re-iterate the same thing I said below where I don't think slightly different messages is worth the added complexity of code

src/azureStorageExplorer/blobContainers/BlobContainerFS.ts
```diff
@@ -229,18 +252,29 @@ export class BlobContainerFS implements vscode.FileSystemProvider {
...
+                blobService.deleteBlob(containerName, prefix, (error?: Error) => {
+                    if (!!error) {
+                        reject(error);
+                    } else {
+                        resolve();
+                    }
+                });
             });
-        });
+        } catch (error) {
```
EricJizbaMSFT: I think you should get rid of the try/catch in this method. It doesn't seem to provide much value when you could just add a try/catch above. It's also pretty unconventional to check `if (error instanceof Error)`. We want to handle all types of errors and sometimes they get returned as other types (like a string)

src/azureStorageExplorer/blobContainers/BlobContainerFS.ts
```diff
@@ -199,27 +198,51 @@ export class BlobContainerFS implements vscode.FileSystemProvider {
...
+    private async deleteFolder(parsedUri: IParsedUri, blobService: azureStorage.BlobService): Promise<Error[]> {
         let dirPaths: string[] = [];
         let dirPath: string | undefined = parsedUri.dirPath;
+
+        let blobDeletionError: Error[] = [];
         while (dirPath) {
             let childBlob = await this.listAllChildBlob(blobService, parsedUri.rootName, dirPath);
             for (const blob of childBlob.entries) {
-                await this.deleteBlob(parsedUri.rootName, blob.name, blobService);
+                let error: Error | undefined = await this.deleteBlob(parsedUri.rootName, blob.name, blobService);
```
EricJizbaMSFT: This is where the try catch should be. Also - you should print out the errors to `ext.outputChannel` as they happen so that the user can follow along if they want.

src/azureStorageExplorer/blobContainers/BlobContainerFS.ts
```diff
@@ -199,27 +198,51 @@ export class BlobContainerFS implements vscode.FileSystemProvider {
...
-                    throw new Error('Cannot delete a Blob Container.');
-                }
-            });
+                    let deletionErrors: Error[] = await this.deleteFolder(parsedUri, blobService);
+                    if (deletionErrors.length > 0) {
+                        context.errorHandling.rethrow = false;
+                        context.errorHandling.suppressDisplay = false;
+                        let message: string = '';
+                        deletionErrors.forEach(element => {
+                            message = `${message}${element.message}\n`;
```
EricJizbaMSFT: As mentioned elsewhere, you should print errors to the output channel. It's much easier to read and then you can do it as they happen instead of waiting for all deletes to finish

src/azureStorageExplorer/blobContainers/BlobContainerFS.ts
```diff
@@ -199,27 +198,51 @@ export class BlobContainerFS implements vscode.FileSystemProvider {
...
+            if (entry instanceof BlobTreeItem) {
+                await vscode.window.withProgress({ location: vscode.ProgressLocation.Notification }, async (progress) => {
                     progress.report({ message: `Deleting blob ${parsedUri.filePath}` });
-                    await this.deleteBlob(parsedUri.rootName, parsedUri.filePath, blobService);
-                } else if (entry instanceof BlobDirectoryTreeItem) {
+                    let deletionErrors: Error | undefined = await this.deleteBlob(parsedUri.rootName, parsedUri.filePath, blobService);
+                    if (deletionErrors) {
+                        context.errorHandling.rethrow = true;
+                        context.errorHandling.suppressDisplay = true;
+                        throw vscode.FileSystemError.NoPermissions(`${parseError(deletionErrors).message}`);
```
EricJizbaMSFT: If you remove the try/catch from `this.deleteBlob`, I think you can just let the original error bubble up as-is. Aka I don't think this one needs a try/catch or `FileSystemError.NoPermissions`

src/azureStorageExplorer/blobContainers/BlobContainerFS.ts
monalgarg: Should I print the errors out to ext.outputChannel in addition to showing them to the user by throwing a No Permissions error?

src/azureStorageExplorer/blobContainers/BlobContainerFS.ts
```diff
@@ -216,27 +217,37 @@ export class BlobContainerFS implements vscode.FileSystemProvider {
...
         let dirPath: string | undefined = parsedUri.dirPath;
+
         while (dirPath) {
             let childBlob = await this.listAllChildBlob(blobService, parsedUri.rootName, dirPath);
             for (const blob of childBlob.entries) {
-                await this.deleteBlob(parsedUri.rootName, blob.name, blobService);
+                try {
+                    await this.deleteBlob(parsedUri.rootName, blob.name, blobService);
+                } catch (error) {
+                    ext.outputChannel.appendLine(`Cannot delete ${blob.name}. ${parseError(error).message}`);
```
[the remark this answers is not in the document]
monalgarg: How do I give them a button that directs them to the output channel?

PR #405, opened 2019-07-23
Improving performance of file explorer view of azure storage files

### LINKED ISSUE #364: [file share] Improve lookup method to scan all children not just the first batch of children

monalgarg:
In order to improve the performance of viewing Azure Storage FIle Share files, FileShareFS uses the caches already set up in FileShareTreeItem and DirectoryTreeItem.  

Fixes #364


EricJizbaMSFT:
Seems like this PR is still being worked on? (based on the amount of commented-out code and weird names like `readDirectory2`, etc.) Let me know when it's ready for review.

src/azureStorageExplorer/fileShares/FileShareFS.ts
```diff
@@ -31,6 +30,10 @@ export class FileShareFS implements vscode.FileSystemProvider {
 
     async stat(uri: vscode.Uri): Promise<vscode.FileStat> {
         return await callWithTelemetryAndErrorHandling('fs.stat', async (context) => {
+            if (uri.path.endsWith('/') || uri.path.endsWith('Load More...')) {
```
EricJizbaMSFT: I thought we decided not to do the "Load More" file because it was too wonky - and just let users do "Load More" from the azure view as a workaround? Even if we decide to add "Load More" I'd prefer to have that as a separate PR since it's a feature by itself

src/azureStorageExplorer/fileShares/FileShareFS.ts
```diff
@@ -45,45 +48,52 @@ export class FileShareFS implements vscode.FileSystemProvider {
...
-            let listFilesandDirectoryResult = await entry.listFiles(<azureStorage.common.ContinuationToken>undefined!);
-            let entries = listFilesandDirectoryResult.entries;
-
-            let result: [string, vscode.FileType][] = [];
-            for (const dir of entries.directories) {
-                result.push([dir.name, vscode.FileType.Directory]);
-            }
-            for (const file of entries.files) {
-                result.push([file.name, vscode.FileType.File]);
+    putChildrenIntoCache(entry: DirectoryTreeItem | FileShareTreeItem, children: AzExtTreeItem[]): [string, vscode.FileType][] {
```
EricJizbaMSFT: Seems like you did a minor refactor to put this logic in its own method, but the method name doesn't feel accurate to me. It's not really putting anything in the cache. I think I liked it better before

src/azureStorageExplorer/fileShares/fileShareNode.ts
```diff
@@ -120,13 +120,28 @@ export class FileShareTreeItem extends AzureParentTreeItem<IStorageRoot> impleme
...
     public async createChildImpl(context: ICreateChildImplContext & IFileShareCreateChildContext): Promise<DirectoryTreeItem | FileTreeItem> {
         if (context.childType === FileTreeItem.contextValue) {
+            if (context.childName) {
+                const file = await createFile('', context.childName, this.share, this.root);
+                const actualFile = await getFile('', file.name, this.share, this.root);
+                return new FileTreeItem(this, actualFile, '', this.share);
+            }
             return askAndCreateEmptyTextFile(this, '', this.share, context);
         } else {
+            if (context.childName) {
```
EricJizbaMSFT: These changes seem to duplicate some of the work in `askAndCreateChildDirectory`. I would prefer if you use that function directly and the only difference is you check `if (context.childName)` before prompting and skip only that part.

src/azureStorageExplorer/fileShares/FileShareFS.ts
```diff
@@ -219,28 +260,19 @@ export class FileShareFS implements vscode.FileSystemProvider {
...
-            let listFilesAndDirectoriesResult: azureStorage.FileService.ListFilesAndDirectoriesResult = await entry.listFiles(<azureStorage.common.ContinuationToken>undefined!);
-            let entries = listFilesAndDirectoriesResult.entries;
-
-            let directoryResultChild = entries.directories.find(element => element.name === part);
-            if (!!directoryResultChild) {
-                entry = new DirectoryTreeItem(entry, parentPath, directoryResultChild, <azureStorage.FileService.ShareResult>entry.share);
-                parentPath = path.posix.join(parentPath, part);
+
+            let temp: AzExtTreeItem[] = await entry.getCachedChildren(context);
+            let childFound: AzExtTreeItem | undefined = temp.find(element => element.label === part);
```
EricJizbaMSFT: It's generally advised to use an id instead of a label when looking for something. `label` is meant to be human readable and could change based on things like a user's language (although it probably won't in this case). `id` is always meant to be the identifying property. Also see if you can leverage `ext.tree.findTreeItem(`

src/azureStorageExplorer/blobContainers/BlobContainerFS.ts
```diff
@@ -90,6 +97,13 @@ export class BlobContainerFS implements vscode.FileSystemProvider {
         await callWithTelemetryAndErrorHandling('blob.createDirectory', async (context) => {
             context.errorHandling.rethrow = true;
 
+            let parsedUri = parseUri(uri, this._blobContainerString);
+
+            // tslint:disable-next-line: no-multiline-string
+            if (parsedUri.baseName.includes(`\\`) || parsedUri.baseName.includes(`\/`) || parsedUri.baseName === "") {
+                throw new Error('Name must not be empty or contain slashes.');
```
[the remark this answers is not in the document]
monalgarg: parsedUri.baseName should never have a '/' in it (I tested this using regex101). So, the check is kind of unnecessary but I thought it just made the code easier to read? I can take it out though.

src/azureStorageExplorer/blobContainers/BlobContainerFS.ts
```diff
@@ -31,7 +31,9 @@ export class BlobContainerFS implements vscode.FileSystemProvider {
 
     async stat(uri: vscode.Uri): Promise<vscode.FileStat> {
         return await callWithTelemetryAndErrorHandling('blob.stat', async (context) => {
-            if (this._virtualDirCreatedUri.has(uri.path)) {
+            if (uri.path.endsWith('/')) {
```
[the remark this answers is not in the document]
monalgarg: VS code calls this the FSP API without any slashes at the end of a uri (regardless of whether its a folder or file). When reading a directory, VS Code calls read directory and then based off the children names will call stat. Since in read directory I have the file type information, I thought I could put it in the uri. Since '/' is a special character and never appears at the end of a uri, I thought I could use the '/' to quickly identify the file type. But, you can only add one '/' to the end of the uri, so either the '/' could be an indicator of a file or directory. I can see why this looks pretty odd and can remove it or modify it with another special character potentially?

src/azureStorageExplorer/blobContainers/BlobContainerFS.ts
```diff
@@ -69,12 +71,17 @@ export class BlobContainerFS implements vscode.FileSystemProvider {
             }
 
             let directoryChildren: [string, vscode.FileType][] = [];
-            listBlobResult.entries.forEach(value => directoryChildren.push([path.basename(value.name), vscode.FileType.File]));
-            listDirectoryResult.entries.forEach(value => directoryChildren.push([path.basename(value.name), vscode.FileType.Directory]));
+            listBlobResult.entries.forEach(value => {
+                let baseName: string = path.basename(value.name);
+                directoryChildren.push([`${baseName}/`, vscode.FileType.File]);
```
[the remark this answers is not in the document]
monalgarg: It is so that I can quickly identify its a file in stat.

src/azureStorageExplorer/blobContainers/BlobContainerFS.ts
```diff
@@ -177,9 +199,7 @@ export class BlobContainerFS implements vscode.FileSystemProvider {
                     }
 
                     let fullPath: string = path.posix.join(parsedUri.rootPath, parentDirPath);
-                    if (this._virtualDirCreatedUri.has(fullPath)) {
-                        this._virtualDirCreatedUri.delete(fullPath);
-                    } else {
+                    if (!this._virtualDirCreatedUri.delete(fullPath)) {
```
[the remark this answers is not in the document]
monalgarg: If a virtual folder is populated, it is no longer a "virtual folder" and doesn't need to be a part of this cache. To be honest, I think it probably is fine if the cache contains folders that contain blobs. But, I thought cleaning the cache would be helpful.

src/azureStorageExplorer/fileShares/FileShareFS.ts
monalgarg: Sorry, I forgot to update you on this. So, I think the object on which I call getCachedChildren may not be the same one that the Storage Explorer extension is working with. So, as a result, the caches are not linked to each other.

But, I can move the Load More into another PR.

src/azureStorageExplorer/blobContainers/BlobContainerFS.ts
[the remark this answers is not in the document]
EricJizbaMSFT: Yeah I think it adds code complexity (aka risk for bugs) without much value. I would remove this logic (although maybe a separate PR since it's not really related to perf)

src/azureStorageExplorer/blobContainers/BlobContainerFS.ts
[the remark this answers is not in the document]
EricJizbaMSFT: Ah - I think using a special character here is pretty confusing. Also - assuming `readDirectory` is properly caching stuff, I don't think this will have any performance benefits

src/azureStorageExplorer/blobContainers/BlobContainerFS.ts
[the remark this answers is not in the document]
EricJizbaMSFT: Yeah I would take it out

src/azureStorageExplorer/fileShares/FileShareFS.ts
EricJizbaMSFT: We should make sure the caches are linked. Let me know if you need help with that

src/azureStorageExplorer/blobContainers/BlobContainerFS.ts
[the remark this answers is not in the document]
EricJizbaMSFT: Still waiting on this

src/azureStorageExplorer/blobContainers/BlobContainerFS.ts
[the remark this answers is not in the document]
EricJizbaMSFT: Still waiting on this

src/azureStorageExplorer/blobContainers/BlobContainerFS.ts
```diff
@@ -226,55 +224,45 @@ export class BlobContainerFS implements vscode.FileSystemProvider {
...
+                } else if (entry instanceof BlobDirectoryTreeItem) {
                     progress.report({ message: `Deleting directory ${parsedUri.filePath}` });
-                    let errors: boolean = await this.deleteFolder(parsedUri, blobService);
-
-                    if (errors) {
-                        vscode.window.showInformationMessage(`Errors occured when deleting ${parsedUri.filePath}. Please look at the output channel for more information.`);
-                        // tslint:disable-next-line: no-multiline-string
-                        ext.outputChannel.appendLine(`Please refresh the viewlet to see the changes made.`);
+                    if (await this.deleteFolder(parsedUri, blobService)) {
+                        vscode.window.showInformationMessage(`Errors occured when deleting ${parsedUri.filePath}. Please refresh the viewlet and refer to the output channel for more information.`);
```
EricJizbaMSFT: Revert these unrelated changes. It would conflict with your other PR anyways

src/azureStorageExplorer/blobContainers/BlobContainerFS.ts
```diff
@@ -226,55 +224,45 @@ export class BlobContainerFS implements vscode.FileSystemProvider {
...
-        let dirPaths: string[] = [];
         let dirPath: string | undefined = parsedUri.dirPath;
-
+        let dirPaths: string[] = [];
         let errors: boolean = false;
 
+        ext.outputChannel.show();
         while (dirPath) {
             let childBlob = await this.listAllChildBlob(blobService, parsedUri.rootName, dirPath);
-            for (const blob of childBlob.entries) {
```
EricJizbaMSFT: A standard for loop is preferred over `forEach` because it preserves the current scope. Take this example:

Screen Shot 2019-08-07 at 9 21 50 AM

There's a squiggle under `x` in the `forEach` with the error "Object is possibly 'undefined'." because typescript no longer knows that x is defined.

src/azureStorageExplorer/fileShares/FileShareFS.ts
```diff
@@ -46,45 +44,43 @@ export class FileShareFS implements vscode.FileSystemProvider {
     }
 
     async readDirectory(uri: vscode.Uri): Promise<[string, vscode.FileType][]> {
-        return await callWithTelemetryAndErrorHandling('fs.readDirectory', async (context) => {
```
EricJizbaMSFT: Revert this change because it hides bugs when you cast a value like this

src/azureStorageExplorer/fileShares/FileShareFS.ts
```diff
@@ -143,27 +141,42 @@ export class FileShareFS implements vscode.FileSystemProvider {
                 });
             });
 
-            if (!fileResultChild.exists && !options.create) {
+            let createNewFile: boolean = !fileResultChild.exists;
+            if (fileResultChild.exists && content.length === 0) {
+                throw new Error(`A file or folder ${parsedUri.baseName} already exists at this location. Please choose a different name`);
```
EricJizbaMSFT: Isn't this already handled when throwing a FileExists error below? And what does content length have to do with this?

src/azureStorageExplorer/fileShares/FileShareFS.ts
```diff
@@ -125,12 +121,14 @@ export class FileShareFS implements vscode.FileSystemProvider {
 
     async writeFile(uri: vscode.Uri, content: Uint8Array, options: { create: boolean; overwrite: boolean; }): Promise<void> {
         await callWithTelemetryAndErrorHandling('fs.writeFile', async (context) => {
+            context.errorHandling.rethrow = true;
```
EricJizbaMSFT: Why did this change to rethrow?

src/azureStorageExplorer/fileShares/FileShareFS.ts
```diff
@@ -202,56 +211,30 @@ export class FileShareFS implements vscode.FileSystemProvider {
...
-        if (root instanceof FileShareTreeItem) {
-            return root;
+    private async lookup(uri: vscode.Uri | string, context: IActionContext): Promise<EntryTreeItem> {
+        let uriString = uri instanceof vscode.Uri ? uri.path : uri;
+        let ti = await ext.tree.findTreeItem(uriString, context);
+        if (ti instanceof FileShareTreeItem || ti instanceof FileTreeItem || ti instanceof DirectoryTreeItem) {
+            return ti;
         } else {
-            throw new RangeError(`Unexpected entry ${root.constructor.name}.`);
+            throw getFileSystemError(uri, context, vscode.FileSystemError.FileNotFound);
```
EricJizbaMSFT: This should only happen if ti is undefined. Otherwise you should throw a range error

src/azureStorageExplorer/fileShares/directoryNode.ts
```diff
@@ -77,16 +77,36 @@ export class DirectoryTreeItem extends AzureParentTreeItem<IStorageRoot> impleme
 
     public async createChildImpl(context: ICreateChildImplContext & IFileShareCreateChildContext): Promise<FileTreeItem | DirectoryTreeItem> {
         if (context.childType === FileTreeItem.contextValue) {
+            if (context.childName) {
+                const file = await createFile(this.fullPath, context.childName, this.share, this.root);
+                const actualFile = await getFile(this.fullPath, file.name, this.share, this.root);
+                return new FileTreeItem(this, actualFile, this.fullPath, this.share);
```
EricJizbaMSFT: There's too much duplicate code here. This should be more tightly integrated with `askAndCreateEmptyTextFile`

EricJizbaMSFT:
You can try doing what I did here (except for file shares) to verify performance: https://github.com/microsoft/vscode-azurestorage/issues/433

Just add a console.log similar to this every time `listFilesAndDirectoriesSegmented` is called before and after your changes:
```
console.log(`${new Date().toLocaleTimeString()}: Querying Azure... Method: listBlobDirectoriesSegmentedWithPrefix directory: "${directory}" share: "${share}"`);
```

src/azureStorageExplorer/fileShares/FileShareFS.ts
monalgarg: It is to differentiate when writing to a file that already exists vs. creating a new file that has the same name as another file. When creating a new file, the content.length is set to 0. But, I didn't consider the case when someone is writing to a file and deletes all of its content...

src/azureStorageExplorer/fileShares/directoryUtils.ts
```diff
@@ -10,21 +10,26 @@ import { AzureParentTreeItem, ICreateChildImplContext, UserCancelledError } from
...
 import { validateDirectoryName } from "./validateNames";
 
 // Supports both file share and directory parents
-export async function askAndCreateChildDirectory(parent: AzureParentTreeItem<IStorageRoot>, parentPath: string, share: azureStorage.FileService.ShareResult, context: ICreateChildImplContext): Promise<DirectoryTreeItem> {
-    const dirName = await window.showInputBox({
-        placeHolder: 'Enter a name for the new directory',
-        validateInput: validateDirectoryName
-    });
+export async function askAndCreateChildDirectory(parent: AzureParentTreeItem<IStorageRoot>, parentPath: string, share: azureStorage.FileService.ShareResult, context: ICreateChildImplContext & IFileShareCreateChildContext): Promise<DirectoryTreeItem> {
+    let dirName: string | undefined = context.childName;
```
EricJizbaMSFT: How about:

```
const dirName: string = context.childName || await window.showInputBox({
    placeHolder: 'Enter a name for the new directory',
    validateInput: validateDirectoryName
});
```

Then you won't have to do this below:
```
let dirNameString: string = <string>dirName;
```

src/azureStorageExplorer/parseUri.ts
```diff
@@ -48,7 +48,7 @@ export interface IParsedUri {
 
 export function parseUri(uri: vscode.Uri | string, fileType: string): IParsedUri {
     let path: string = uri instanceof vscode.Uri ? uri.path : uri;
-    const matches: RegExpMatchArray | null = path.match(`^(\/subscriptions\/[^\/]+\/resourceGroups\/[^\/]+\/providers\/Microsoft\.Storage\/storageAccounts\/[^\/]+\/${fileType}\/([^\/]+))\/?((.*?\/?)([^\/]*))$`);
+    const matches: RegExpMatchArray | null = path.match(`^(\/subscriptions\/[^\/]+\/resourceGroups\/[^\/]+\/providers\/Microsoft\.Storage\/storageAccounts\/[^\/]+\/${fileType}\/([^\/]+))\/?((.*?\/?)([^\/]*))\/?$`);
```
EricJizbaMSFT: What's this change for?

src/azureStorageExplorer/findRoot.ts
```diff
@@ -9,20 +9,12 @@ import { ext } from '../extensionVariables';
...
-    } else {
-        root = await ext.tree.findTreeItem(rootPath, context);
-        if (!root) {
-            throw getFileSystemError(uri, context, vscode.FileSystemError.FileNotFound);
-        } else {
-            rootMap.set(rootPath, root);
-            return root;
-        }
+    let root = await ext.tree.findTreeItem(rootPath, context);
+    if (!root) {
```
EricJizbaMSFT: Should we just get rid of this file? Seems pretty simple now and I believe it's only used by blob file provider at this point

src/azureStorageExplorer/fileShares/FileShareFS.ts
```diff
@@ -143,27 +140,40 @@ export class FileShareFS implements vscode.FileSystemProvider {
...
-                                resolve();
+                        let parentUri: vscode.Uri = vscode.Uri.file(path.posix.join(parsedUri.rootPath, parsedUri.parentDirPath));
+                        let parent = await this.lookupAsDirectory(parentUri, context);
+
+                        try {
+                            await parent.createChild(<IFileShareCreateChildContext>{ ...context, childType: 'azureFile', childName: parsedUri.baseName });
+                        } catch (error) {
+                            let pe = parseError(error);
+                            if (pe.errorType === 'ResourceTypeMismatch') {
+                                throw new Error(`A file or folder ${parsedUri.baseName} already exists at this location. Please choose a different name`);
```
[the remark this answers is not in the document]
monalgarg: If I try to create a file with the same name as a folder that is not included within this batch, the following error shows up. I thought this error wasn't clear so I threw this error.

Screen Shot 2019-08-07 at 5 43 53 PM

src/azureStorageExplorer/parseUri.ts
monalgarg: It was when I was adding '/' at the end to identify a folder, but not needed anymore.

src/azureStorageExplorer/blobContainers/BlobContainerFS.ts
```diff
@@ -246,16 +239,15 @@ export class BlobContainerFS implements vscode.FileSystemProvider {
...
     }
 
     private async deleteFolder(parsedUri: IParsedUri, blobService: azureStorage.BlobService): Promise<boolean> {
-        let dirPaths: string[] = [];
         let dirPath: string | undefined = parsedUri.dirPath;
-
+        let dirPaths: string[] = [];
         let errors: boolean = false;
 
+        ext.outputChannel.show();
```
EricJizbaMSFT: Revert this - unrelated to performance and users have complained in the past about output channel popping up too often. We don't even print anything unless there's an error

src/azureStorageExplorer/fileShares/FileShareFS.ts
```diff
@@ -123,14 +119,14 @@ export class FileShareFS implements vscode.FileSystemProvider {
         }) || Buffer.from('');
     }
 
-    async writeFile(uri: vscode.Uri, content: Uint8Array, options: { create: boolean; overwrite: boolean; }): Promise<void> {
+    async writeFile(uri: vscode.Uri, content: Uint8Array, options: { create: boolean; overwrite: boolean; accessConditions: boolean }): Promise<void> {
```
EricJizbaMSFT: Revert accessConditions

src/azureStorageExplorer/fileShares/fileShareNode.ts
```diff
@@ -91,7 +91,7 @@ export class FileShareTreeItem extends AzureParentTreeItem<IStorageRoot> impleme
     listFiles(currentToken: azureStorage.common.ContinuationToken): Promise<azureStorage.FileService.ListFilesAndDirectoriesResult> {
         return new Promise((resolve, reject) => {
             let fileService = this.root.createFileService();
-            fileService.listFilesAndDirectoriesSegmented(this.share.name, '', currentToken, { maxResults: 50 }, (err?: Error, result?: azureStorage.FileService.ListFilesAndDirectoriesResult) => {
+            fileService.listFilesAndDirectoriesSegmented(this.share.name, '', currentToken, { maxResults: 2 }, (err?: Error, result?: azureStorage.FileService.ListFilesAndDirectoriesResult) => {
```
EricJizbaMSFT: Revert maxResults change

src/azureStorageExplorer/blobContainers/BlobContainerFS.ts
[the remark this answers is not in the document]
EricJizbaMSFT: ping again

PR #406, opened 2019-07-24
Leverage latest UI and dev packages

EricJizbaMSFT:
Depends on https://github.com/microsoft/vscode-azuretools/pull/541

I moved `Test*` classes to the dev package instead of ui package

### DIFF
--- a/src/azureStorageExplorer/AzureAccountTreeItem.ts
+++ b/src/azureStorageExplorer/AzureAccountTreeItem.ts
@@ -3,11 +3,11 @@
  *  Licensed under the MIT License. See License.md in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import { AzureAccountTreeItemBase, ISubscriptionContext, TestAzureAccount } from "vscode-azureextensionui";
+import { AzureAccountTreeItemBase, ISubscriptionContext } from "vscode-azureextensionui";
 import { SubscriptionTreeItem } from "./SubscriptionTreeItem";
 
 export class AzureAccountTreeItem extends AzureAccountTreeItemBase {
-    public constructor(testAccount?: TestAzureAccount) {
+    public constructor(testAccount?: {}) {
         super(undefined, testAccount);
     }
     public createSubscriptionTreeItem(root: ISubscriptionContext): SubscriptionTreeItem {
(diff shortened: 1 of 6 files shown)


src/azureStorageExplorer/AzureAccountTreeItem.ts
```diff
@@ -3,11 +3,11 @@
  *  Licensed under the MIT License. See License.md in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import { AzureAccountTreeItemBase, ISubscriptionContext, TestAzureAccount } from "vscode-azureextensionui";
+import { AzureAccountTreeItemBase, ISubscriptionContext } from "vscode-azureextensionui";
 import { SubscriptionTreeItem } from "./SubscriptionTreeItem";
 
 export class AzureAccountTreeItem extends AzureAccountTreeItemBase {
-    public constructor(testAccount?: TestAzureAccount) {
+    public constructor(testAccount?: {}) {
```
nturinski: Why do you have to use an empty object rather than just using AzureAccount?

test/storageAccountActions.test.ts
```diff
@@ -29,16 +28,15 @@ suite('Storage Account Actions', async function (this: ISuiteCallbackContext): P
 
         ext.azureAccountTreeItem = new AzureAccountTreeItem(testAccount);
         ext.tree = new AzExtTreeDataProvider(ext.azureAccountTreeItem, 'azureStorage.loadMore');
-        storageAccountClient = getStorageManagementClient(testAccount);
```
nturinski: Why move this out of suiteSetup since it gets used multiple times?

src/azureStorageExplorer/AzureAccountTreeItem.ts
EricJizbaMSFT: It's not an empty object - it's a _type_ of object. I don't want to define the type for TestAzureAccount in the ui package

nturinski: But how come we can't use AzureAccount here since TestAzureAccount implements AzureAccount?

EricJizbaMSFT: I also don't want to expose the types for AzureAccount in the ui package. Those are defined in the azure account extension

test/storageAccountActions.test.ts
EricJizbaMSFT: Meh I didn't think about it too much. I mostly wanted to get rid of `getStorageManagementClient` because one-line functions annoy me and it was probably easier to copy-paste or something. I can change that aspect back

PR #414, opened 2019-07-31
Add copy access keys test cases

v-etzha:
1. Add Copy Connection String test case.
2. Add Copy Primary Key test case.


test/storageAccountActions.test.ts
```diff
@@ -60,9 +62,33 @@ suite('Storage Account Actions', async function (this: ISuiteCallbackContext): P
...
 
+    test("copyConnectionString", async () => {
+        const createdAccount: StorageAccount = await client.storageAccounts.getProperties(resourceName, resourceName);
+        assert.ok(createdAccount);
+        clipboardy.writeSync('');
+        await testUserInput.runWithInputs([resourceName], async () => {
+            await vscode.commands.executeCommand('azureStorage.copyConnectionString');
+        });
+        const connectionString: string = clipboardy.readSync();
+        const blobService: BlobService = createBlobService(connectionString);
```
nturinski: So I don't think `createBlobService` will verify whether or not the connectionString is valid.  Since this is synchronous, it isn't making an API call to Azure to verify that the connection string worked.  You will have to try running a command that uses the blob service and see if it returns the results that you expect.

test/storageAccountActions.test.ts
```diff
@@ -60,14 +63,50 @@ suite('Storage Account Actions', async function (this: ISuiteCallbackContext): P
...
+        const createdAccount: StorageAccount = await client.storageAccounts.getProperties(resourceName, resourceName);
+        assert.ok(createdAccount);
         await testUserInput.runWithInputs([resourceName, DialogResponses.deleteResponse.title], async () => {
             await vscode.commands.executeCommand('azureStorage.deleteStorageAccount');
         });
         await assertThrowsAsync(async () => await client.storageAccounts.getProperties(resourceName, resourceName), /Error/);
     });
+
+    async function validateBlobContainer(blobService: BlobService): Promise<void> {
+        const containerName: string = getRandomHexString().toLowerCase();
```
nturinski: Please add this comment here
`// storage account must have lower case name`

test/storageAccountActions.test.ts
```diff
@@ -60,14 +63,50 @@ suite('Storage Account Actions', async function (this: ISuiteCallbackContext): P
...
+        await new Promise((resolve, reject): void => {
```
nturinski: What happens if there is an error at `createContainerIfNotExists`?  Does the whole test exit because an error is thrown?

Issue #420, opened 2019-08-01
Update the error message when moving folders or files


v-dantch:
**OS:** Windows_NT x64 10.0.17763
**Build Version:** [20190731.14](https://dev.azure.com/ms-azuretools/AzCode/_build/results?buildId=8307)
**VS Code Version**: 1.37.0-insider (user setup)
**Commit:** d5900c642b359fe7a9eb1e3f9e2d0977f6e0905c
**Regression:** Not a Regression

**Repro Steps:**
1. Enable "Azure Storage: Enable View in File Explorer" in User Setting.
2. Drill down from one storage account --> "Blob Containers" node -->Expand one blob container --> Click "Open in File Explorer...".
3. Create a new folder and file--> Move the file to the folder.

**Expect:**
Pop up an error like "Moving folders or files not supported".

**Actual:**
Pop up below error:
![image](https://user-images.githubusercontent.com/34729092/62271404-63882a80-b3ed-11e9-9e4d-acba19ff8457.png)

PR #421, opened 2019-08-01
Differentiate between rename and move in File Share and Blob Container files

### LINKED ISSUE #419: No error pops up when renaming a file/folder for AzureStorageBlob/ AzureStorageFile
**OS:** Windows_NT x64 10.0.17763
**Build Version:** [20190731.14](https://dev.azure.com/ms-azuretools/AzCode/_build/results?buildId=8307)
**VS Code Version**: 1.37.0-insider (user setup)
**Commit:** d5900c642b359fe7a9eb1e3f9e2d0977f6e0905c
**Regression:** Previous build ([20190730.9](https://dev.azure.com/ms-azuretools/AzCode/_build/results?buildId=8267))

**Repro Steps:**
2. Drill down from one storage account --> "Blob Containers" node -->Expand one blob container --> Click "Open in File Explorer...".
3. Select one file/folder --> F2 --> Type a new name --> Enter.

**Expect:**
Pop up an error "Renaming folders or files not supported".

**Actual:**
No error pops up when executing a **un**-supported action.

**More Info:**
The name hasn't been renamed actually.

monalgarg:
Fixes #419 #420 

The FileShareProvider deals with errors thrown by renaming and moving file/folders differently. To accommodate this, I differentiate between rename and move by comparing the basename and throw errors differently. 

Known: If I rename a folder/file such that the folder/file would be moved to a different location, no error message is shown. For example, if there exists a folder and file b at the same location, if I rename file b to be a/b, the UI will not change and the file's location will remain the same but no error message will be shown. For now, I will leave this as is.

### DIFF
--- a/src/azureStorageExplorer/blobContainers/BlobContainerFS.ts
+++ b/src/azureStorageExplorer/blobContainers/BlobContainerFS.ts
@@ -12,6 +12,7 @@ import { ext } from "../../extensionVariables";
 import { findRoot } from "../findRoot";
 import { getFileSystemError } from "../getFileSystemError";
 import { IParsedUri, parseUri } from "../parseUri";
+import { showRenameError } from "../showRenameError";
 import { BlobContainerTreeItem } from './blobContainerNode';
 import { BlobDirectoryTreeItem } from "./BlobDirectoryTreeItem";
 import { BlobTreeItem } from './blobNode';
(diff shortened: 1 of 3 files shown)


src/azureStorageExplorer/blobContainers/BlobContainerFS.ts
```diff
@@ -282,11 +282,18 @@ export class BlobContainerFS implements vscode.FileSystemProvider {
...
+    async rename(oldUri: vscode.Uri, newUri: vscode.Uri, _options: { overwrite: boolean; }): Promise<void> {
         return await callWithTelemetryAndErrorHandling('blob.rename', async (context) => {
+            let oldUriParsed = parseUri(oldUri, this._blobContainerString);
+            let newUriParsed = parseUri(newUri, this._blobContainerString);
+
             context.errorHandling.rethrow = true;
-            context.errorHandling.suppressDisplay = true;
-            throw new Error('Renaming/moving folders or files not supported.');
+            if (oldUriParsed.baseName === newUriParsed.baseName) {
+                context.errorHandling.suppressDisplay = true;
```
EricJizbaMSFT: Add a comment (in the code) why suppressDisplay is true for moving, but not renaming. We talked about it offline, but it's not clear just from looking at the code.

src/azureStorageExplorer/fileShares/FileShareFS.ts
```diff
@@ -195,11 +195,18 @@ export class FileShareFS implements vscode.FileSystemProvider {
         });
     }
 
-    async rename(_oldUri: vscode.Uri, _newUri: vscode.Uri, _options: { overwrite: boolean; }): Promise<void> {
+    async rename(oldUri: vscode.Uri, newUri: vscode.Uri, _options: { overwrite: boolean; }): Promise<void> {
         return await callWithTelemetryAndErrorHandling('fs.rename', async (context) => {
+            let oldUriParsed = parseUri(oldUri, this._fileShareString);
```
EricJizbaMSFT: All of this is duplicate - can you have one function for both file systems. Maybe `showRenameError` or something

PR #423, opened 2019-08-05
Update shared packages to fix icons

EricJizbaMSFT:
See https://github.com/microsoft/vscode-azuretools/pull/550 for more info

Breaking change in dev package doesn't affect this extension

### DIFF
--- a/src/extensionVariables.ts
+++ b/src/extensionVariables.ts
@@ -15,6 +15,8 @@ export namespace ext {
     export let outputChannel: OutputChannel;
     export let ui: IAzureUserInput;
     export let reporter: ITelemetryReporter;
+    // tslint:disable-next-line: strict-boolean-expressions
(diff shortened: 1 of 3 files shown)


src/extensionVariables.ts
```diff
@@ -15,6 +15,8 @@ export namespace ext {
     export let outputChannel: OutputChannel;
     export let ui: IAzureUserInput;
     export let reporter: ITelemetryReporter;
+    // tslint:disable-next-line: strict-boolean-expressions
+    export let ignoreBundle: boolean = !/^(false|0)?$/i.test(process.env.AZCODE_STORAGE_IGNORE_BUNDLE || '');
```
PrashanthCorp: Where is this being used?

EricJizbaMSFT: https://github.com/microsoft/vscode-azuretools/pull/551/files#diff-8f1846e49e9b035ba2d2fa57d750fabeL21

PR #430, opened 2019-08-07
Remove feature flag

### LINKED ISSUE #351: Remove feature flag for FileSystemProvider to display blob containers/file shares
If we use VS Code's [FileSystemProvider](https://code.visualstudio.com/api/references/vscode-api#FileSystemProvider) api, it has the following benefits:
1. Users will automatically get their normal icons
1. Users will automatically get their normal sorting (folders on top, then files)
1. We won't have to write files to a temp folder in order to save and update

Here's VS Code's example: https://github.com/microsoft/vscode-extension-samples/tree/master/fsprovider-sample

Also related to https://github.com/microsoft/vscode-azuretools/issues/177, where I did a prototype for the App Service extension.

monalgarg:
Fixes #351


EricJizbaMSFT:
Closing until the feature is ready

PR #443, opened 2019-08-09
[file, blob] Updating telemetry

### LINKED ISSUE #365: [file share] Improve telemetry on FileShareFS.ts
Get telemetry for all methods from the FileSystemProvider API.

monalgarg:
Fixes #365 

Collecting telemetry on actions that are done from the context menu (creating, deleting, or renaming files/folders). 

Updating Azure Tools versions to use the telemetry.suppressAll for stat, readDirectory and readFile.


src/azureStorageExplorer/fileShares/FileShareFS.ts
```diff
@@ -30,6 +30,7 @@ export class FileShareFS implements vscode.FileSystemProvider {
 
     async stat(uri: vscode.Uri): Promise<vscode.FileStat> {
         return await callWithTelemetryAndErrorHandling('fs.stat', async (context) => {
+            context.telemetry.suppressAll = true;
```
EricJizbaMSFT: I'm not sure we want to suppress _all_ errors from these - maybe just the _expected_ errors. For example "File Not Found"

monalgarg: Should I suppress all telemetry though for stat?

EricJizbaMSFT: I would probably do suppressIfSuccessful by default and suppressAll if you know the error is expected.

We want telemetry on unexpected errors so that we can see it in telemetry and fix it. For example, if users are getting a bunch of RangeErrors - then we want to know ASAP and fix it.

src/azureStorageExplorer/blobContainers/BlobContainerFS.ts
```diff
@@ -49,6 +51,7 @@ export class BlobContainerFS implements vscode.FileSystemProvider {
 
     async readDirectory(uri: vscode.Uri): Promise<[string, vscode.FileType][]> {
         return await callWithTelemetryAndErrorHandling('blob.readDirectory', async (context) => {
+            context.telemetry.suppressAll = true;
```
EricJizbaMSFT: Same comment for readDirectory and readFile as stat. We still want to track unexpected errors

PR #461, opened 2019-09-12
Add create blob container test case

v-etzha:
1. Add create blob container test case and update vscode.windows to ext.ui.
2. Add a common method to simplify the code.

### DIFF
--- a/test/storageAccountActions.test.ts
+++ b/test/storageAccountActions.test.ts
@@ -64,8 +64,7 @@ suite('Storage Account Actions', async function (this: ISuiteCallbackContext): P
     });
 
     test("copyConnectionString", async () => {
-        const createdAccount: StorageAccount = await client.storageAccounts.getProperties(resourceName, resourceName);
-        assert.ok(createdAccount);
+        await validateAccountExists(resourceName, resourceName);
         clipboardy.writeSync('');
         await testUserInput.runWithInputs([resourceName], async () => {
             await vscode.commands.executeCommand('azureStorage.copyConnectionString');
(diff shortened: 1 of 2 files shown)


test/storageAccountActions.test.ts
```diff
@@ -87,15 +85,29 @@ suite('Storage Account Actions', async function (this: ISuiteCallbackContext): P
         await validateBlobService(blobService);
     });
 
+    test("createBlobContainer", async () => {
+        await validateAccountExists(resourceName, resourceName);
+        await testUserInput.runWithInputs([resourceName, resourceName], async () => {
```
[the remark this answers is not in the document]
v-etzha: Hi Eric, thanks for the comments! Updated the code to use a different blob container name.

test/storageAccountActions.test.ts
```diff
@@ -87,15 +85,31 @@ suite('Storage Account Actions', async function (this: ISuiteCallbackContext): P
...
     });
 
+    test("createBlobContainer", async () => {
+        await validateAccountExists(resourceName, resourceName);
+        // Blob contaienr must have lower case name
+        const containerName = getRandomHexString().toLowerCase();
+        await testUserInput.runWithInputs([resourceName, containerName], async () => {
+            await vscode.commands.executeCommand('azureStorage.createBlobContainer');
+        });
+        const createdContainer: BlobContainer = await client.blobContainers.get(resourceName, resourceName, resourceName);
```
EricJizbaMSFT: I believe you have to reference `containerName` here as well otherwise it'll fail to get the container

v-etzha: Thanks for the comments! Updated the resourceName to containerName.

PR #462, opened 2019-09-12
FSExplorer: Clear cache when user explicitly refreshes explorer

### LINKED ISSUE #413: Refresh button doesn't clear the cache
**OS:** Win10
**Build Version:** [20190730.9](https://dev.azure.com/ms-azuretools/AzCode/_build/results?buildId=8267)
**Regression**: Not a Regression

**Repro Steps:**
2. Drill down from one storage account --> "Blob Containers" node -->Expand one blob container -->Click "Open in File Explorer...".
3. Refresh the File Explorer.

**Expect:**
The File Explorer refreshes well with obvious indication.

**Actual:**
There is no any indication.

PrashanthCorp:
Fixes #413

Looking at the bug and flow of code, it looks like VSCode calls stat only on refreshes, extension opens, or fileSystemProvider registration. Given the FileSystemProvider API, it looks like we would need to keep track of when to return cached results, and when we should do a re-fetch. 

For blobs, there's already a policy of explicitly refreshing at every `readDirectory` call.

### DIFF
--- a/src/azureStorageExplorer/fileShares/FileShareFS.ts
+++ b/src/azureStorageExplorer/fileShares/FileShareFS.ts
@@ -23,6 +23,7 @@ export class FileShareFS implements vscode.FileSystemProvider {
 
     private _emitter: vscode.EventEmitter<vscode.FileChangeEvent[]> = new vscode.EventEmitter<vscode.FileChangeEvent[]>();
     readonly onDidChangeFile: vscode.Event<vscode.FileChangeEvent[]> = this._emitter.event;
+    private _syncRequired: Map<vscode.Uri, boolean> = new Map<vscode.Uri, boolean>();
 
     watch(_uri: vscode.Uri, _options: { recursive: boolean; excludes: string[]; }): vscode.Disposable {
         throw new Error("Method not implemented.");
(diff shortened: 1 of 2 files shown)


src/azureStorageExplorer/fileShares/FileShareFS.ts
```diff
@@ -48,7 +50,13 @@ export class FileShareFS implements vscode.FileSystemProvider {
         return await callWithTelemetryAndErrorHandling('fs.readDirectory', async (context) => {
             context.telemetry.suppressIfSuccessful = true;
             let entry: DirectoryTreeItem | FileShareTreeItem = await this.lookupAsDirectory(uri, context);
-            let children: AzExtTreeItem[] = await entry.getCachedChildren(context);
+            const shouldRefresh: boolean = this._syncRequired.get(uri) || true;
+            let children: AzExtTreeItem[];
+            if (shouldRefresh) {
```
EricJizbaMSFT: > For blobs, there's already a policy of explicitly refreshing at every readDirectory call.

I feel like we should be consistent either way. The blob option seems simpler - is there any downside to that? In other words, what's the benefit to your `_syncRequired` map?

EricJizbaMSFT:

PR #466, opened 2019-09-24
Add create file share test case

v-etzha:
Add create file share test case.

### DIFF
--- a/test/storageAccountActions.test.ts
+++ b/test/storageAccountActions.test.ts
@@ -7,7 +7,7 @@ import * as assert from 'assert';
 import { ResourceManagementClient } from 'azure-arm-resource';
 import { StorageManagementClient } from 'azure-arm-storage';
 import { BlobContainer, StorageAccount } from 'azure-arm-storage/lib/models';
-import { BlobService, createBlobService } from 'azure-storage';
+import { BlobService, createBlobService, createFileService, FileService } from 'azure-storage';
 import { IHookCallbackContext, ISuiteCallbackContext } from 'mocha';
 import * as vscode from 'vscode';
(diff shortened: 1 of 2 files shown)


test/storageAccountActions.test.ts
```diff
@@ -95,6 +91,28 @@ suite('Storage Account Actions', async function (this: ISuiteCallbackContext): P
...
+    test("createFileShare", async () => {
+        await validateAccountExists(resourceName, resourceName);
+        // file share must have lower case name
+        const shareName = getRandomHexString().toLowerCase();
+        await testUserInput.runWithInputs([resourceName, shareName, '5120'], async () => {
+            await vscode.commands.executeCommand('azureStorage.createFileShare');
+        });
+        const connectionString: string = await getConnectionString(resourceName);
+        const fileService: FileService = createFileService(connectionString);
+        const createdShare = await new Promise((resolve, reject) => {
```
nturinski: Could you add a type definition here?  It should be boolean | undefined, I'm assuming?

test/storageAccountActions.test.ts
```diff
@@ -64,11 +64,7 @@ suite('Storage Account Actions', async function (this: ISuiteCallbackContext): P
 
     test("copyConnectionString", async () => {
```
nturinski: I think you can get rid of this test since `createFileShare` is testing the validity of the connection string.

test/storageAccountActions.test.ts
nturinski: Also, I think this promise could be moved into its own function called `doesShareExist`.  I think having it in the test makes it a bit harder to read.

PR #471, opened 2019-10-02
Remove calls to vscode.workspace.rootPath

### LINKED ISSUE #453: Avoid using workspace.rootPath
This extension appears to be using the `workspace.rootPath` API. This is a deprecated property and may not work as expected. Please use `workspaceFolders` instead. More info:  https://github.com/microsoft/vscode/wiki/Adopting-Multi-Root-Workspace-APIs#eliminating-rootpath

https://github.com/microsoft/vscode/issues/69335

wwlorey:
Fixes #453

### DIFF
--- a/src/utils/workspaceUtils.ts
+++ b/src/utils/workspaceUtils.ts
@@ -1,0 +1,6 @@
+import * as vscode from 'vscode';
+
+export function getSingleRootWorkspace(): vscode.WorkspaceFolder | undefined {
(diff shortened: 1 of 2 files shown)


src/utils/workspaceUtils.ts
```diff
@@ -0,0 +1,6 @@
+import * as vscode from 'vscode';
```
nturinski: Needs the Microsoft license at the top

PR #473, opened 2019-10-03
Fix creating empty text blobs

### LINKED ISSUE #454: Pop up an error when creating an empty text blob
**Build Version:** [20190828.1](https://dev.azure.com/ms-azuretools/AzCode/_build/results?buildId=9223)
**OS:** Win10 & Mac

**Repro Steps:**
1. Create a storage account -> Create a blob container.
2. Right click the new blob container -> Select "Create Empty Text Blob...".
3. Check the result.

**Expect:**
Prompt to enter a name for the new block blob -> Succeed to create.

**Actual:**
Pop up an error.
image

**More Info:**

1. Prompt to enter a name for the new file and succeed to create a new empty text file for file share.
2. This issue doesn't reproduce on the previous release 0.7.1.

wwlorey:
Fixes #454 

`context.childType` wasn't guaranteed to be defined for the 'create empty text blob' scenario


wwlorey:
Fixes #454

PR #475, opened 2019-10-07
Ignore only .git and .vscode folders during deployment

### LINKED ISSUE #457: Deploy to static website ignores .well-known folder
image

When I click on Deploy Static Website on my webapp folder, it ignores the .well-known folder.

Has anyone had this issue or do I need to change any config ?

wwlorey:
Fixes #457

### DIFF
--- a/src/azureStorageExplorer/blobContainers/blobContainerNode.ts
+++ b/src/azureStorageExplorer/blobContainers/blobContainerNode.ts
@@ -348,7 +348,7 @@ export class BlobContainerTreeItem extends AzureParentTreeItem<IStorageRoot> imp
                     glob(
                         path.join(sourceFolderPath, '{**,.!(git|vscode)/**}'),
                         {
-                            mark: true, // Add '/' to folders
+                            mark: true // Add '/' to folders
                         },
                         (err, matches) => {
                             if (err) {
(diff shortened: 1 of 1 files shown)


src/azureStorageExplorer/blobContainers/blobContainerNode.ts
```diff
@@ -346,7 +346,7 @@ export class BlobContainerTreeItem extends AzureParentTreeItem<IStorageRoot> imp
             let filePathsWithAzureSeparator = await new Promise<string[]>(
                 (resolve, reject) => {
                     glob(
-                        path.join(sourceFolderPath, '**'),
+                        path.join(sourceFolderPath, '{**,.!(git|vscode)/**}'),
                         {
                             mark: true // Add '/' to folders
```
EricJizbaMSFT: I believe your fix will only include folders like `.example/`, but not `subfolder/.example/`. I suggest setting `dot` equal to `true` in the options here, instead of trying to do it all in the glob pattern above

Also you can leverage the `ignore` option here to exclude .git and .vscode

StephenWeatherford: Also, will "." match any characters, meaning agit and bvscode will match?

src/azureStorageExplorer/blobContainers/blobContainerNode.ts
```diff
@@ -303,6 +303,7 @@ export class BlobContainerTreeItem extends AzureParentTreeItem<IStorageRoot> imp
      *
      * @returns The primary web endpoint
      */
+    // tslint:disable-next-line: max-func-body-length
```
StephenWeatherford: Refactor?

EricJizbaMSFT: This is being included for a patch release, so @wwlorey you should keep the PR as simple as possible and we can refactor separately if desired

PR #476, opened 2019-10-07
Add 'Create Storage Account' button to sash

### LINKED ISSUE #306: [Suggestion]Add a button to create storage accounts
**OS**: All
**Build Version:** [20190214.4](https://dev.azure.com/ms-azuretools/AzCode/_build/results?buildId=3207)
**Regression:**  Not a Regression

**Repro Steps:**
Check the Storage Extension view.

**Actual:**
There is no button to create storage accounts.

**Suggestion:**
Add a button to create storage accounts.
image

### LINKED ISSUE #472: Add a plus button to create storage account in the sash
Screen Shot 2019-10-02 at 11 37 42 AM

wwlorey:
Fixes #306

Adds a plus button on the sash

Screen Shot 2019-10-07 at 1 34 54 PM

### DIFF
--- a/package.json
+++ b/package.json
@@ -257,8 +257,12 @@
 			},
 			{
 				"command": "azureStorage.createGpv2Account",
-				"title": "Create Storage Account",
-				"category": "Azure Storage"
+				"title": "Create Storage Account...",
+				"category": "Azure Storage",
+				"icon": {
+					"light": "resources/light/createStorageAccount.svg",
+					"dark": "resources/dark/createStorageAccount.svg"
+				}
 			},
 			{
 				"command": "azureStorage.deleteFile",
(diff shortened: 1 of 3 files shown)


EricJizbaMSFT:
So I believe we didn't do this in the past because we don't actually support very many options when it comes to creating a storage account. Ours only supports GPv2, but I think the portal has a bunch of options. Need to confirm this feature with @fiveisprime

fiveisprime:
This was something that Jing intervened on. I think it's fine to have a create button, but it needs to be a "create for web deployment" button and not a generic "create storage account" button.

wwlorey:
@fiveisprime We have the generic "create storage account" verbiage when you right click on a subscription. Should we update that as well?

Screen Shot 2019-10-09 at 11 35 14 AM

fiveisprime:
I didn't realize that one made it in. May be a case of "ship it and see if people ask for more." So.. I say go ahead with adding the button in the sash and see if people find it too limiting.

PR #477, opened 2019-10-08
add create table test case

v-etzha:
Add create table test case


test/storageAccountActions.test.ts
```diff
@@ -125,6 +137,19 @@ suite('Storage Account Actions', async function (this: ISuiteCallbackContext): P
         });
     }
 
+    // validate the table exists or not by its name and table service
+    async function doesTableExist(tableName: string, tableService: TableService): Promise<boolean | undefined> {
```
nturinski: This code is almost identical to `doesShareExist`.  I think you could refactor `doesShareExist` to be more generic so that it accepts a data type and a function as parameters to check if it exists.

Something like:

```
async function doesResourceExist<T>(resourceName: string, service: T, doesExist: async () =>: Promise<boolean>): Promise<boolean | undefined> {
return new Promise((resolve,reject) => {
  service.doesExist(...
...
...
  }
}

```

I didn't test this out, but I think that something like this should work.

nturinski:
Sorry for the long delay in comments.  I was thinking for `doesResourceExist` you could do something more like this:

```
async function doesResourceExist<T>(fn: (s: string, cb: ErrorOrResult<T>) => void, name: string): Promise<T> {
        return new Promise((resolve, reject) => fn(name, (err: Error | undefined, res: T) => {
            if (err) {
                reject(err);
            } else {
                resolve(res);
            }
        }));
    }
```

test/storageAccountActions.test.ts
```diff
@@ -93,8 +93,21 @@ suite('Storage Account Actions', async function (this: ISuiteCallbackContext): P
         });
         const connectionString: string = await getConnectionString(resourceName);
         const fileService: FileService = createFileService(connectionString);
-        const createdShare: boolean | undefined = await doesShareExist(shareName, fileService);
-        assert.ok(createdShare);
+        const createdShare: FileService.FileResult = await doesResourceExist<FileService.FileResult>(fileService, fileService.doesShareExist, shareName);
```
nturinski: Doesn't this have to be `<ShareResult>` as `doesShareExist` returns a `ShareResult` object, I believe?

PR #486, opened 2019-10-14
Ignore trailing slash in parentDirPath when parsing URIs

### LINKED ISSUE #455: Pop up an error when executing "New File" in File Explorer for FileShare
**Build Version:** [20190828.1](https://dev.azure.com/ms-azuretools/AzCode/_build/results?buildId=9223)
**OS:** Win10 & Mac

**Repro Steps:**
1. Create a storage account -> Create a file share.
2. Expand the new file share -> Click "Open in File Explorer...".
3. Click "New Folder" button to create a new folder ->Right click this new folder -> Select "New File".

**Expect:**
Succeed to create the new file without any error.

**Actual:**
Pop up an error.
image

**More Info:**

1. No error pops up if creating a new file directly in the root folder.
2. No error pops up when creating a new file in the new created folder for blob container.

wwlorey:
Fixes #455

The trailing `/` led to the parent tree item not being found in file creation due to different expectations for trailing slashes.

### DIFF
--- a/src/azureStorageExplorer/parseUri.ts
+++ b/src/azureStorageExplorer/parseUri.ts
@@ -35,7 +35,7 @@ export interface IParsedUri {
 
     /**
      * Path of parent directory within container or file share
-     * e.g. parentdir1/
+     * e.g. parentdir1
      */
     parentDirPath: string;
 
--- a/src/azureStorageExplorer/parseUri.ts
+++ b/src/azureStorageExplorer/parseUri.ts
@@ -48,7 +48,7 @@ export interface IParsedUri {
 
 export function parseUri(uri: vscode.Uri | string, fileType: string): IParsedUri {
     let path: string = uri instanceof vscode.Uri ? uri.path : uri;
-    const matches: RegExpMatchArray | null = path.match(`^(\/subscriptions\/[^\/]+\/resourceGroups\/[^\/]+\/providers\/Microsoft\.Storage\/storageAccounts\/[^\/]+\/${fileType}\/([^\/]+))\/?((.*?\/?)([^\/]*))$`);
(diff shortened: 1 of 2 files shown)


src/azureStorageExplorer/parseUri.ts
```diff
@@ -48,7 +48,7 @@ export interface IParsedUri {
 
 export function parseUri(uri: vscode.Uri | string, fileType: string): IParsedUri {
     let path: string = uri instanceof vscode.Uri ? uri.path : uri;
-    const matches: RegExpMatchArray | null = path.match(`^(\/subscriptions\/[^\/]+\/resourceGroups\/[^\/]+\/providers\/Microsoft\.Storage\/storageAccounts\/[^\/]+\/${fileType}\/([^\/]+))\/?((.*?\/?)([^\/]*))$`);
+    const matches: RegExpMatchArray | null = path.match(`^(\/subscriptions\/[^\/]+\/resourceGroups\/[^\/]+\/providers\/Microsoft\.Storage\/storageAccounts\/[^\/]+\/${fileType}\/([^\/]+))\/?((.*?)\/?([^\/]*))$`);
```
EricJizbaMSFT: Your change is fine, but it made me realize `.*?` should be `.*`. It doesn't make sense to have both quantifiers

src/azureStorageExplorer/parseUri.ts
```diff
@@ -48,7 +48,7 @@ export interface IParsedUri {
 
 export function parseUri(uri: vscode.Uri | string, fileType: string): IParsedUri {
```
EricJizbaMSFT: Please update the example above in the docs for `parentDirPath`

src/azureStorageExplorer/parseUri.ts
wwlorey: It appears there's a slight difference between the two:

Screen Shot 2019-10-14 at 4 34 57 PM

Screen Shot 2019-10-14 at 4 35 08 PM

The matching doesn't work as expected without `?` so I'll keep it in

Issue #431, opened 2019-08-07
[Discussion] Finalize uri format for storage file provider


EricJizbaMSFT:
File share uris currently look like this:
> azurestoragefile:/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/rgname/providers/Microsoft.Storage/storageAccounts/accountname/File Shares/filesharename

Blob uris currently look like this:
> azurestorageblob:/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/rgname/providers/Microsoft.Storage/storageAccounts/accountname/Blob Containers/blobname

We should make extra sure we agree on the format of the uri because changing this will break users in the future. A few things to consider:
1. Do we need separate 'azurestoragefile' and 'azurestorageblob' schemes? The id's above already say if it's a blob or file share, so I don't see any reason to have separate ones
1. Do these uris still work in sovereign accounts?
1. Should we make our uri more similar to storage explorer? This is what a storage explorer uris looks like:
> storageexplorer://v=1&accountid=%2Fsubscriptions%2F00000000-0000-0000-0000-000000000000%2FresourceGroups%2Frgname%2Fproviders%2FMicrosoft.Storage%2FstorageAccounts%2Faccountname&subscriptionid=00000000-0000-0000-0000-000000000000&source=VSCODE-AzureStorage&resourcetype=Azure.BlobContainer&resourcename=filesharename

fiveisprime:
1. Nope, that seems redundant.
1. That's definitely something that should be tested..
1. Is there any compelling reason not to follow that pattern? It's a bit uglier, but what does that matter?

EricJizbaMSFT:
I agree it's uglier and I'm not sure we need all those extra parameters haha. I would probably advocate for something like this:
> azurestorage:///folder1/subfolder2/hello.txt?resourceId=%2Fsubscriptions%2F00000000-0000-0000-0000-000000000000%2FresourceGroups%2Frgname%2Fproviders%2FMicrosoft.Storage%2FstorageAccounts%2Faccountname%2FFile%20Shares%2Ffilesharename

I think I'm sold on leveraging the query section for two reasons:
1. VS Code displays the uri in the open editors section, so I'm hoping just "folder1/subfolder2/hello.txt" would get displayed instead of the whole id like it is in master today:
[...]

PR #488, opened 2019-10-15
Add "Open in File Explorer" commands to context menu & command palette

### LINKED ISSUE #412: [Suggestion]Support "Open in File Explorer..." from the context menu and Command Palette
**OS:** Win10
**Build Version:** [20190730.9](https://dev.azure.com/ms-azuretools/AzCode/_build/results?buildId=8267)
**Regression**: Not a Regression

**Repro Steps:**
2. Drill down from one storage account --> "Blob Containers" node -->expand one blob container.
3. Check the blob container context menu and Command Palette.

**Actual**:
There is no "Open in File Explorer...".

**Suggestion:**
Support "Open in File Explorer..." from the context menu and Command Palette.
image

wwlorey:
Fixes #412 

This adds a menu entry point to blob container and file share tree items:
Screen Shot 2019-10-15 at 11 28 48 AM

As well as a catch-all command to the command palette:
Screen Shot 2019-10-15 at 11 31 54 AM
Screen Shot 2019-10-15 at 11 31 40 AM


wwlorey:
**Offline discussion**: Combine all "Open in File Explorer" commands into one. Also wait to fix this until progress is made on #431

src/extension.ts
```diff
@@ -71,14 +71,21 @@ export async function activateInternal(context: vscode.ExtensionContext, perfSta
...
+                    let quickPicks: IAzureQuickPickItem<string>[] = [
+                        {
+                            label: "Blob Container",
+                            data: BlobContainerTreeItem.contextValue
+                        },
+                        {
+                            label: "File Share",
+                            data: FileShareTreeItem.contextValue
+                        }];
+                    let contextValue: string = (await ext.ui.showQuickPick(quickPicks, { placeHolder: "Select the resource type to open in File Explorer" })).data;
```
EricJizbaMSFT: You could actually do all of this through `showTreeItemPicker` instead of creating your own prompt. It would involve passing both `BlobContainerTreeItem.contextValue` and `FileShareTreeItem.contextValue` to `showTreeItemPicker` and modifying `pickTreeItemImpl` and/or `isAncestorOfImpl` in a few places. You're welcome to try if you want, but for this PR I'm fine merging as-is since I think  this case is unique. If we run into similar cases in the future, it will be more worth it.

PR #452, opened 2019-08-22
Monal's PR: Fix URI format for virtual file

PrashanthCorp:
I'm creating the PR on @monalgarg's repo. 
The aim is to implement the URI format mentioned in Eric's comment: https://github.com/microsoft/vscode-azurestorage/issues/431#issuecomment-519591306

PR #489, opened 2019-10-16
Update uri format

### LINKED ISSUE #431: [Discussion] Finalize uri format for storage file provider

wwlorey:
Fixes #431 
Based on #452

This combines the `FileShareFS` and `BlobContainerFS` files into `AzureStorageFS` with a unified uri format. The new format deviates from what was suggested in #431 slightly: the fileshare/blobcontainer name is prepended to the `path` section of the uri because I ran into issues when no files/blobs existed.


src/azureStorageExplorer/AzureStorageFS.ts
```diff
@@ -0,0 +1,467 @@
...
+        }) || { type: vscode.FileType.Unknown, ctime: 0, mtime: 0, size: 0 };
+    }
+
+    async readDirectory(uri: vscode.Uri): Promise<[string, vscode.FileType][]> {
+        return this.isFileShareUri(uri) ? await this.readDirectoryFileShare(uri) : await this.readDirectoryBlobContainer(uri);
+    }
+
+    async readDirectoryFileShare(uri: vscode.Uri): Promise<[string, vscode.FileType][]> {
+        return await callWithTelemetryAndErrorHandling('readDirectoryFileShare', async (context) => {
+            context.telemetry.suppressIfSuccessful = true;
```
EricJizbaMSFT: Methods like `readDirectoryFileShare` and `readDirectoryBlobContainer` are almost exactly the same. I'd prefer to see all of those methods combined so that we have easier maintenance and consistent behavior

EricJizbaMSFT: So this was meant to be a generic comment for all cases where you call `isFileShareUri` - not just for readDirectory. I'm kind of waiting to fully review this PR until this is done because it'll affect basically the whole file

src/azureStorageExplorer/parseUri.ts
```diff
@@ -23,42 +31,58 @@ export interface IParsedUri {
...
+            rootName: queryMatches[3],
+            filePath: pathMatches[1],
+            dirPath: pathMatches[1] ? `${pathMatches[1]}/` : '',
+            parentDirPath: pathMatches[2],
+            baseName: pathMatches[3]
         };
     }
 }
+
+export function idToUri(id: string): vscode.Uri {
```
EricJizbaMSFT: I'm not a huge fan of this function. It seems to parse out the whole id only to reconstruct it basically the same at the end.

Also when you use it in AzureStorageFS, you're essentially converting an id to a uri and back to an id in `lookup`. Instead `lookup` could allow you to directly pass in `rootPath` and `filePath` (I believe the only things used in `lookup`?) in place of a `uri`

src/azureStorageExplorer/AzureStorageFS.ts
```diff
@@ -0,0 +1,355 @@
...
+                                        reject(error);
+                                    } else {
+                                        resolve(result);
+                                    }
+                                });
+                            });
+                        } else {
+                            await new Promise<void>((resolve, reject) => {
+                                let contentType: string | null = mime.getType(parsedUri.filePath);
+                                let temp: string | undefined = contentType === null ? undefined : contentType;
```
EricJizbaMSFT: Can re-use this: https://github.com/microsoft/vscode-azurestorage/blob/3ecff167f07475be1d57497568aa3be8491aacd7/src/azureStorageExplorer/blobContainers/blobContainerNode.ts#L587

src/azureStorageExplorer/parseUri.ts
wwlorey: It looks like we need a `uri` for `lookup` to throw file system errors. Would keeping the `uri` parameter and adding `rootPath` & `filePath` as parameters be alright? Just trying to limit calls to `parseUri`

src/azureStorageExplorer/parseUri.ts
```diff
@@ -23,42 +31,58 @@ export interface IParsedUri {
...
+     * e.g. parentdir/subdir/blob
      */
     filePath: string;
 
     /**
      * Same as `filePath`, just with a delimiter at the end in case this is a directory
-     * e.g. parentdir1/subdir/
+     * e.g. parentdir/subdir/blob/
      */
     dirPath: string;
```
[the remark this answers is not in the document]
wwlorey: I ended up using `dirPath` after addressing comments

src/azureStorageExplorer/AzureStorageFS.ts
```diff
@@ -0,0 +1,336 @@
...
+                throw getFileSystemError(uri, context, vscode.FileSystemError.FileExists);
+            } else {
+                await vscode.window.withProgress({ location: vscode.ProgressLocation.Notification }, async (progress) => {
+                    if (resultChild.exists) {
+                        progress.report({ message: `Saving ${writeToFileShare ? 'file' : 'blob'} ${parsedUri.filePath}` });
+
+                        if (treeItem instanceof FileShareTreeItem) {
+                            await createFile(parsedUri.dirPath, parsedUri.filePath, treeItem.share, treeItem.root);
+                        } else {
+                            await treeItem.createTextBlockBlob(parsedUri.rootName, content.toString());
```
EricJizbaMSFT: rootName seems wrong. Isn't that the container name, not the blob name?

src/azureStorageExplorer/blobContainers/blobContainerNode.ts
```diff
@@ -584,15 +584,15 @@ export class BlobContainerTreeItem extends AzureParentTreeItem<IStorageRoot> imp
     }
 
     // tslint:disable-next-line:promise-function-async // Grandfathered in
-    private createTextBlockBlob(name: string): Promise<azureStorage.BlobService.BlobResult> {
+    public createTextBlockBlob(name: string, text?: string | Buffer): Promise<azureStorage.BlobService.BlobResult> {
         return new Promise((resolve, reject) => {
             let blobService = this.root.createBlobService();
             const options = <azureStorage.BlobService.CreateBlobRequestOptions>{
                 contentSettings: {
                     contentType: 'text/plain'
```
EricJizbaMSFT: this has different behavior for contentType than the previous code. Also doesn't match this: https://github.com/microsoft/vscode-azurestorage/blob/d3d71093ec846c472d7230ffd39bf56d8d14b8f6/src/azureStorageExplorer/blobContainers/BlobDirectoryTreeItem.ts#L90

I feel like the other examples are more accurate than always using 'text/plain'

src/azureStorageExplorer/parseUri.ts
```diff
@@ -23,42 +24,47 @@ export interface IParsedUri {
...
+            filePath: pathMatches[1],
+            dirPath: pathMatches[1] ? `${pathMatches[1]}/` : '',
+            parentDirPath: pathMatches[2],
+            baseName: pathMatches[3]
         };
     }
 }
+
+export function idToUri(rootPath: string, filePath?: string): vscode.Uri {
+    const rootName = rootPath.split(path.sep).pop();
```
EricJizbaMSFT: How about `path.basename(rootPath)`. Does that work?

src/azureStorageExplorer/AzureStorageFS.ts
```diff
@@ -0,0 +1,336 @@
...
+        }
+    }
+
+    private async lookupBlobContainer(uri: vscode.Uri, context: IActionContext, rootPath: string, filePath: string, endSearchEarly?: boolean): Promise<AzureStorageBlobTreeItem> {
+        let treeItem: AzureStorageBlobTreeItem = <BlobContainerTreeItem>await this.lookupRoot(uri, context, rootPath);
+        if (filePath === '') {
+            return treeItem;
+        }
+
+        let pathToLook = filePath.split(path.sep);
```
EricJizbaMSFT: `path.sep` will be OS-specific, which probably won't work on Windows since the uri just uses posix separators

src/azureStorageExplorer/fileShares/fileUtils.ts
```diff
@@ -59,7 +59,7 @@ async function doesFileExist(fileName: string, parent: AzureParentTreeItem<IStor
 export function getFile(directoryPath: string, name: string, share: FileService.ShareResult, root: IStorageRoot): Promise<azureStorage.FileService.FileResult> {
```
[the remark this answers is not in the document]
wwlorey: My bad - we can easily use `doesFileExist`/`doesBlobExist`

src/azureStorageExplorer/parseUri.ts
```diff
@@ -23,42 +24,47 @@ export interface IParsedUri {
...
 }
 
-export function parseUri(uri: vscode.Uri | string, fileType: string): IParsedUri {
-    let path: string = uri instanceof vscode.Uri ? uri.path : uri;
-    const matches: RegExpMatchArray | null = path.match(`^(\/subscriptions\/[^\/]+\/resourceGroups\/[^\/]+\/providers\/Microsoft\.Storage\/storageAccounts\/[^\/]+\/${fileType}\/([^\/]+))\/?((.*?)\/?([^\/]*))$`);
-    if (!matches) {
-        throw new Error(`Invalid ${fileType} uri. Cannot view or modify ${uri}.`);
+export function parseUri(uri: vscode.Uri): IParsedUri {
+    const queryMatches: RegExpMatchArray | null = uri.query.match('^resourceId=(\/subscriptions\/[^\/]+\/resourceGroups\/[^\/]+\/providers\/Microsoft\.Storage\/storageAccounts\/[^\/]+\/[^\/]+\/([^\/]+))$');
+    const pathMatches: RegExpMatchArray | null = uri.path.match('^\/[^\/]+[^\/]+\/?((.*?)\/?([^\/]*))$');
```
EricJizbaMSFT: I'd like to see more use of built-in methods for path stuff too. For example:
```
const filePath: string = uri.path.substring(uri.path.indexOf('/', 1) + 1); // remove root name (option 1)
const filePath2: string = uri.path.replace(/^\/.*?\//, ''); // remove root name (option 2)
const parentDirPath = path.dirname(filePath);
const baseName = path.basename(filePath);
```

Issue #438, opened 2019-08-08
Replace BaseEditor with Virtual File System Provider


EricJizbaMSFT:
Aka remove this: https://github.com/microsoft/vscode-azurestorage/blob/master/src/azureServiceExplorer/editors/RemoteFileEditor.ts

Which saves files to a temp folder, and use the new file view implementation

PR #396, opened 2019-07-22
Better error handling and telemetry data for viewing azure storage in file explorer

monalgarg:
Fixes #394

PR #494, opened 2019-10-28
Fix cmd + clicking on a file path to open file

### LINKED ISSUE #446: Cmd + clicking one a file path does not correctly open the file

wwlorey:
Fixes #446


src/azureStorageExplorer/AzureStorageFS.ts
```diff
@@ -246,13 +251,24 @@ export class AzureStorageFS implements vscode.FileSystemProvider {
 
     async rename(oldUri: vscode.Uri, newUri: vscode.Uri, _options: { overwrite: boolean; }): Promise<void> {
         return await callWithTelemetryAndErrorHandling('rename', async (context) => {
-            showRenameError(oldUri, newUri, context);
+            let oldUriParsed = this.parseUri(oldUri);
+            let newUriParsed = this.parseUri(newUri);
```
EricJizbaMSFT: This seems unrelated to cmd clicking a file path. I prefer smaller PRs that have a single purpose. See comments like [this](https://github.com/microsoft/vscode-azurestorage/pull/396#pullrequestreview-265033653) for more info

src/azureStorageExplorer/AzureStorageFS.ts
```diff
@@ -329,7 +345,36 @@ export class AzureStorageFS implements vscode.FileSystemProvider {
         }
     }
 
+    private verifyUri(uri: vscode.Uri): vscode.Uri {
+        // Fallback to the cached query if this uri doesn't have a query defined
+        this._queryCache = uri.query || this._queryCache;
```
EricJizbaMSFT: I like using a cache for this, but I wonder if we should be a little more careful. For example, here are a few ideas:
1. Instead of a single cached query - use a Map of rootName to query (since rootName is in both the path and the resourceId). This should handle multi-root workspaces as long as the rootName is different for each workspace folder
1. Every time you set the cache for a folder, check if it equals the existing cache value. If it doesn't, somehow mark the cache as invalid and "give up" (aka throw an error that we can track in telemetry) for the rest of that VS Code session

src/utils/uriUtils.ts
```diff
@@ -0,0 +1,12 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.md in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as path from "path";
+import * as vscode from 'vscode';
+
+export function idToUri(resourceId: string, filePath?: string): vscode.Uri {
```
EricJizbaMSFT: I think I'd rather make this a static method on `AzureStorageFS`. utils can become a confusing dumping ground for any code, so we try to avoid it if there's an alternative

src/azureStorageExplorer/IParsedUri.ts
```diff
@@ -3,10 +3,6 @@
  *  Licensed under the MIT License. See License.md in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import * as path from "path";
-import * as querystring from 'querystring';
-import * as vscode from 'vscode';
-
 /**
```
EricJizbaMSFT: Is `IParsedUri` used outside `AzureStorageFS`? If not, I say move it to that file and make it private (aka don't export it)

src/azureStorageExplorer/AzureStorageFS.ts
EricJizbaMSFT: Never mind is this just copied from `showRenameFile.ts`? If so - I'm fine with including that. Curious about the `refreshFilesExplorer` thing above though. Was that copied from elsewhere or is that a separate bug fix?

wwlorey: Yeah it was copied from `showRenameFile.ts`. I added the file explorer refresh to fix a bug that cmd+clicking into files introduced. Should that be addressed in a separate PR?

src/azureStorageExplorer/AzureStorageFS.ts
```diff
@@ -17,18 +18,51 @@ import { FileShareTreeItem, IFileShareCreateChildContext } from "./fileShares/fi
...
 
 type AzureStorageFileTreeItem = FileTreeItem | DirectoryTreeItem | FileShareTreeItem;
 type AzureStorageBlobTreeItem = BlobTreeItem | BlobDirectoryTreeItem | BlobContainerTreeItem;
 type AzureStorageTreeItem = AzureStorageFileTreeItem | AzureStorageBlobTreeItem;
 type AzureStorageDirectoryTreeItem = DirectoryTreeItem | FileShareTreeItem | BlobDirectoryTreeItem | BlobContainerTreeItem;
 
+/**
+ * Example uri: azurestorage:///container1/parentdir/subdir/blob?resourceId=/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/resourcegroup1/providers/Microsoft.Storage/storageAccounts/storageaccount1/Blob Containers/container1
+ */
+interface IParsedUri {
```
EricJizbaMSFT: nit: Usually the exported class goes at the top of the file and "private" interfaces/functions/classes go below

src/azureStorageExplorer/AzureStorageFS.ts
```diff
@@ -17,18 +18,51 @@ import { FileShareTreeItem, IFileShareCreateChildContext } from "./fileShares/fi
...
+    /**
+     * Name of file or directory
+     * e.g. blob
+     */
+    baseName: string;
+}
+
 export class AzureStorageFS implements vscode.FileSystemProvider {
     private _emitter: vscode.EventEmitter<vscode.FileChangeEvent[]> = new vscode.EventEmitter<vscode.FileChangeEvent[]>();
+    private _queryCache: { [rootName: string]: { query: string, invalid?: boolean } } = {};
```
EricJizbaMSFT: This type is fine, but are you aware of [`Map`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map)? Sometimes it looks a little cleaner.

If you stick with this type, I believe you need to change it to this:
```
{ [rootName: string]: { query: string, invalid?: boolean } | undefined }
```

Just so that strictNullChecks can enforce the fact that the entry might not exist in the cache

src/azureStorageExplorer/AzureStorageFS.ts
```diff
@@ -329,7 +377,66 @@ export class AzureStorageFS implements vscode.FileSystemProvider {
...
+
+        if (rootName in this._queryCache) {
+            if (this._queryCache[rootName].invalid) {
+                throw new Error('Invalid URI query cache.');
+            }
+
+            if (uri.query) {
+                if (this._queryCache[rootName].query !== uri.query) {
+                    this._queryCache[rootName].invalid = true;
+                    throw new Error('Invalid URI query cache.');
```
EricJizbaMSFT: I don't think this should throw an error. The current operation will work fine since the uri already has a query. The cache only matters for operations where we _don't_ have a query

src/azureStorageExplorer/AzureStorageFS.ts
```diff
@@ -329,7 +377,66 @@ export class AzureStorageFS implements vscode.FileSystemProvider {
...
+                }
+            } else {
+                // Fallback to the cached query because this uri's query doesn't exist
+                return vscode.Uri.parse(`azurestorage://${uri.path}?${this._queryCache[rootName].query}`);
```
EricJizbaMSFT: Should we be caching the whole query or just the resourceId? Honestly just want to know what you think - haven't thought about it much myself

src/azureStorageExplorer/AzureStorageFS.ts
```diff
@@ -329,7 +377,66 @@ export class AzureStorageFS implements vscode.FileSystemProvider {
...
+        return match ? match[0] : '';
+    }
+
+    private verifyUri(uri: vscode.Uri, rootName?: string): vscode.Uri {
+        // Remove slashes from rootName for consistency
+        rootName = (rootName || this.getRootName(uri)).replace(/\//g, '');
+
+        if (rootName in this._queryCache) {
+            if (this._queryCache[rootName].invalid) {
+                throw new Error('Invalid URI query cache.');
```
EricJizbaMSFT: Can we make this error more human-understandable? For all the errors in this function, this one seems the most likely to happen. Something about how we can't auto detect the resource to go with the path and if they reload with a single resource it might work better

src/azureStorageExplorer/AzureStorageFS.ts
wwlorey: The query structure is `resourceId=<resourceId>` so we're pretty much only caching resourceId. Keeping the `resourceId=` part makes it a bit nicer to reconstruct a URI from the cached query

src/azureStorageExplorer/AzureStorageFS.ts
```diff
@@ -329,7 +348,93 @@ export class AzureStorageFS implements vscode.FileSystemProvider {
...
+    }
+
+    private verifyUri(uri: vscode.Uri, rootName?: string): vscode.Uri {
+        // Remove slashes from rootName for consistency
+        rootName = (rootName || this.getRootName(uri)).replace(/\//g, '');
+
+        let cache = this._queryCache.get(rootName);
+        if (cache) {
+            if (uri.query) {
+                this._queryCache.set(rootName, { query: cache.query, invalid: cache.query !== uri.query });
```
EricJizbaMSFT: Once the cache becomes invalid, we want it to stay invalid (Since it's not clear which query we should use). I believe this is what we want:
```
invalid: cache.invalid || cache.query !== uri.query
```

src/azureStorageExplorer/AzureStorageFS.ts
```diff
@@ -386,24 +357,22 @@ export class AzureStorageFS implements vscode.FileSystemProvider {
...
+        if (cache) {
             if (uri.query) {
-                if (this._queryCache[rootName].query !== uri.query) {
-                    this._queryCache[rootName].invalid = true;
-                    throw new Error('Invalid URI query cache.');
-                }
+                this._queryCache.set(rootName, { query: cache.query, invalid: cache.query !== uri.query });
             } else {
+                if (cache.invalid) {
+                    throw new Error('Cannot auto-detect resource path. Try opening it in a single-root workspace.');
```
EricJizbaMSFT: This isn't really a "single-root" vs. "multi-root" problem. It will be more of a problem when we fix https://github.com/microsoft/vscode-azurestorage/issues/438. We want to get rid of the existing editing experience when clicking a file from the Azure view and instead open a url to a single file at a time using AzureStorageFS, which means people could open urls to multiple resources in one workspace.

How about something like this:
> Cannot auto-detect resource path. For this functionality, re-load VS Code and only open one resource per workspace.

src/azureStorageExplorer/AzureStorageFS.ts
EricJizbaMSFT: Well - that's the _ideal_ query structure. Users can type any url they want to open, plus we might add more parameters to the query in the future. Anyways I'm fine caching the whole query

PR #499, opened 2019-11-01
Make file/blob metadata & properties persistent after update

### LINKED ISSUE #327: Saving file share removes metadata
Saving a file with metadata removes all metadata.

This issue was reported in #20 then "fixed" #24 but later noted it was only partially fixed.

Issue still persists to this day and makes systems depending on metadata unusable with this extension.

Thanks

wwlorey:
Fixes #327

### DIFF
--- a/src/azureServiceExplorer/editors/RemoteFileEditor.ts
+++ b/src/azureServiceExplorer/editors/RemoteFileEditor.ts
@@ -8,6 +8,10 @@ import * as path from "path";
 import { TextDocument, window } from 'vscode';
 import * as vscode from "vscode";
 import { DialogResponses, IActionContext, UserCancelledError } from 'vscode-azureextensionui';
+import { BlobContainerTreeItem } from '../../azureStorageExplorer/blobContainers/blobContainerNode';
+import { BlobTreeItem } from '../../azureStorageExplorer/blobContainers/blobNode';
+import { FileTreeItem } from '../../azureStorageExplorer/fileShares/fileNode';
+import { getFile, getFileMetadata } from '../../azureStorageExplorer/fileShares/fileUtils';
 import { TemporaryFile } from '../../components/temporaryFile';
 import { ext } from '../../extensionVariables';
 import { IRemoteFileHandler } from './IRemoteFileHandler';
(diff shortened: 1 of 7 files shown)


src/azureStorageExplorer/blobContainers/blobContainerNode.ts
```diff
@@ -603,15 +603,15 @@ export class BlobContainerTreeItem extends AzureParentTreeItem<IStorageRoot> imp
...
-            const options = <azureStorage.BlobService.CreateBlobRequestOptions>{
-                contentSettings: {
-                    contentType: contentType ? contentType : undefined
-                }
-            };
+
+            // tslint:disable: strict-boolean-expressions
+            options = options || {};
+            options.contentSettings = options.contentSettings || {};
+            options.contentSettings.contentType = mime.getType(name) || undefined;
```
wwlorey: I don't feel great about always overwriting `contentType` but any type read from `updateBlockBlob` would be `text/plain` and it seemed cleaner to deal with that here.

wwlorey:
This only fixes files/blobs updated by AzureStorageFS. I need to update for the Azure view use case.

src/azureStorageExplorer/blobContainers/blobContainerNode.ts
wwlorey: `contentType` is the only thing that would get overwritten here. I could change `updateBlockBlob` to not call `createBlockBlob` when saving blob changes. This could leave `createBlockBlob` for just creating new blobs.

My comment about `text/plain` isn't entirely correct - the `contentType` is set to `text/plain` in many cases for blobs created using the extension because we call `createBlockBlobFromText` when creating.

src/azureServiceExplorer/editors/RemoteFileEditor.ts
```diff
@@ -28,6 +32,18 @@ export class RemoteFileEditor<ContextT> implements vscode.Disposable {
         if (filePath) {
             actionContext.telemetry.suppressIfSuccessful = false;
             const context: ContextT = this.fileMap[filePath][1];
+
+            if (context instanceof BlobTreeItem) {
```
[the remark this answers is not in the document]
wwlorey: Yeah the logic in `uploadFile` was from a previous PR that didn't fix the prob entirely. I'll move what I added in `RemoteFileEditor` there

src/azureStorageExplorer/blobContainers/blobFileHandler.ts
```diff
@@ -89,35 +89,11 @@ export class BlobFileHandler implements IRemoteFileHandler<BlobTreeItem> {
     async uploadFile(treeItem: BlobTreeItem, filePath: string): Promise<void> {
         await this.checkCanUpload(treeItem, filePath);
 
-        let blobService = treeItem.root.createBlobService();
-        let createOptions: azureStorage.BlobService.CreateBlockBlobRequestOptions = {};
-
-        if (treeItem.blob.contentSettings) {
-            createOptions.contentSettings = treeItem.blob.contentSettings;
-            createOptions.contentSettings.contentMD5 = undefined; // Needs to be filled in by SDK
+        let container: BlobContainerTreeItem | undefined = <BlobContainerTreeItem | undefined>treeItem.parent;
```
EricJizbaMSFT: Couldn't parent be `BlobDirectoryTreeItem`? If so - maybe you move those methods outside BlobContainerTreeItem and into a blobUtils file like already exists for fileUtils

src/azureStorageExplorer/blobContainers/blobContainerNode.ts
```diff
@@ -6,7 +6,6 @@
 import * as azureStorage from "azure-storage";
 import * as fse from 'fs-extra';
 import * as glob from 'glob';
-import * as mime from "mime";
```
EricJizbaMSFT: Can you remove 'mime' from package.json now?

wwlorey: mime is still used in BlobDirectoryTreeItem

PR #504, opened 2019-11-08
Add create queue test case

v-etzha:
add create queue test case


test/storageAccountActions.test.ts
```diff
@@ -97,6 +97,19 @@ suite('Storage Account Actions', async function (this: ISuiteCallbackContext): P
         assert.ok(createdShare.exists);
     });
 
+    test("createQueue", async () => {
+        await validateAccountExists(resourceName, resourceName);
+        // queue must have lower case name
+        const queueName = getRandomHexString().toLowerCase();
```
nturinski: nit: I didn't notice before, but could you type define all the names (containerName, shareName, queueName) as strings?

PR #380, opened 2019-06-27
Add retryPolicy on creating services (blob/queue/fileshare/table)

PrashanthCorp:
The timeout errors arise due to the lack of retries in blobService. 

Repro: 
```
 npx create react-app my-react-app
 cd my-react-app
 npm install
```
 Deploy the entire app, then deploy it again and choose yes to delete all files

Fixes #352 
Fixes  #345 
Fixes  #370 
Fixes #334 
Fixes #340 
Fixes #338 
Fixes #339 
Fixes #342
Fixes #317
Fixes #362

PR #513, opened 2019-11-18
Update ⬆️ to newest Azure Storage SDK for blobs 🗄

### LINKED ISSUE #328: Switch over to azure-sdk-for-js
https://github.com/Azure/azure-sdk-for-js/blob/master/documentation/Migration.md

wwlorey:
Fixes partially: #328


src/azureStorageExplorer/blobContainers/BlobDirectoryTreeItem.ts
```diff
@@ -1,218 +0,0 @@
-/*---------------------------------------------------------------------------------------------
```
EricJizbaMSFT: This file was one of the few files that _is_ named correctly (the name should match the class). Also ideally file name changes happen in their own PR so that it doesn't mess with the diff

src/azureStorageExplorer/blobContainers/blobFileHandler.ts
```diff
@@ -56,39 +54,15 @@ export class BlobFileHandler implements IRemoteFileHandler<BlobTreeItem> {
...
-            () => {
-                const completed = <string>speedSummary.getCompleteSize(true);
-                const total = <string>speedSummary.getTotalSize(true);
-                const percent = speedSummary.getCompletePercent(0);
-                const msg = `${blob.name}: ${completed}/${total} (${percent}%)`;
-                return msg;
-            });
-
+        const blockBlobClient = treeItem.root.createBlockBlobClient(treeItem.container.name, treeItem.fullPath);
+        await blockBlobClient.downloadToFile(filePath);
```
EricJizbaMSFT: Can we keep the logic that displays progress?

package.json
```diff
@@ -195,7 +195,7 @@
 			},
 			{
 				"command": "azureStorage.createBlockTextBlob",
-				"title": "Create Empty Text Blob...",
```
EricJizbaMSFT: > I snuck in the fix for text/plain content type defaults for blobs. Now it guesses the content type when creating/updating and respects the existing content type if it exists. I'll address the fix for file shares separately.

Can we unsneak the fix and do it separately? 🙃

src/azureStorageExplorer/blobContainers/blobContainerGroupNode.ts
```diff
@@ -45,19 +44,12 @@ export class BlobContainerGroupTreeItem extends AzureParentTreeItem<IStorageRoot
...
-                    resolve(result);
-                }
-            });
-        });
+    async listContainers(continuationToken?: string, maxPageSize: number = 50): Promise<azureStorageBlob.ListContainersSegmentResponse> {
+        const blobServiceClient = this.root.createBlobServiceClient();
+        let response = blobServiceClient.listContainers().byPage({ continuationToken, maxPageSize });
+
+        // tslint:disable-next-line: no-unsafe-any
+        return (await response.next()).value;
```
EricJizbaMSFT: What's the "unsafe-any" here? Is the package missing correct types?

src/azureStorageExplorer/blobContainers/blobUtils.ts
```diff
@@ -3,56 +3,130 @@
...
+}
+
+export async function getExistingProperties(parent: BlobTreeItem | BlobContainerTreeItem | BlobDirectoryTreeItem, blobPath: string): Promise<azureStorageBlob.BlockBlobUploadOptions | undefined> {
+    const blockBlobClient = parent.root.createBlockBlobClient(parent.container.name, blobPath);
+    if (await blockBlobClient.exists()) {
+        let existingProperties: azureStorageBlob.BlobGetPropertiesResponse = await blockBlobClient.getProperties();
+        return {
+            metadata: existingProperties.metadata,
+            blobHTTPHeaders: {
+                blobCacheControl: existingProperties.cacheControl,
```
EricJizbaMSFT: Not a huge fan of copying out each property here. Seems like we could get into problems if they ever add properties. Can we just return `existingProperties` as is? And only modify blobContentMD5?

src/azureStorageExplorer/storageAccounts/storageAccountNode.ts
```diff
@@ -415,4 +400,14 @@ export class StorageAccountTreeItem extends AzureParentTreeItem<IStorageRoot> {
             throw new Error("This storage account does not support static website hosting.");
         }
     }
+
+    private createBlobServiceClient(): azureStorageBlob.BlobServiceClient {
+        const credential = new azureStorageBlob.StorageSharedKeyCredential(this.storageAccount.name, this.key.value);
+        return new azureStorageBlob.BlobServiceClient(`https://${this.storageAccount.name}.blob.core.windows.net`, credential);
```
EricJizbaMSFT: I don't think this'll work with sovereign accounts because the url will be different. Check out `this.storageAccount.primaryEndpoints.blob` which should have the url you need without hard-coding anything

src/azureStorageExplorer/storageAccounts/storageAccountNode.ts
```diff
@@ -157,6 +158,20 @@ export class StorageAccountTreeItem extends AzureParentTreeItem<IStorageRoot> {
...
+            },
+            createBlobClient: (containerName: string, blobName: string) => {
+                const blobContainerClient = this.createBlobContainerClient(containerName);
+                return blobContainerClient.getBlobClient(blobName);
+            },
+            createBlockBlobClient: (containerName: string, blobName: string) => {
+                const blobContainerClient = this.createBlobContainerClient(containerName);
+                return blobContainerClient.getBlockBlobClient(blobName);
+            },
             createBlobService: () => {
```
EricJizbaMSFT: Can we get rid of the old `createBlobService` now?

src/azureStorageExplorer/storageAccounts/storageAccountNode.ts
```diff
@@ -157,6 +158,20 @@ export class StorageAccountTreeItem extends AzureParentTreeItem<IStorageRoot> {
...
                 return azureStorage.createBlobService(this.storageAccount.name, this.key.value, this.storageAccount.primaryEndpoints.blob).withFilter(new azureStorage.ExponentialRetryPolicyFilter());
             },
```
EricJizbaMSFT: Is there an analagous `ExponentialRetryPolicyFilter` for the new sdk? This was a pretty important fix: https://github.com/microsoft/vscode-azurestorage/pull/380

src/azureStorageExplorer/blobContainers/BlobDirectoryTreeItem.ts
wwlorey: Should I wait to rename those files in a separate PR?

EricJizbaMSFT: Yes please :) Also feel free to bother someone to review that PR as soon as you submit it since it should have no functional changes (aka easy to review) and so that we can avoid a ton of merge conflicts

src/azureStorageExplorer/blobContainers/blobContainerGroupNode.ts
wwlorey: As far as I can tell, types aren't provided here :(

For some context, here's an updated version with more typing:

```typescript
        let response: AsyncIterableIterator<azureStorageBlob.ServiceListContainersSegmentResponse> = blobServiceClient.listContainers().byPage({ continuationToken, maxPageSize });
        // tslint:disable-next-line:no-any
        let next: IteratorResult<azureStorageBlob.ServiceListContainersSegmentResponse, any> = await response.next();

        // tslint:disable-next-line: no-unsafe-any
        return next.value;
```

src/azureStorageExplorer/blobContainers/blobUtils.ts
wwlorey: Yeah this isn't ideal. The reason I enumerated everything is because the property names don't match up. Do you know if there's a cleaner way to take care of this?

src/azureStorageExplorer/storageAccounts/storageAccountNode.ts
wwlorey: That retry behavior appears to be built in. The default is an exponential retry policy. Those config settings are under [retryOptions](https://github.com/Azure/azure-sdk-for-js/blob/e3e31c8c/sdk/storage/storage-blob/src/Pipeline.ts#L152) of type [StorageRetryOptions](https://github.com/Azure/azure-sdk-for-js/blob/070905e649e2f7ace20884393663be5fcbd20cb7/sdk/storage/storage-blob/src/StorageRetryPolicyFactory.ts#L15)

src/azureStorageExplorer/IStorageRoot.ts
```diff
@@ -3,13 +3,17 @@
...
+import * as azureStorageBlob from "@azure/storage-blob";
 import * as azureStorage from "azure-storage";
 import { ISubscriptionContext } from "vscode-azureextensionui";
 import { StorageAccountWrapper } from "../components/storageWrappers";
 
 export interface IStorageRoot extends ISubscriptionContext {
     storageAccount: StorageAccountWrapper;
-    createBlobService(): azureStorage.BlobService;
+    createBlobServiceClient(): azureStorageBlob.BlobServiceClient;
+    createBlobContainerClient(containerName: string): azureStorageBlob.ContainerClient;
```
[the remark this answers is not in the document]
wwlorey: Would it be better to move them to `blobUtils`?

src/azureStorageExplorer/blobContainers/blobContainerNode.ts
```diff
@@ -452,19 +453,41 @@ export class BlobContainerTreeItem extends AzureParentTreeItem<IStorageRoot> imp
     }
 
     private async uploadFileToBlockBlob(filePath: string, blobPath: string, suppressLogs: boolean = false): Promise<void> {
-        let blobFriendlyPath = `${this.friendlyContainerName}/${blobPath}`;
+        const blobFriendlyPath = `${this.friendlyContainerName}/${blobPath}`;
```
nturinski: Could you add typings wherever you can?  I'm sure we'll inevitably turn on that tslint rule eventually and it'd just be easier to do it as we go

src/azureStorageExplorer/blobContainers/blobContainerNode.ts
```diff
@@ -452,19 +453,41 @@ export class BlobContainerTreeItem extends AzureParentTreeItem<IStorageRoot> imp
...
-        };
-        await blockBlobClient.uploadFile(filePath, options);
+        await window.withProgress({ title: `Uploading ${filePath} as ${blobFriendlyPath}`, location: ProgressLocation.Notification }, async (notificationProgress) => {
+            const options = {
+                blobHTTPHeaders: {
+                    // tslint:disable-next-line: strict-boolean-expressions
+                    blobContentType: mime.getType(blobPath) || undefined
+                },
+                onProgress: (progress: TransferProgressEvent) => {
+                    if (lastUpdated + updateTimerMs < Date.now()) {
```
nturinski: Write a comment explaining why you need to debounce the progress.report.

src/azureStorageExplorer/blobContainers/blobFileHandler.ts
```diff
@@ -56,7 +56,34 @@ export class BlobFileHandler implements IRemoteFileHandler<BlobTreeItem> {
...
+        // tslint:disable-next-line: strict-boolean-expressions
+        const totalBytes = (await blockBlobClient.getProperties()).contentLength || 1;
+
+        ext.outputChannel.show();
+        ext.outputChannel.appendLine(`Downloading ${treeItem.blob.name} to ${filePath}...`);
+
+        await window.withProgress({ title: `Downloading ${treeItem.blob.name}`, location: ProgressLocation.Notification }, async (notificationProgress) => {
+            await blockBlobClient.downloadToFile(filePath, undefined, undefined, {
+                onProgress: (progress) => {
+                    if (lastUpdated + updateTimerMs < Date.now()) {
```
nturinski: Since you're using this in two different spots, it probably makes sense to extract this to another file and just reuse it.

src/azureStorageExplorer/blobContainers/blobContainerGroupNode.ts
```diff
@@ -70,26 +62,31 @@ export class BlobContainerGroupTreeItem extends AzureParentTreeItem<IStorageRoot
...
-                }
-            });
-        });
+    private async createBlobContainer(name: string): Promise<azureStorageBlob.ContainerItem> {
+        const containerClient = this.root.createBlobContainerClient(name);
+        await containerClient.create();
+
+        let containersResponse = await this.listContainers();
+        let createdContainer: azureStorageBlob.ContainerItem | undefined;
+        for (let container of containersResponse.containerItems) {
```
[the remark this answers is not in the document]
wwlorey: `containerClient.exists()` only returns a bool and I needed a reference to the container here

src/azureStorageExplorer/blobContainers/blobDirectoryNode.ts
```diff
@@ -0,0 +1,104 @@
...
+import { BlobTreeItem, ISuppressMessageContext } from "./blobNode";
+import { createChildAsNewBlockBlob, IBlobContainerCreateChildContext, loadMoreBlobChildren } from './blobUtils';
+
+export class BlobDirectoryTreeItem extends AzureParentTreeItem<IStorageRoot> {
+    private _continuationToken: string | undefined;
+
+    public static contextValue: string = 'azureBlobDirectory';
+    public contextValue: string = 'azureBlobDirectory';
+
+    public fullPath: string = path.posix.join(this.parentPath, this.directory.name);
```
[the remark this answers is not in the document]
wwlorey: It's the blob's path on the user OS. I'll update that

src/azureStorageExplorer/blobContainers/blobUtils.ts
```diff
@@ -3,56 +3,130 @@
...
+import { BlobContainerTreeItem } from './blobContainerNode';
+import { BlobDirectoryTreeItem } from './blobDirectoryNode';
+import { BlobTreeItem } from './blobNode';
 
-export async function updateBlockBlobFromLocalFile(blobName: string, containerName: string, root: IStorageRoot, filePath: string): Promise<void> {
-    const createOptions = await getExistingCreateOptions(blobName, containerName, root);
-    await createBlockBlobFromLocalFile(blobName, containerName, root, filePath, createOptions);
+export async function loadMoreBlobChildren(parent: BlobContainerTreeItem | BlobDirectoryTreeItem, continuationToken?: string): Promise<{ children: AzExtTreeItem[], continuationToken?: string }> {
+    const dirPath = parent instanceof BlobDirectoryTreeItem ? parent.dirPath : '';
+    const listOptions = dirPath ? { prefix: dirPath } : undefined;
```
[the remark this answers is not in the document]
wwlorey: Prefix is OK to be undefined, I'll simplify that logic

src/azureStorageExplorer/IStorageRoot.ts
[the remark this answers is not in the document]
nturinski: Yeah, I think so.  Eventually we'll have to have similar functions for File, Queue, and Table as well and this file seems like it'd get kind of messy in functionality.

src/azureStorageExplorer/blobContainers/blobContainerGroupNode.ts
```diff
@@ -70,26 +63,31 @@ export class BlobContainerGroupTreeItem extends AzureParentTreeItem<IStorageRoot
...
+            if (container.name === name) {
+                createdContainer = container;
+                break;
+            }
+        }
+
+        if (!createdContainer) {
+            throw new Error(`Could not find container ${name}`);
```
nturinski: Rather not could not find, I think the error should indicate that there was an error in the creation itself.  I'd assume if we can't find it, it means it wasn't created properly?

src/azureStorageExplorer/blobContainers/blobContainerNode.ts
```diff
@@ -68,14 +68,13 @@ export class BlobContainerTreeItem extends AzureParentTreeItem<IStorageRoot> imp
     public contextValue: string = BlobContainerTreeItem.contextValue;
 
     public hasMoreChildrenImpl(): boolean {
-        return !!this._continuationTokenBlob || !!this._continuationTokenDirectory;
+        return !!this._continuationToken;
     }
 
     public async loadMoreChildrenImpl(clearCache: boolean): Promise<AzExtTreeItem[]> {
-        const result: AzExtTreeItem[] = [];
+        let result: AzExtTreeItem[] = [];
```
nturinski: Why did you change this from const to let?

wwlorey: Not sure why I did that 🤦‍♂️ I'll change it back

src/azureStorageExplorer/blobContainers/blobContainerNode.ts
```diff
@@ -115,50 +108,23 @@ export class BlobContainerTreeItem extends AzureParentTreeItem<IStorageRoot> imp
         this._websiteHostingEnabled = hostingStatus.enabled;
     }
 
-    // tslint:disable-next-line:promise-function-async // Grandfathered in
-    listBlobs(currentToken: azureStorage.common.ContinuationToken, maxResults: number = 50): Promise<azureStorage.BlobService.ListBlobsResult> {
-        return new Promise<azureStorage.BlobService.ListBlobsResult>((resolve, reject) => {
-            console.log(`${new Date().toLocaleTimeString()}: Querying Azure... Method: listBlobsSegmentedWithPrefix blobContainerName: "${this.container.name}" prefix: ""`);
```
nturinski: Do you know if you were supposed to remove this output with this PR?

src/azureStorageExplorer/blobContainers/blobContainerNode.ts
```diff
@@ -115,50 +108,23 @@ export class BlobContainerTreeItem extends AzureParentTreeItem<IStorageRoot> imp
...
 
         // tslint:disable-next-line:no-constant-condition
         while (true) {
             this.throwIfCanceled(cancellationToken, properties, "listAllBlobs");
-            // currentToken argument typed incorrectly in SDK
-            let result = await this.listBlobs(<azureStorage.common.ContinuationToken>currentToken, 5000);
-            blobs.push(...result.entries);
-            currentToken = result.continuationToken;
+
+            const containerClient: azureStorageBlob.ContainerClient = createBlobContainerClient(this.root, this.container.name);
```
nturinski: I think this can go outside the while loop so you're not creating a new containerClient with each iteration

src/azureStorageExplorer/blobContainers/blobContainerNode.ts
wwlorey: Some SDK calls had those logs and others didn't so I removed them when I consolidated functions. Do you think that would be good to add back in?

nturinski: I'm not sure how busy the output looks today.  When does it output?  I don't see it when I expand a blobContainer tree item.

But also, if you do add it back in, you can leverage ext.outputChannel.appendLog() to have the timestamp prefixed to the output message (rather than manually inserting it as it was coded before)

src/azureStorageExplorer/blobContainers/blobContainerNode.ts
```diff
@@ -115,50 +108,23 @@ export class BlobContainerTreeItem extends AzureParentTreeItem<IStorageRoot> imp
...
-        });
-    }
-
-    private async listAllBlobs(cancellationToken?: vscode.CancellationToken, properties?: TelemetryProperties): Promise<azureStorage.BlobService.BlobResult[]> {
+    private async listAllBlobs(cancellationToken?: vscode.CancellationToken, properties?: TelemetryProperties): Promise<azureStorageBlob.BlobItem[]> {
         // tslint:disable-next-line:no-any
-        let currentToken: azureStorage.common.ContinuationToken | undefined;
-        let blobs: azureStorage.BlobService.BlobResult[] = [];
+        let currentToken: string | undefined;
+        let blobs: azureStorageBlob.BlobItem[] = [];
```
nturinski: nit: This could be const too since it's never reassigned (probably should enable that tslint rule)

src/azureStorageExplorer/blobContainers/blobContainerNode.ts
```diff
@@ -171,14 +137,8 @@ export class BlobContainerTreeItem extends AzureParentTreeItem<IStorageRoot> imp
...
-            await new Promise((resolve, reject) => {
-                // tslint:disable-next-line:no-any
-                blobService.deleteContainer(this.container.name, (err?: any) => {
-                    // tslint:disable-next-line:no-void-expression // Grandfathered in
-                    err ? reject(err) : resolve();
-                });
-            });
+            const containerClient: azureStorageBlob.ContainerClient = createBlobContainerClient(this.root, this.container.name);
+            await containerClient.delete();
         } else {
```
nturinski: This can be removed now.  ext.ui.showWarningMessage handles userCancelled errors for us.

https://github.com/microsoft/vscode-azuretools/blob/master/ui/src/AzureUserInput.ts#L86-L87

src/azureStorageExplorer/blobContainers/blobContainerNode.ts
wwlorey: That outputs during a static website deployment when all blobs to delete are found. I can add it back in

src/azureStorageExplorer/blobContainers/blobDirectoryNode.ts
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    private async deleteFolder(context: IActionContext): Promise<boolean> {
+        let dirPaths: BlobDirectoryTreeItem[] = [];
+        // tslint:disable-next-line: no-var-self
+        let dirPath: BlobDirectoryTreeItem | undefined = this;
+        let errors: boolean = false;
+
+        // tslint:disable-next-line: strict-boolean-expressions
+        while (dirPath) {
```
nturinski: I'm not sure if this was like this before or not, but why use a while loop instead of just using a for loop that iterates over dirPaths?

wwlorey: This function wasn't modified by the SDK changes. It looks like they went with a while loop to handle folders & nested folders more easily

src/azureStorageExplorer/blobContainers/blobDirectoryNode.ts
```diff
@@ -0,0 +1,104 @@
...
+import { IStorageRoot } from "../IStorageRoot";
+import { BlobContainerTreeItem, IExistingBlobContext } from "./blobContainerNode";
+import { BlobTreeItem, ISuppressMessageContext } from "./blobNode";
+import { createChildAsNewBlockBlob, IBlobContainerCreateChildContext, loadMoreBlobChildren } from './blobUtils';
+
+export class BlobDirectoryTreeItem extends AzureParentTreeItem<IStorageRoot> {
+    private _continuationToken: string | undefined;
+
+    public static contextValue: string = 'azureBlobDirectory';
+    public contextValue: string = 'azureBlobDirectory';
```
nturinski: Should be able to just change this to `= BlobDirectoryTreeItem.contextValue` rather than having repeated strings.

src/azureStorageExplorer/blobContainers/blobUtils.ts
```diff
@@ -3,56 +3,176 @@
...
+
+        notificationProgress.report({ message: state.message, increment: state.percentage - state.lastPercentage });
+
+        state.lastPercentage = state.percentage;
+        state.lastUpdated = Date.now();
+    }
+}
+
+export class TransferProgressState {
+    public readonly updateTimerMs: number = 200;
```
nturinski: This should also be something that you can configure.  I can see instances where people would only want the % to update every second or something.

src/azureStorageExplorer/blobContainers/blobUtils.ts
wwlorey: Per offline conversation: reading properties from some resource returns a lot of info we don't explicitly need when setting new properties. Also the `BlobGetPropertiesResponse` object structure is different from `BlockBlobUploadOptions` needed when uploading resources

src/azureStorageExplorer/blobContainers/blobUtils.ts
```diff
@@ -3,56 +3,176 @@
...
+
+    constructor(
+        public percentage: number = 0,
```
nturinski: I feel like that these should all not be in the constructor actually.  I'm not sure what use-case you would have where you create a new TransferProgressState with these already set.

I think it might make sense to make handleTransferProgress a function of TransferProgressState and make all of those properties private so that you ensure nothing else is setting the properties except handleTransferProgress.

src/azureStorageExplorer/storageAccounts/storageAccountNode.ts
```diff
@@ -302,23 +271,26 @@ export class StorageAccountTreeItem extends AzureParentTreeItem<IStorageRoot> {
...
             value: oldStatus.errorDocument404Path ? oldStatus.errorDocument404Path : "",
             placeHolder: 'e.g. error/documents/error.html',
             validateInput: (value: string): string | undefined => this.validateErrorDocumentName(value)
-        });
-        let newStatus: azureStorage.common.models.ServicePropertiesResult.StaticWebsiteProperties = {
-            Enabled: true,
-            ErrorDocument404Path: errorDocument404Path,
-            IndexDocument: indexDocument
+            // tslint:disable-next-line: strict-boolean-expressions
+        }) || undefined;
```
nturinski: Why was this logic added?

wwlorey: Per offline discussion: `errorDocument404Path` can't be an empty string, so default to undefined. Also `value` passed into `showInputBox` must be of type string

PR #514, opened 2019-11-19
Add choices for how to "Open in File Explorer"

### LINKED ISSUE #434: Support "Add to workspace" and "Open in current window" for file view
Rather than always open in a new window, we might want to support three options:
1. Open in current window
1. Open in new window
1. Add to workspace

Similar to what Azure functions does when creating a new project
Screen Shot 2019-08-07 at 10 08 08 AM

Might help https://github.com/microsoft/vscode-azurestorage/issues/387

wwlorey:
Fixes #434. This implementation was lifted from functions.

Screen Shot 2019-11-19 at 3 30 04 PM


wwlorey:
Known issue: #515

src/localize.ts
```diff
@@ -0,0 +1,8 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as nls from 'vscode-nls';
```
EricJizbaMSFT: Add "vscode-nls" to dependencies in package.json since it's now referenced directly

src/azureStorageExplorer/openInFileExplorer/TreeItemStep.ts
```diff
@@ -0,0 +1,37 @@
...
+                {
+                    label: "Blob Container",
+                    data: BlobContainerTreeItem.contextValue
+                },
+                {
+                    label: "File Share",
+                    data: FileShareTreeItem.contextValue
+                }];
+            let contextValue: string = (await ext.ui.showQuickPick(quickPicks, { placeHolder })).data;
+            context.treeItem = <BlobContainerTreeItem | FileShareTreeItem>await ext.tree.showTreeItemPicker(contextValue, context);
```
EricJizbaMSFT: Uuhhh this probably has weird behavior with the back button - doesn't it? Since the tree item picker isn't _really_ in the wizard. Also as mentioned [here](https://github.com/microsoft/vscode-azurestorage/pull/488#discussion_r339791894), I think this prompt can go away and we can just use the treeItemPicker eventually. 

In conclusion, I don't think it's worth creating a separate step for this. Just keep this prompt as you had it before and only include `OpenBehaviorStep` in the prompt steps (and remove the wizard title since it would be weird to have a title for one step)

src/azureStorageExplorer/openInFileExplorer/OpenTreeItemStep.ts
```diff
@@ -0,0 +1,39 @@
...
+    public async execute(context: IOpenInFileExplorerWizardContext): Promise<void> {
+        // tslint:disable-next-line:strict-boolean-expressions
+        const openFolders: WorkspaceFolder[] = workspace.workspaceFolders || [];
+        if (context.openBehavior === 'AddToWorkspace' && openFolders.length === 0) {
+            // no point in adding to an empty workspace
+            context.openBehavior = 'OpenInCurrentWindow';
+        }
+
+        if (!context.treeItem) {
+            throw new Error('Could not locate resource to open.');
```
EricJizbaMSFT: Does this repo have `nonNullProp` like other repos? Either way, I would use that for the sake of consistency. Easier to search telemetry for "Internal error"

src/extension.ts
```diff
@@ -82,11 +87,17 @@ export async function activateInternal(context: vscode.ExtensionContext, perfSta
...
                             data: FileShareTreeItem.contextValue
                         }];
-                    let contextValue: string = (await ext.ui.showQuickPick(quickPicks, { placeHolder: "Select the resource type to open in File Explorer" })).data;
+                    let contextValue: string = (await ext.ui.showQuickPick(quickPicks, { placeHolder })).data;
                     treeItem = <BlobContainerTreeItem | FileShareTreeItem>await ext.tree.showTreeItemPicker(contextValue, _actionContext);
                 }
-                await commands.executeCommand('vscode.openFolder', AzureStorageFS.idToUri(treeItem.fullId));
-                await commands.executeCommand('workbench.view.explorer');
+
+                const wizardContext: Partial<IOpenInFileExplorerWizardContext> & IActionContext = Object.assign(_actionContext, { treeItem });
```
EricJizbaMSFT: nit: rename `_actionContext` to `actionContext`. We use an underscore similar to "tslint:disable" when lint complains about unused variables. Once it's used, we can remove that underscore

src/azureStorageExplorer/openInFileExplorer/IOpenInFileExplorerWizardContext.ts
```diff
@@ -0,0 +1,15 @@
...
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { IActionContext } from "vscode-azureextensionui";
+import { BlobContainerTreeItem } from '../blobContainers/blobContainerNode';
+import { FileShareTreeItem } from '../fileShares/fileShareNode';
+
+export interface IOpenInFileExplorerWizardContext extends IActionContext {
+    treeItem?: BlobContainerTreeItem | FileShareTreeItem;
```
EricJizbaMSFT: treeItem shouldn't be optional since you prompt for it before you create the wizard

src/extension.ts
EricJizbaMSFT: Can't this type just be `IOpenInFileExplorerWizardContext` instead of `Partial<IOpenInFileExplorerWizardContext> & IActionContext`?

Issue #505, opened 2019-11-08
Add basic creation mode


wwlorey:
Related to #344

> - [ ] Basic creation mode
>    - [x] Create resource group in West US
>    - [ ] Create set index.html for index and error documents when user chooses to enable static hosting

wwlorey:
**Offline discussion**: If deploying a static website via the blue arrow, don't ask to configure static website after creating/selecting a storage account

PR #519, opened 2019-12-03
Add basic create for storage accounts

wwlorey:
Related to #505. The old 'normal' create is now advanced create.

Screen Shot 2019-12-02 at 4 17 56 PM

Screen Shot 2019-12-02 at 4 16 41 PM


src/azureStorageExplorer/createStorageAccount.ts
```diff
@@ -0,0 +1,20 @@
...
+ *  Licensed under the MIT License. See License.md in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { IActionContext, ICreateChildImplContext } from 'vscode-azureextensionui';
+import { ext } from '../extensionVariables';
+import { SubscriptionTreeItem } from './SubscriptionTreeItem';
+
+export async function createStorageAccount(context: IActionContext & Partial<ICreateChildImplContext>, treeItem?: SubscriptionTreeItem): Promise<void> {
+    if (!treeItem) {
+        treeItem = <SubscriptionTreeItem>await ext.tree.showTreeItemPicker(SubscriptionTreeItem.contextValue, context);
```
EricJizbaMSFT: If you're doing this, you should be able to show these commands in the command palette

test/storageAccountActions.test.ts
```diff
@@ -61,13 +61,24 @@ suite('Storage Account Actions', async function (this: ISuiteCallbackContext): P
...
         resourceGroupsToDelete.push(resourceName);
-        await testUserInput.runWithInputs([resourceName, '$(plus) Create new resource group', resourceName, 'East US'], async () => {
+        await testUserInput.runWithInputs([resourceName], async () => {
             await vscode.commands.executeCommand('azureStorage.createGpv2Account');
         });
         const createdAccount: StorageAccount = await client.storageAccounts.getProperties(resourceName, resourceName);
         assert.ok(createdAccount);
     });
 
+    await deleteStorageAccount(resourceName);
```
EricJizbaMSFT: This is outside a test right? I don't think it'll get awaited. In any case, resources should already be cleaned up in `suiteTeardown` so I think you can just get rid of this line

src/azureStorageExplorer/SubscriptionTreeItem.ts
```diff
@@ -31,19 +30,31 @@ export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
...
             title: "Create storage account",
             promptSteps,
-            executeSteps: [new StorageAccountCreateStep({ kind: StorageAccountKind.StorageV2, performance: StorageAccountPerformance.Standard, replication: StorageAccountReplication.LRS })],
+            executeSteps
         });
 
         await wizard.prompt();
 
+        if (!context.advancedCreation) {
+            wizardContext.newResourceGroupName = `storage_${defaultLocation}`;
```
EricJizbaMSFT: The rg name should be based on the storage account name. Similar to this logic: https://github.com/microsoft/vscode-azureappservice/blob/master/src/explorer/SubscriptionTreeItem.ts#L108

test/storageAccountActions.test.ts
```diff
@@ -61,13 +61,24 @@ suite('Storage Account Actions', async function (this: ISuiteCallbackContext): P
...
+
+    test("createStorageAccountAdvanced", async () => {
```
EricJizbaMSFT: Have you run a nightly test in Dev Ops with your changes to make sure these pass?

test/storageAccountActions.test.ts
```diff
@@ -22,6 +22,7 @@ suite('Storage Account Actions', async function (this: ISuiteCallbackContext): P
     const testAccount: TestAzureAccount = new TestAzureAccount(vscode);
     let client: StorageManagementClient;
     const resourceName: string = getRandomHexString().toLowerCase();
+    const resourceNameAdvanced: string = getRandomHexString().toLowerCase(); // Used when testing advanced create
```
EricJizbaMSFT: Please move this constant into the test for `createStorageAccountAdvanced`. AFAIK the only reasons these other constants are outside a test is because they're used in multiple tests

test/storageAccountActions.test.ts
```diff
@@ -63,13 +64,22 @@ suite('Storage Account Actions', async function (this: ISuiteCallbackContext): P
...
+        await testUserInput.runWithInputs([resourceName], async () => {
             await vscode.commands.executeCommand('azureStorage.createGpv2Account');
         });
         const createdAccount: StorageAccount = await client.storageAccounts.getProperties(resourceName, resourceName);
         assert.ok(createdAccount);
     });
 
+    test("createStorageAccountAdvanced", async () => {
+        resourceGroupsToDelete.push(resourceNameAdvanced);
+        await testUserInput.runWithInputs([resourceNameAdvanced, '$(plus) Create new resource group', resourceNameAdvanced, 'East US'], async () => {
```
EricJizbaMSFT: To more throughly test advanced creation, you should use different names for the storage account and resource group. For example, this wouldn't catch a bug if resource group name is ignored and storage account name is used for both (similar to basic create behavior).

PR #529, opened 2019-12-11
Add a constant for maxPageSize

EricJizbaMSFT:
To ensure we're consistent and to make it easier when debugging things related to "Load More"

### DIFF
--- a/src/azureStorageExplorer/blobContainers/BlobContainerGroupTreeItem.ts
+++ b/src/azureStorageExplorer/blobContainers/BlobContainerGroupTreeItem.ts
@@ -8,7 +8,7 @@ import * as path from 'path';
 import * as vscode from 'vscode';
 import { Uri } from 'vscode';
 import { AzExtTreeItem, AzureParentTreeItem, ICreateChildImplContext, UserCancelledError } from 'vscode-azureextensionui';
-import { getResourcesPath } from "../../constants";
+import { getResourcesPath, maxPageSize } from "../../constants";
 import { ext } from "../../extensionVariables";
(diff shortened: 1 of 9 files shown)


src/azureStorageExplorer/blobContainers/BlobContainerGroupTreeItem.ts
```diff
@@ -45,9 +45,9 @@ export class BlobContainerGroupTreeItem extends AzureParentTreeItem<IStorageRoot
         return !!this._continuationToken;
     }
 
-    async listContainers(continuationToken?: string, maxPageSize: number = 50): Promise<azureStorageBlob.ListContainersSegmentResponse> {
+    async listContainers(continuationToken?: string): Promise<azureStorageBlob.ListContainersSegmentResponse> {
         const blobServiceClient: azureStorageBlob.BlobServiceClient = this.root.createBlobServiceClient();
-        let response: AsyncIterableIterator<azureStorageBlob.ServiceListContainersSegmentResponse> = blobServiceClient.listContainers().byPage({ continuationToken, maxPageSize });
+        let response: AsyncIterableIterator<azureStorageBlob.ServiceListContainersSegmentResponse> = blobServiceClient.listContainers().byPage({ continuationToken, maxPageSize: maxPageSize });
```
wwlorey: nit: Since the property name and constant name are the same we could do

```suggestion
        let response: AsyncIterableIterator<azureStorageBlob.ServiceListContainersSegmentResponse> = blobServiceClient.listContainers().byPage({ continuationToken, maxPageSize });
```

PR #531, opened 2019-12-12
Add telemetry for opening Storage Explorer

### LINKED ISSUE #84: Need better failure error if checking registry for StorageExplorer.exe fails
private static async getStorageExplorerExecutable(): Promise<string> {
         let regVal: string;
         try {
             regVal = await WindowsStorageExplorerLauncher.getWindowsRegistryValue(regKey.hive, regKey.key);
        } catch (err) {
             // ignore and prompt to download.
            console.error(err); <<<<<<<<<<<<<
         } finally {

Preferably also some telemetry.

wwlorey:
Fixes #84


src/storageExplorerLauncher/windowsStorageExplorerLauncher.ts
```diff
@@ -55,13 +56,15 @@ export class WindowsStorageExplorerLauncher implements IStorageExplorerLauncher
             } else {
                 let selected: "Download" = <"Download">await vscode.window.showWarningMessage("Cannot find a compatible Storage Explorer. Would you like to download the latest Storage Explorer?", "Download");
```
EricJizbaMSFT: I would think `context.telemetry.properties.storageExplorerNotFound = 'true';` should be in this case as well

src/storageExplorerLauncher/macOSStorageExplorerLauncher.ts
```diff
@@ -19,25 +19,30 @@ export class MacOSStorageExplorerLauncher implements IStorageExplorerLauncher {
...
-                let userSelectedAppLocation = await MacOSStorageExplorerLauncher.showOpenDialog();
-                await vscode.workspace.getConfiguration('azureStorage').update('storageExplorerLocation', userSelectedAppLocation, vscode.ConfigurationTarget.Global);
-                return await MacOSStorageExplorerLauncher.getStorageExplorerExecutable("Selected app is not a valid Storage Explorer installation. Browse to existing installation location or download and install Storage Explorer.");
-            } else if (selected === "Download") {
-                await MacOSStorageExplorerLauncher.downloadStorageExplorer();
-                throw new UserCancelledError();
-            } else {
-                throw new UserCancelledError();
+        return await callWithTelemetryAndErrorHandling('getStorageExplorerExecutableMac', async context => {
+            let selectedLocation = vscode.workspace.getConfiguration('azureStorage').get<string>('storageExplorerLocation');
```
EricJizbaMSFT: Just to confirm - how often is `getStorageExplorerExecutable` called? Is it an "activationEvent" and should we mark it that way? If it's called super often, we might not want all this telemetry

PR #540, opened 2019-12-17
Don't display progress notifications when uploading blobs

### LINKED ISSUE #536: The notification dialog keeps flashing when deploying a project to a static website
**OS**: Win10 & Mac
**Build Version:** [20191216.1](https://dev.azure.com/ms-azuretools/AzCode/_build/results?buildId=13157&view=results)
**Regression**: Previous Release (0.7.2)

**Repro Steps:**
1. Create a new storage account -->Right click the new created storage account  --> Select "Deploy to Static Website...".
2. Enable website hosting --> Enter the index document name -->Enter the 404 error document path --> Select a folder to deploy.
3. Check the deploying.

**Expect:**
Easy to see the processing of deploying.
image

**Actual:**
The notification dialog keeps flashing.

wwlorey:
An alternative to this approach would be to suppress notifications if `suppressLogs` is true. But this implementation seemed cleaner and returns us to the old functionality before the SDK update.

Fixes #536

### DIFF
--- a/src/editors/BlobFileHandler.ts
+++ b/src/editors/BlobFileHandler.ts
@@ -67,7 +67,7 @@ export class BlobFileHandler implements IRemoteFileHandler<BlobTreeItem> {
         await window.withProgress({ title: `Downloading ${treeItem.blob.name}`, location: ProgressLocation.Notification }, async (notificationProgress) => {
             const transferProgress: TransferProgress = new TransferProgress();
             await blockBlobClient.downloadToFile(filePath, undefined, undefined, {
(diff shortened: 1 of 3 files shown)


EricJizbaMSFT:
Can you give examples of what the logs looked like before and what they look like now?

wwlorey:
Here's the old logs:
Screen Shot 2019-12-19 at 2 13 26 PM

And new logs:
Screen Shot 2019-12-19 at 2 11 43 PM

Blob deletion logs were removed in #513, do we want those back?

EricJizbaMSFT:
> Blob deletion logs were removed in #513, do we want those back?

Yeah I think so

PR #558, opened 2020-01-08
Update ⬆️ to newest Azure Storage SDK for file shares 🗂

### LINKED ISSUE #328: Switch over to azure-sdk-for-js

wwlorey:
Fixes #328

### DIFF
--- a/src/tree/fileShare/DirectoryTreeItem.ts
+++ b/src/tree/fileShare/DirectoryTreeItem.ts
@@ -48,7 +48,7 @@ export class DirectoryTreeItem extends AzureParentTreeItem<IStorageRoot> impleme
             this._continuationToken = undefined;
         }
 
-        let { files, directories, continuationToken }: { files: azureStorageShare.FileItem[]; directories: azureStorageShare.DirectoryItem[]; continuationToken: string; } = await this.listFiles(this._continuationToken);
+        let { files, directories, continuationToken }: { files: azureStorageShare.FileItem[]; directories: azureStorageShare.DirectoryItem[]; continuationToken: string; } = await listFilesInDirectory(this.fullPath, this.shareName, this.root, this._continuationToken);
         this._continuationToken = continuationToken;
 
         return (<(DirectoryTreeItem | FileTreeItem)[]>[])
@@ -63,18 +63,10 @@ export class DirectoryTreeItem extends AzureParentTreeItem<IStorageRoot> impleme
     public async copyUrl(): Promise<void> {
         // Use this.fullPath here instead of this.directoryName. Otherwise only the leaf directory is displayed in the URL
         let directoryClient: azureStorageShare.ShareDirectoryClient = createDirectoryClient(this.root, this.shareName, this.fullPath);
-
-        // URLs for nested directories aren't automatically decoded properly
-        const url = decodeURIComponent(directoryClient.url);
-
+        const url = directoryClient.url;
         await vscode.env.clipboard.writeText(url);
         ext.outputChannel.show();
         ext.outputChannel.appendLine(`Directory URL copied to clipboard: ${url}`);
-    }
-
-    // tslint:disable-next-line:promise-function-async // Grandfathered in
-    listFiles(currentToken: string | undefined): Promise<{ files: azureStorageShare.FileItem[], directories: azureStorageShare.DirectoryItem[], continuationToken: string }> {
-        return listFilesInDirectory(this.fullPath, this.shareName, this.root, currentToken);
     }
 
     public async createChildImpl(context: ICreateChildImplContext & IFileShareCreateChildContext): Promise<FileTreeItem | DirectoryTreeItem> {
--- a/src/tree/fileShare/DirectoryTreeItem.ts
+++ b/src/tree/fileShare/DirectoryTreeItem.ts
@@ -3,7 +3,7 @@
  *  Licensed under the MIT License. See License.md in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import * as azureStorage from "azure-storage";
+import * as azureStorageShare from '@azure/storage-file-share';
 import * as path from 'path';
 import * as vscode from 'vscode';
 import { MessageItem, Uri, window } from 'vscode';
@@ -11,7 +11,7 @@ import { AzureParentTreeItem, DialogResponses, IActionContext, ICreateChildImplC
 import { getResourcesPath } from "../../constants";
 import { ext } from "../../extensionVariables";
 import { askAndCreateChildDirectory, deleteDirectoryAndContents, listFilesInDirectory } from '../../utils/directoryUtils';
-import { askAndCreateEmptyTextFile } from '../../utils/fileUtils';
+import { askAndCreateEmptyTextFile, createDirectoryClient } from '../../utils/fileUtils';
 import { ICopyUrl } from '../ICopyUrl';
 import { IStorageRoot } from "../IStorageRoot";
 import { IFileShareCreateChildContext } from "./FileShareTreeItem";
@@ -21,13 +21,13 @@ export class DirectoryTreeItem extends AzureParentTreeItem<IStorageRoot> impleme
     constructor(
         parent: AzureParentTreeItem,
         public readonly parentPath: string,
-        public readonly directory: azureStorage.FileService.DirectoryResult, // directory.name should not include parent path
-        public readonly share: azureStorage.FileService.ShareResult) {
+        public readonly directoryName: string, // directoryName should not include parent path
+        public readonly shareName: string) {
         super(parent);
     }
 
-    private _continuationToken: azureStorage.common.ContinuationToken | undefined;
-    public label: string = this.directory.name;
+    private _continuationToken: string | undefined;
+    public label: string = this.directoryName;
     public static contextValue: string = 'azureFileShareDirectory';
     public contextValue: string = DirectoryTreeItem.contextValue;
     public iconPath: { light: string | Uri; dark: string | Uri } = {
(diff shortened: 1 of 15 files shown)


src/tree/fileShare/DirectoryTreeItem.ts
```diff
@@ -48,38 +48,40 @@ export class DirectoryTreeItem extends AzureParentTreeItem<IStorageRoot> impleme
...
             }));
     }
 
     public async copyUrl(): Promise<void> {
-        let fileService = this.root.createFileService();
-        let url = fileService.getUrl(this.share.name, this.fullPath);
+        // Use this.fullPath here instead of this.directoryName. Otherwise only the leaf directory is displayed in the URL
+        let directoryClient: azureStorageShare.ShareDirectoryClient = createDirectoryClient(this.root, this.shareName, this.fullPath);
+
+        // URLs for nested directories aren't automatically decoded properly
```
EricJizbaMSFT: I feel like this is probably a bug in the sdk. Can we file an issue on them and link to it from this comment?

src/tree/fileShare/FileShareGroupTreeItem.ts
```diff
@@ -33,34 +32,40 @@ export class FileShareGroupTreeItem extends AzureParentTreeItem<IStorageRoot> {
...
-            let fileService = this.root.createFileService();
-            fileService.listSharesSegmented(currentToken, { maxResults: maxPageSize }, (err?: Error, result?: azureStorage.FileService.ListSharesResult) => {
-                if (err) {
-                    reject(err);
-                } else {
-                    resolve(result);
-                }
-            });
-        });
+    async listFileShares(currentToken: string | undefined): Promise<{ shares: azureStorageShare.ShareItem[], continuationToken: string }> {
```
EricJizbaMSFT: I'm not sure if this was intentional or for your own debugging, but this method actually loads _all_ shares instead of just one page. Once you fix it to do a single page, the code is actually pretty simple and I'd prefer if it was in `loadMoreChildrenImpl` instead of a separate method.

src/tree/fileShare/FileShareGroupTreeItem.ts
```diff
@@ -80,30 +85,27 @@ export class FileShareGroupTreeItem extends AzureParentTreeItem<IStorageRoot> {
                 return await window.withProgress({ location: ProgressLocation.Window }, async (progress) => {
                     context.showCreatingTreeItem(shareName);
                     progress.report({ message: `Azure Storage: Creating file share '${shareName}'` });
-                    const share = await this.createFileShare(shareName, Number(quotaGB));
-                    return new FileShareTreeItem(this, share);
+                    const shareResponse = await this.createFileShare(shareName, Number(quotaGB));
+
+                    if (shareResponse.errorCode) {
+                        throw new Error(`Could not create share ${shareName}. ${shareResponse.errorCode}`);
```
EricJizbaMSFT: We have to check this ourselves? Previously the sdk would validate error codes

src/tree/fileShare/FileShareGroupTreeItem.ts
```diff
@@ -80,30 +85,27 @@ export class FileShareGroupTreeItem extends AzureParentTreeItem<IStorageRoot> {
...
-            };
-            shareService.createShare(name, options, (err?: Error, result?: azureStorage.FileService.ShareResult) => {
-                if (err) {
-                    reject(err);
-                } else {
-                    resolve(result);
-                }
-            });
-        });
+    private async createFileShare(name: string, quotaGB: number): Promise<azureStorageShare.ShareCreateResponse> {
```
EricJizbaMSFT: Without the `new Promise` thing, this is a pretty simple function and it's only used once. I'd prefer if you just moved the code up into the calling method. There's a few other spots like this throughout the PR, including in directoryUtils

test/storageAccountActions.test.ts
```diff
@@ -232,6 +242,13 @@ suite('Storage Account Actions', async function (this: ISuiteCallbackContext): P
         });
         return await vscode.env.clipboard.readText();
     }
+
+    async function createShareClient(_resourceName: string, _shareName: string): Promise<azureStorageShare.ShareClient> {
```
EricJizbaMSFT: nit: Remove `_` from the variable names. We only do that for unused variables

src/utils/fileUtils.ts
```diff
@@ -33,126 +47,79 @@ export async function askAndCreateEmptyTextFile(parent: AzureParentTreeItem<ISto
...
-
-    return {
-        contentSettings: propertiesResult.contentSettings,
-        metadata: metadataResult.metadata
-    };
+async function getExistingCreateOptions(directoryPath: string, name: string, shareName: string, root: IStorageRoot): Promise<azureStorageShare.FileCreateOptions> {
+    const fileClient: azureStorageShare.ShareFileClient = createFileClient(root, shareName, directoryPath, name);
+    let propertiesResult: azureStorageShare.FileGetPropertiesResponse = await fileClient.getProperties();
+    let options: azureStorageShare.FileCreateOptions = {};
+    options.fileHttpHeaders = {};
```
EricJizbaMSFT: So I missed the conversation when I was OOF on Thanksgiving on why we're doing it like this. If it's a limitation of the new sdk, I think we should file an issue because the old sdk handled this much better. Does the Storage Explorer team use this version of the sdk and do they do the same thing?

src/utils/fileUtils.ts
```diff
@@ -33,126 +47,79 @@ export async function askAndCreateEmptyTextFile(parent: AzureParentTreeItem<ISto
...
 }
 
-export async function updateFileFromText(directoryPath: string, name: string, share: FileService.ShareResult, root: IStorageRoot, text: string | Buffer): Promise<azureStorage.FileService.FileResult> {
-    const options = await getExistingCreateOptions(directoryPath, name, share, root);
-    return await createFile(directoryPath, name, share, root, text, options);
+export async function updateFileFromText(directoryPath: string, name: string, shareName: string, root: IStorageRoot, text: string | Buffer): Promise<void> {
+    const options: azureStorageShare.FileCreateOptions = await getExistingCreateOptions(directoryPath, name, shareName, root);
+    const fileClient: azureStorageShare.ShareFileClient = createFileClient(root, shareName, directoryPath, name);
+
+    // uploadData hangs indefinitely if the text length is zero
```
EricJizbaMSFT: confused

Can we file an issue on the sdk? Also the fact that you have to delete/create below worries me

src/utils/directoryUtils.ts
```diff
@@ -26,60 +26,57 @@ export async function askAndCreateChildDirectory(parent: AzureParentTreeItem<ISt
...
-    });
+export async function listFilesInDirectory(directory: string, shareName: string, root: IStorageRoot, currentToken?: string): Promise<{ files: azureStorageShare.FileItem[], directories: azureStorageShare.DirectoryItem[], continuationToken: string }> {
+    let responseValue: azureStorageShare.DirectoryListFilesAndDirectoriesSegmentResponse;
+    let files: azureStorageShare.FileItem[] = [];
+    let directories: azureStorageShare.DirectoryItem[] = [];
+    const directoryClient: azureStorageShare.ShareDirectoryClient = createDirectoryClient(root, shareName, directory);
+    const response: AsyncIterableIterator<azureStorageShare.DirectoryListFilesAndDirectoriesSegmentResponse> = directoryClient.listFilesAndDirectories().byPage({ continuationToken: currentToken, maxPageSize });
+
+    // tslint:disable-next-line: no-constant-condition
+    while (true) {
```
EricJizbaMSFT: Similar feedback about loading all vs. one page

src/tree/fileShare/DirectoryTreeItem.ts
wwlorey: No need to decode after all - encoded URLs work

https://github.com/Azure/azure-sdk-for-js/issues/6886

src/utils/fileUtils.ts
wwlorey: Offline discussion: use the old SDK for this one function

src/utils/directoryUtils.ts
```diff
@@ -26,60 +26,39 @@ export async function askAndCreateChildDirectory(parent: AzureParentTreeItem<ISt
...
-        fileService.listFilesAndDirectoriesSegmented(share, directory, <azureStorage.common.ContinuationToken>currentToken, { maxResults: maxPageSize }, (err?: Error, result?: azureStorage.FileService.ListFilesAndDirectoriesResult) => {
-            if (err) {
-                reject(err);
-            } else {
-                resolve(result);
-            }
-        });
-    });
+    // tslint:disable-next-line: no-unsafe-any
+    let responseValue: azureStorageShare.DirectoryListFilesAndDirectoriesSegmentResponse = (await response.next()).value;
```
EricJizbaMSFT: nit: Not sure I see the point in returning `{ files: azureStorageShare.FileItem[], directories: azureStorageShare.DirectoryItem[], continuationToken: string }` when you could just return `DirectoryListFilesAndDirectoriesSegmentResponse` directly and it has all the same information

wwlorey: I just liked saving `segment.fileItems` as `files`, etc. It seemed cleaner for functions that call `listFilesInDirectory` to deal with

src/utils/fileUtils.ts
jeremymeng: We'd like to know what is missing from the new sdk. /cc @XiaoningLiu

wwlorey: @jeremymeng It would be easier to update a file's contents while maintaining its metadata & properties if `FileCreateOptions` and `FileGetPropertiesResponse` had the same object structure. Either that or an update function that uploaded a file's new contents while preserving properties. 

Today we have to read the existing properties/metadata and include them when updating a file via `uploadFile`.

jeremymeng: @wwlorey agree. Thanks for the feedback! I logged an issue at https://github.com/azure/azure-sdk-for-js/issues/7097

PR #570, opened 2020-01-14
Fix multiple errors for invalid folders.url

### LINKED ISSUE #546: Display multiple errors when the folders.url is invalid for a blob container
**OS:** All
**Build Version:** [20191222.1](https://dev.azure.com/ms-azuretools/AzCode/_build/results?buildId=13295&view=results)

**Repro Steps:**
1. Enable "Azure Storage: Enable View In File Explorer".
2. Create a new storage account --> Create a blob container --> Open this blob container in File Explorer.
3. Save the workspace locally --> Set the folders.url as invalid like remove the blob container name.
4. Open the edited workspace in VS Code --> Pop up an error.
5. Check the Output window.

**Expect:**
Pop up one error.

**Actual:**
Display multiple errors.
image

**More Info:**
This issue doesn't reproduce for file share.

wwlorey:
For an invalid `folders.url` (created by removing the blob container name), `lookupRoot` is called many times for all the URIs present. Each time, an error would be thrown since the root `BlobContainerGroupTreeItem` didn't exist.

This fix displays only one `lookupRoot` error per `rootName`

Fixes #546

### DIFF
--- a/src/AzureStorageFS.ts
+++ b/src/AzureStorageFS.ts
@@ -34,6 +34,7 @@ export class AzureStorageFS implements vscode.FileSystemProvider {
     private _fireSoonHandle?: NodeJS.Timer;
 
     private _queryCache: Map<string, { query: string, invalid?: boolean }> = new Map(); // Key: rootName
+    private _rangeErrorCache: Map<string, boolean> = new Map(); // Key: rootName, value: has a RangeError been thrown for this rootName?
     readonly onDidChangeFile: vscode.Event<vscode.FileChangeEvent[]> = this._emitter.event;
 
     static idToUri(resourceId: string, filePath?: string): vscode.Uri {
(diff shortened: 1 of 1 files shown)


src/AzureStorageFS.ts
```diff
@@ -385,7 +386,20 @@ export class AzureStorageFS implements vscode.FileSystemProvider {
...
-            throw new RangeError(`Unexpected entry ${treeItem.constructor.name}.`);
+            // Ensure only one RangeError is displayed per rootName
+            if (!context.errorHandling.suppressDisplay) {
+                let cache = this._rangeErrorCache.get(rootName);
+
+                // A RangeError has already been or will be thrown for this rootName
+                this._rangeErrorCache.set(rootName, true);
+
+                if (!cache) {
+                    throw new RangeError(`Unexpected entry ${treeItem.constructor.name}.`);
```
EricJizbaMSFT: A RangeError like this is usually only meant for internal use so we can catch it in telemetry. If users are seeing it (which seems possible based on the repro steps), I think that's a sign we should handle this better. For example:
- Can you try throwing a `FileSystemError` and see if that prevents "lookupRoot is called many times" (since VS Code theoretically handles those errors differently than normal errors)
- Let's improve the message. I'm thinking something as simple as "Invalid uri" would be better

wwlorey: There's still one error per URI with `FileSystemError`. Would it be better to use that or just `Error`?

wwlorey: And if we use an error message like "invalid uri", would be display the uri as well? If so we may want to see all the error messages since each one would be different

EricJizbaMSFT: What do you mean "per uri"? How many uris are there and what are they?

wwlorey: Each of the errors happened because it tried to read some file as a `BlobContainerGroupTreeItem` or the equivalent for fie shares. I modified the original error message to show those files:

Screen Shot 2020-01-14 at 1 55 18 PM

I'm not sure where they are coming from. Maybe they're from where VS Code opened the workspace?

EricJizbaMSFT: Oh did you use `getFileSystemError` or FileSystemError directly? Check a few lines up for an example of `getFileSystemError`

EricJizbaMSFT: It makes sense that VS Code is looking for those files on startup. It needs to know it's own settings, git config, etc. because that directly affects some of VS Code's behavior. I believe the proper fix in this case is to throw a FileNotFound using `getFileSystemError` in effect telling VS Code "We ain't got no settings for you"

wwlorey: Using `getFileSystemError` does the trick 👌

PR #572, opened 2020-01-14
Display new nested blob names properly

### LINKED ISSUE #565: Only display the file name "blob.txt" when creating a new blob with name "this/is/a/blob.txt" without refreshing
**OS**: All
**Build Version:** [20200113.1](https://dev.azure.com/ms-azuretools/AzCode/_build/results?buildId=13678&view=results)
**Regression:** Previous Release (0.7.2)

**Repro Steps:**
1. Create a new storage account --> Create a new blob container --> Create an empty blob with name "this/is/a/blob.txt".
2. Check the new created blob name.

**Expect:**

Succeed to display the new created blob with full path "this/is/a/blob.txt".

**Actual:**
1. Only display the file name "blob.txt".
2. Display well with full path after refreshing.
image

wwlorey:
Fixes #565

### DIFF
--- a/src/tree/blob/BlobContainerTreeItem.ts
+++ b/src/tree/blob/BlobContainerTreeItem.ts
@@ -155,6 +155,9 @@ export class BlobContainerTreeItem extends AzureParentTreeItem<IStorageRoot> imp
             child = new BlobDirectoryTreeItem(this, context.childName, this.container);
         } else {
             child = await createChildAsNewBlockBlob(this, context);
+
+            // Refresh tree item to prevent https://github.com/microsoft/vscode-azurestorage/issues/565
+            await ext.tree.refresh(this);
         }
 
         AzureStorageFS.fireCreateEvent(child);
--- a/src/tree/blob/BlobContainerTreeItem.ts
+++ b/src/tree/blob/BlobContainerTreeItem.ts
@@ -155,9 +155,6 @@ export class BlobContainerTreeItem extends AzureParentTreeItem<IStorageRoot> imp
             child = new BlobDirectoryTreeItem(this, context.childName, this.container);
         } else {
             child = await createChildAsNewBlockBlob(this, context);
-
-            // Refresh tree item to prevent https://github.com/microsoft/vscode-azurestorage/issues/565
-            await ext.tree.refresh(this);
         }
 
         AzureStorageFS.fireCreateEvent(child);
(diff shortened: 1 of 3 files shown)


src/tree/blob/BlobContainerTreeItem.ts
```diff
@@ -155,6 +155,9 @@ export class BlobContainerTreeItem extends AzureParentTreeItem<IStorageRoot> imp
             child = new BlobDirectoryTreeItem(this, context.childName, this.container);
         } else {
             child = await createChildAsNewBlockBlob(this, context);
+
+            // Refresh tree item to prevent https://github.com/microsoft/vscode-azurestorage/issues/565
+            await ext.tree.refresh(this);
```
EricJizbaMSFT: We have to be careful about refreshing a tree item because it will clear it's cache of children. That means:
1. Users will have to wait _again_ if it took a long time to load
1. The file system might have errors since it expected all children to already be loaded

Do you have a blob container with lots of blobs? Hopefully you can see what I mean just by messing around with that, but let me know if you need more concrete repro steps

wwlorey: I was able to repro what you mean - I'll look for another solution

wwlorey:
It looks like when we updated the tree item class members for blobs it changed the way labels worked. 

This fix returns us to how it worked originally: when you create a nested blob, it displays in the tree view like this (with the full path)

Screen Shot 2020-01-15 at 11 18 21 AM

Then once the tree view is refreshed, it displays the hierarchical view (you have to expand the directories to see them all)

Screen Shot 2020-01-15 at 11 19 44 AM

src/tree/blob/BlobTreeItem.ts
```diff
@@ -31,14 +31,11 @@ export class BlobTreeItem extends AzureTreeItem<IStorageRoot> implements ICopyUr
      */
     public readonly blobPath: string;
 
-    constructor(parent: AzureParentTreeItem, blobPath: string, public readonly container: azureStorageBlob.ContainerItem) {
+    constructor(parent: AzureParentTreeItem, blobPath: string, public readonly container: azureStorageBlob.ContainerItem, public readonly label: string = '') {
```
EricJizbaMSFT: nit: I have to say it at least once - I'm not a fan of the "public readonly" pattern inside the constructor. I think it's confusing and ugly haha. I prefer having it as a property above.

But I think that's just personal preference so ¯\\\_(ツ)\_/¯ do what you want

wwlorey: Yeah it's a bit ugly - I'll keep that in mind for future PRs. I'd change it here but then `container` would probs need to change too 🤷‍♂️

PR #573, opened 2020-01-14
Add warning when deleting an already deleted storage account

### LINKED ISSUE #568: No error pops up when deleting a non-existing storage account
**OS**: All
**Build Version:** [20200113.1](https://dev.azure.com/ms-azuretools/AzCode/_build/results?buildId=13678&view=results)

**Repro Steps:**
1. Open 2 windows of VS Code --> Select and delete one storage account like "test" in one window.
2. Navigate to another window --> Try to delete the same storage account like "test" again without refreshing this window.
3. Check the result.

**Expect:**
Pop up an error told the users that the special storage account is non-existing.

**Actual:**
No error pops up when deleting a non-existing storage account.

wwlorey:
Warning message: 

Screen Shot 2020-01-14 at 3 43 27 PM

Fixes #568

### DIFF
--- a/src/tree/StorageAccountTreeItem.ts
+++ b/src/tree/StorageAccountTreeItem.ts
@@ -11,6 +11,7 @@ import * as azureStorage from "azure-storage";
 // tslint:disable-next-line:no-require-imports
 import opn = require('opn');
 import * as path from 'path';
+import * as vscode from 'vscode';
 import { commands, MessageItem, Uri, window } from 'vscode';
 import { AzureParentTreeItem, AzureTreeItem, createAzureClient, DialogResponses, IActionContext, ISubscriptionContext, UserCancelledError } from 'vscode-azureextensionui';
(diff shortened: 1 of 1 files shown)


src/tree/StorageAccountTreeItem.ts
```diff
@@ -151,7 +152,13 @@ export class StorageAccountTreeItem extends AzureParentTreeItem<IStorageRoot> {
             let storageManagementClient = createAzureClient(this.root, StorageManagementClient);
             let parsedId = this.parseAzureResourceId(this.storageAccount.id);
             let resourceGroupName = parsedId.resourceGroups;
+            const accountExistsBeforeDelete: boolean = await this.exists(storageManagementClient, resourceGroupName);
```
EricJizbaMSFT: I don't think this code is worth it per https://github.com/microsoft/vscode-azurestorage/issues/568#issuecomment-574436636

Issue #442, opened 2019-08-09
[blob] Mimic azure storage behavior when creating new folders in file explorer


monalgarg:
Azure Storage prevents users from creating folders that contain '/', '\\' or are empty strings. Users are allowed to create files with '/', this just creates folders and subfolders. However, Azure Storage treats '/' and '\\' similarly and will create folders and subfolders accordingly.

PrashanthCorp:
Verify: replacing '\\' with '/' solves the issue.

wwlorey:
VS Code will only create nested directories for paths that use the `/` separator.

For a path like `dir1\dir2\my_blob`, replacing `\` with `/` doesn't work because when VS Code calls `AzureStorageFS.createDirectory`, it expects to find a new directory with the URI it passed in

![Screen Shot 2020-01-15 at 4 57 30 PM](https://user-images.githubusercontent.com/22795803/72484136-26a3ca80-37b8-11ea-9a95-f64358f5a053.png)

PR #575, opened 2020-01-16
Create nested directories with '/' in file explorer

wwlorey:
This prevents nested directories for files or blobs from being created with the `\` separator. Backslashes are avoided because of [this problem](https://github.com/microsoft/vscode-azurestorage/issues/442#issuecomment-574930762) and because behavior is different between VS Code on Windows & Mac for `\` separators.

Fixes https://github.com/microsoft/vscode-azurestorage/issues/442

### DIFF
--- a/src/utils/validateNames.ts
+++ b/src/utils/validateNames.ts
@@ -7,7 +7,7 @@ const invalidBlobDirectoryChar = '\\'; // Prevents differing behavior when creat
 const invalidFileChars = ['"', '/', '\\', ':', '|', '<', '>', '?', '*'];
 const invalidFileCharsString = invalidFileChars.join(', ');
 
-export function validateBlobDirectoryName(name: string): string | undefined | null {
+export function validateBlobDirectoryName(name: string): string | undefined {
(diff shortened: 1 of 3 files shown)


src/utils/validateNames.ts
```diff
@@ -3,10 +3,19 @@
...
 
-const invalidChars = ['"', '/', '\\', ':', '|', '<', '>', '?', '*'];
-const invalidCharsString = invalidChars.join(', ');
+const invalidBlobDirectoryChar = '\\'; // Keeps behavior consistent on Mac & Windows when creating blob directories via the file explorer
+const invalidFileChars = ['"', '/', '\\', ':', '|', '<', '>', '?', '*'];
+const invalidFileCharsString = invalidFileChars.join(', ');
 
-export function validateDirectoryName(name: string): string | undefined | null {
+export function validateBlobDirectoryName(name: string): string | undefined {
+    if (!name || name.indexOf(invalidBlobDirectoryChar) !== -1) {
```
EricJizbaMSFT: nit: `name.includes(...)` instead of `name.indexOf(...) !== -1`

Issue #576, opened 2020-01-16
Trailing forward slashes in file/blob/directory names cause problems


wwlorey:
1. VS Code doesn't include the trailing `/` when calling `createDirectory`. This causes an error when VS Code then looks for the new directory and expects the trailing `/`
2. Creating a blob `my_blob/` is interpreted as a directory by Storage Explorer, not a blob. In VS Code, it's interpreted as a file with name `my_blob`. After refreshing in VS Code, it displays `my_blob/my_blob` - creating a phantom directory. Best to avoid this behavior

PR #579, opened 2020-01-17
Allow trailing slashes for directories, not for files/blobs

wwlorey:
Actually fixes https://github.com/microsoft/vscode-azurestorage/issues/576

### DIFF
--- a/src/AzureStorageFS.ts
+++ b/src/AzureStorageFS.ts
@@ -68,7 +68,7 @@ export class AzureStorageFS implements vscode.FileSystemProvider {
             if (uri.path.endsWith('/')) {
                 // Ignore trailing forward slashes
                 // https://github.com/microsoft/vscode-azurestorage/issues/576
-                uri = uri.with({ scheme: uri.scheme, path: uri.path.slice(0, -1), query: uri.query });
+                uri = uri.with({ path: uri.path.slice(0, -1) });
             }
 
             let treeItem: AzureStorageTreeItem = await this.lookup(uri, context);
--- a/src/AzureStorageFS.ts
+++ b/src/AzureStorageFS.ts
@@ -64,6 +64,13 @@ export class AzureStorageFS implements vscode.FileSystemProvider {
     async stat(uri: vscode.Uri): Promise<vscode.FileStat> {
         return await callWithTelemetryAndErrorHandling('stat', async (context) => {
             context.telemetry.suppressIfSuccessful = true;
+
+            if (uri.path.endsWith('/')) {
+                // Ignore trailing forward slashes
+                // https://github.com/microsoft/vscode-azurestorage/issues/576
+                uri = vscode.Uri.parse(`azurestorage://${uri.path.slice(0, -1)}?${uri.query}`);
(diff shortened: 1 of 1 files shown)


src/AzureStorageFS.ts
```diff
@@ -64,6 +64,13 @@ export class AzureStorageFS implements vscode.FileSystemProvider {
     async stat(uri: vscode.Uri): Promise<vscode.FileStat> {
         return await callWithTelemetryAndErrorHandling('stat', async (context) => {
             context.telemetry.suppressIfSuccessful = true;
+
+            if (uri.path.endsWith('/')) {
+                // Ignore trailing forward slashes
+                // https://github.com/microsoft/vscode-azurestorage/issues/576
+                uri = vscode.Uri.parse(`azurestorage://${uri.path.slice(0, -1)}?${uri.query}`);
```
EricJizbaMSFT: Take a look at `uri.with()` to simplify this code

src/AzureStorageFS.ts
```diff
@@ -188,6 +195,11 @@ export class AzureStorageFS implements vscode.FileSystemProvider {
                 throw getFileSystemError(uri, context, vscode.FileSystemError.NoPermissions);
             }
 
+            if (uri.path.endsWith('/')) {
+                // https://github.com/microsoft/vscode-azurestorage/issues/576
+                throw getFileSystemError(uri, context, () => { return new vscode.FileSystemError("File/blob names with a trailing '/' are not allowed."); });
```
EricJizbaMSFT: `getFileSystemError` will suppress errors from telemetry, which is useful for things like "File not found" because it's both common and expected. I feel like this error should be rare and might be worth tracking (right? I feel like we would reconsider this behavior if a bunch of people were getting it.)

wwlorey: Using `getFileSystemError` produces one error:
Screen Shot 2020-01-21 at 12 00 02 PM

Throwing an `Error` shows two messages, since VS Code attempts to read the newly created file but we don't actually create it:
Screen Shot 2020-01-21 at 11 57 47 AM

Maybe we could silently throw an error here just for telem and also throw using `getFileSystemError`? Or do you think having two errors is alright?

src/AzureStorageFS.ts
```diff
@@ -64,6 +64,13 @@ export class AzureStorageFS implements vscode.FileSystemProvider {
     async stat(uri: vscode.Uri): Promise<vscode.FileStat> {
         return await callWithTelemetryAndErrorHandling('stat', async (context) => {
             context.telemetry.suppressIfSuccessful = true;
+
+            if (uri.path.endsWith('/')) {
+                // Ignore trailing forward slashes
+                // https://github.com/microsoft/vscode-azurestorage/issues/576
+                uri = uri.with({ scheme: uri.scheme, path: uri.path.slice(0, -1), query: uri.query });
```
EricJizbaMSFT: I think the point of `with` is that you only specify the parts you're changing from the existing uri. You shouldn't need to specify scheme or query

PR #580, opened 2020-01-17
Add error for deploying from the file explorer

wwlorey:


src/commands/storageAccountActionHandlers.ts
```diff
@@ -56,7 +56,12 @@ async function deployStaticWebsite(context: IActionContext, target?: vscode.Uri
 
     // Disambiguate context this was executed from
     if (target instanceof vscode.Uri) {
-        // Command called from file view on a folder
+        // Command called from file view
+        if (target.scheme === 'azurestorage') {
+            context.telemetry.properties.contextValue = 'BlobContainer/FileShare';
+            throw new Error('Cannot deploy from an Azure resource via the file explorer.');
```
nturinski: VS Code calls that pane the "Explorer" so I'd use that language.  Also, I think "from" sounds weird.  Maybe something like:

`"Deploying an Azure resource from the Explorer is not supported."`

PR #588, opened 2020-01-28
Add 'Upload to Azure Storage...' context menu item for folders in the file explorer

wwlorey:
Related to https://github.com/microsoft/vscode-azurestorage/issues/159. We'll still have to add support for uploading individual files

I tried to make `TransferProgress` more portable to support progress for deploys as well as uploads. Refactoring it made the notification for deployments looks a bit different.

Old deploy notification:

old

New deploy notification: (the same notification shows deleting then uploading)

new

Here's what the new context menu item looks like: 

Screen Shot 2020-01-28 at 3 39 24 PM


src/commands/uploadToAzureStorage.ts
```diff
@@ -0,0 +1,50 @@
...
+export async function uploadToAzureStorage(actionContext: IActionContext, target?: vscode.Uri): Promise<void> {
+    let sourceFolderPath: string;
+    if (target) {
+        if (target.scheme === 'azurestorage') {
+            throw new Error('Cannot upload to Azure from an Azure resource.');
+        }
+
+        sourceFolderPath = target.fsPath;
+    } else {
+        const browseResult: vscode.Uri[] = await ext.ui.showOpenDialog({
```
nturinski: I assume that you got this logic from here:
https://github.com/microsoft/vscode-azurefunctions/blob/master/src/utils/workspace.ts#L20-L26

But the difference is that the defaultUri is used only if the user hits `Browse` from the quickPick list.  All folders will be displayed in the quickPick so automatically defaulting to `workspaceFolders[0]` is fine.  But in this case, it will always automatically open the dialog in the first folder, which if I was using multi-root, would be really annoying.

This function solves that issue https://github.com/microsoft/vscode-azurefunctions/blob/master/src/utils/workspace.ts#L54-L72

src/utils/TransferProgress.ts
```diff
@@ -0,0 +1,57 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.md in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as vscode from 'vscode';
+import { ext } from '../extensionVariables';
+
+export class TransferProgress {
```
nturinski: This would be a lot easier to review if you made the changes, get those approved and merged, and then change the file name.  😄 

I know the changes in here are very small, but because they're two totally files now, it's cumbersome to review.

src/utils/quickPickUtils.ts
```diff
@@ -68,3 +70,17 @@ export async function showWorkspaceFoldersQuickPick(placeHolderString: string, c
         return pickedItem.data;
     }
 }
+
+export async function showContainerShareQuickPick(context: ITreeItemPickerContext, placeHolder: string): Promise<BlobContainerTreeItem | FileShareTreeItem> {
+    let quickPicks: IAzureQuickPickItem<string>[] = [
```
nturinski: I know that this is legacy (at least, for this PR, you just copied/pasted it) but could we change this behavior?  It's an odd flow to pick Blob Container/File Share, the account, and then the blob container/file share after that.  

showTreeItemPicker accepts `string[]` so that you can pass both the `BlobContainerTreeItem.contextValue` and the `FileShareTreeItem.contextValue`.

You can change the `pickTreeItemImpl` of `StorageAccoutnTreeItem` to account for this situation, and have it return `this._blobContainerGroupTreeItem` and `this._fileShareGroupTreeItem`

wwlorey:
My bad - this is a pain to review as is. I made a new [PR](https://github.com/microsoft/vscode-azurestorage/pull/589) to start the process over and I'll incorporate feedback from this PR as I go

PR #589, opened 2020-01-30
Make TransferProgress more portable

wwlorey:
This lets us use `TransferProgress` for more than just single blob transfers

Related to #159

### DIFF
--- a/src/utils/blobUtils.ts
+++ b/src/utils/blobUtils.ts
@@ -127,31 +127,41 @@ export class TransferProgress {
     private lastUpdated: number = Date.now();
 
     constructor(
+        private readonly totalWork: number,
+        private readonly messagePrefix?: string,
         private readonly updateTimerMs: number = 200
     ) { }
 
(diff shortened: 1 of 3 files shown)


src/utils/blobUtils.ts
```diff
@@ -127,31 +127,41 @@ export class TransferProgress {
...
         this.postReport();
     }
 
-    private preReport(blobPath: string, loadedBytes: number, totalBytes: number): void {
-        this.percentage = Math.trunc((loadedBytes / totalBytes) * 100);
-        this.message = `${blobPath}: ${loadedBytes}/${totalBytes} (${this.percentage}%)`;
+    private preReport(finishedWork: number): void {
+        this.percentage = Math.trunc((finishedWork / this.totalWork) * 100);
+
+        let prefix: string = '';
```
nturinski: nit: You could just use a ternary statement to reduce this down to one line.

PR #590, opened 2020-01-30
Use virtual file system provider

wwlorey:
Fixes https://github.com/microsoft/vscode-azurestorage/issues/438

`checkCanUpload `, `checkCanDownload`, and the contents of `downloadFile` were moved from the deleted handler files into the tree item classes


nturinski:
Still pending on fixing that bug where saving the file opens the File Explorer view.

PR #595, opened 2020-02-03
Add 'Upload to Azure Storage...' command

wwlorey:
This adds support for uploading folders only

Related to https://github.com/microsoft/vscode-azurestorage/issues/159

Addresses comments from https://github.com/microsoft/vscode-azurestorage/pull/588

### DIFF
--- a/src/tree/StorageAccountTreeItem.ts
+++ b/src/tree/StorageAccountTreeItem.ts
@@ -12,23 +12,17 @@ import * as azureStorage from "azure-storage";
 import opn = require('opn');
 import * as path from 'path';
 import { commands, MessageItem, Uri, window } from 'vscode';
-import { AzureParentTreeItem, AzureTreeItem, createAzureClient, DialogResponses, IActionContext, IAzureQuickPickItem, ISubscriptionContext, UserCancelledError } from 'vscode-azureextensionui';
+import { AzureParentTreeItem, AzureTreeItem, createAzureClient, DialogResponses, IActionContext, ISubscriptionContext, UserCancelledError } from 'vscode-azureextensionui';
 import { getResourcesPath, staticWebsiteContainerName } from '../constants';
 import { ext } from "../extensionVariables";
-import { localize } from '../utils/localize';
 import { nonNullProp } from '../utils/nonNull';
 import { StorageAccountKeyWrapper, StorageAccountWrapper } from '../utils/storageWrappers';
 import { BlobContainerGroupTreeItem } from './blob/BlobContainerGroupTreeItem';
 import { BlobContainerTreeItem } from "./blob/BlobContainerTreeItem";
-import { DirectoryTreeItem } from './fileShare/DirectoryTreeItem';
 import { FileShareGroupTreeItem } from './fileShare/FileShareGroupTreeItem';
-import { FileShareTreeItem } from './fileShare/FileShareTreeItem';
-import { FileTreeItem } from './fileShare/FileTreeItem';
 import { IStorageRoot } from './IStorageRoot';
 import { QueueGroupTreeItem } from './queue/QueueGroupTreeItem';
-import { QueueTreeItem } from './queue/QueueTreeItem';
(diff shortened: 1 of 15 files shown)


src/tree/StorageAccountTreeItem.ts
```diff
@@ -102,30 +103,71 @@ export class StorageAccountTreeItem extends AzureParentTreeItem<IStorageRoot> {
...
+            case TableGroupTreeItem.contextValue:
+            case TableTreeItem.contextValue:
+                return this._tableGroupTreeItem;
+            default:
         }
 
         return undefined;
     }
 
+    private getLabel(contextValue: string): string {
```
nturinski: Rather than defining these all, you should be able to just use the label property on the tree items.

src/utils/uploadUtils.ts
```diff
@@ -0,0 +1,47 @@
...
+    for (const sourceFileIndex of filePathsToUpload.keys()) {
+        throwIfCanceled(cancellationToken, properties, "uploadFiles");
+        let sourceFilePath: string = filePathsToUpload[sourceFileIndex];
+        let relativeFile: string = path.relative(sourceFolder, sourceFilePath);
+        let destFilePath: string = path.join(destFolder, relativeFile);
+        ext.outputChannel.appendLine(`Uploading ${sourceFilePath}...`);
+
+        try {
+            if (destTreeItem instanceof BlobContainerTreeItem) {
+                await destTreeItem.uploadFileToBlockBlob(sourceFilePath, destFilePath, true /* suppressLogs */);
```
nturinski: These seem to take the exact same parameters.  It seems like you could just name them both `uploadLocalFile` or just `uploadFile` and not even check the `destTreeItem` type

src/utils/uploadUtils.ts
```diff
@@ -0,0 +1,47 @@
...
+import * as path from 'path';
+import * as vscode from 'vscode';
+import { parseError, TelemetryProperties } from "vscode-azureextensionui";
+import { ext } from '../extensionVariables';
+import { TransferProgress } from '../TransferProgress';
+import { BlobContainerTreeItem } from '../tree/blob/BlobContainerTreeItem';
+import { FileShareTreeItem } from '../tree/fileShare/FileShareTreeItem';
+import { throwIfCanceled } from './errorUtils';
+
+export async function uploadFiles(
```
nturinski: Since this is only being used in one file right now, it doesn't need to have its own util file.  I think you could just define this in the `uploadToAzure.ts` file as a separate function

wwlorey:
Here's what that problem prompt looks like now:

Screen Shot 2020-02-05 at 4 05 46 PM

I thought 'Select resource type' seemed better than 'Select resource'

src/tree/StorageAccountTreeItem.ts
```diff
@@ -103,71 +99,6 @@ export class StorageAccountTreeItem extends AzureParentTreeItem<IStorageRoot> {
...
-        }
-
-        switch (contextValue) {
-            case BlobContainerGroupTreeItem.contextValue:
-            case BlobContainerTreeItem.contextValue:
-                return this._blobContainerGroupTreeItem;
-            case FileShareGroupTreeItem.contextValue:
-            case FileShareTreeItem.contextValue:
-            case DirectoryTreeItem.contextValue:
-            case FileTreeItem.contextValue:
```
nturinski: I think you need to add `FileTreeItem` and `DirectoryTreeItem` to `isAncestorOfImpl`.

src/tree/blob/BlobContainerGroupTreeItem.ts
```diff
@@ -70,6 +70,14 @@ export class BlobContainerGroupTreeItem extends AzureParentTreeItem<IStorageRoot
         throw new UserCancelledError();
     }
 
+    public isAncestorOfImpl(contextValue: string): boolean {
```
nturinski: Could just make this
`return contextValue === BlobContainerTreeItem.context`

Issue #601, opened 2020-02-05
Error when creating nested blobs in the Azure view


wwlorey:
I broke this when I made the [virtual file system changes](https://github.com/microsoft/vscode-azurestorage/commit/12a4411054b675becef71119073bf5728ffa8b3a).

![Screen Shot 2020-02-05 at 10 35 29 AM](https://user-images.githubusercontent.com/22795803/73872136-f7570b00-4803-11ea-8f3a-f4d8a1ba5071.png)

The new blob can be opened after refreshing the Azure view.

Issue #574, opened 2020-01-15
Show the file tree hierarchy when creating a nested blob via the Azure view


wwlorey:
When creating a nested blob such as `b/c/d/e`

Instead of displaying the blob's full path like this

![](https://user-images.githubusercontent.com/22795803/72463725-c2b5dd80-3788-11ea-8620-b3023b3900f8.png)

Display the hierarchical view without having to refresh

![](https://user-images.githubusercontent.com/22795803/72463807-f2fd7c00-3788-11ea-9d5c-00e90b0b7777.png)

Related: https://github.com/microsoft/vscode-azurestorage/pull/572

wwlorey:
Maybe I should have made this a "possible won't fix" 😂

PR #607, opened 2020-02-10
Create blobs & directories manually in 'createBlockBlob'

wwlorey:
Fixes https://github.com/microsoft/vscode-azurestorage/issues/601
Fixes https://github.com/microsoft/vscode-azurestorage/issues/574

### DIFF
--- a/src/commands/blob/blobContainerActionHandlers.ts
+++ b/src/commands/blob/blobContainerActionHandlers.ts
@@ -3,14 +3,16 @@
  *  Licensed under the MIT License. See License.md in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
+import * as path from 'path';
 import * as vscode from 'vscode';
 import { IActionContext, registerCommand } from 'vscode-azureextensionui';
 import { AzureStorageFS } from '../../AzureStorageFS';
-import { ext } from '../../extensionVariables';
 import { storageExplorerLauncher } from '../../storageExplorerLauncher/storageExplorerLauncher';
 import { BlobContainerTreeItem } from '../../tree/blob/BlobContainerTreeItem';
+import { BlobDirectoryTreeItem } from '../../tree/blob/BlobDirectoryTreeItem';
 import { BlobTreeItem } from '../../tree/blob/BlobTreeItem';
-import { showBlobPathInputBox } from '../../utils/blobUtils';
+import { createBlobDirectory, findBlobTreeItem, getBlobResourceType, IBlobContainerCreateChildContext, showBlobPathInputBox } from '../../utils/blobUtils';
+import { localize } from '../../utils/localize';
 import { deleteNode } from '../commonTreeCommands';
 
 export function registerBlobContainerActionHandlers(): void {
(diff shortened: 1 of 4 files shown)


src/utils/blobUtils.ts
```diff
@@ -56,26 +56,14 @@ export async function loadMoreBlobChildren(parent: BlobContainerTreeItem | BlobD
...
-    });
+    let blobPath: string = context.childName || await showBlobPathInputBox(parent);
 
     return await vscode.window.withProgress({ location: vscode.ProgressLocation.Window }, async (progress) => {
         context.showCreatingTreeItem(blobPath);
         progress.report({ message: `Azure Storage: Creating block blob '${blobPath}'` });
         await createOrUpdateBlockBlob(parent, blobPath);
         // Use the full blob path as the label since this is a newly created blob. https://github.com/microsoft/vscode-azurestorage/issues/565
-        return new BlobTreeItem(parent, blobPath, parent.container, blobPath);
+        return new BlobTreeItem(parent, blobPath, parent.container);
```
EricJizbaMSFT: Remove the comment above since it no longer applies. Also just get rid of `label` as a parameter to the constructor in `BlobTreeItem` since I don't think it's ever used

src/commands/blob/blobContainerActionHandlers.ts
```diff
@@ -6,17 +6,46 @@
 import * as vscode from 'vscode';
```
EricJizbaMSFT: Is this related to 574 or does it fix 574? It seems like it fixes it, in which case please remove the "wont-fix" label and put it in the current milestone

src/commands/blob/blobContainerActionHandlers.ts
```diff
@@ -6,17 +6,46 @@
...
+
+        for (const dirName of dirNames) {
+            currentPath += `${dirName}/`;
+            const dirUri: vscode.Uri = AzureStorageFS.idToUri(parent.fullId, currentPath);
+
+            try {
+                let fileType: vscode.FileType = (await ext.azureStorageFS.stat(dirUri)).type;
+                if (fileType !== vscode.FileType.Directory) {
+                    // Remove trailing slash from currentPath for error message
+                    throw new Error(`${currentPath.slice(0, -1)} is not a directory`);
```
EricJizbaMSFT: nit: Please get in the habit of using `localize` for new strings

src/commands/blob/blobContainerActionHandlers.ts
```diff
@@ -6,17 +6,46 @@
...
+                }
```
EricJizbaMSFT: Any time you have a dynamically inserted value in a string, wrap it in quotes

`This is not a directory` vs. `"This" is not a directory`

src/commands/blob/blobContainerActionHandlers.ts
```diff
@@ -6,17 +6,46 @@
...
 
 export function registerBlobContainerActionHandlers(): void {
     registerCommand("azureStorage.openBlobContainer", openBlobContainerInStorageExplorer);
     registerCommand("azureStorage.editBlob", async (_context: IActionContext, treeItem: BlobTreeItem) => AzureStorageFS.showEditor(treeItem), 250);
     registerCommand("azureStorage.deleteBlobContainer", async (context: IActionContext, treeItem?: BlobContainerTreeItem) => await deleteNode(context, BlobContainerTreeItem.contextValue, treeItem));
-    registerCommand("azureStorage.createBlockBlob", async (context: IActionContext, treeItem: BlobContainerTreeItem) => {
-        let childTreeItem = await treeItem.createChild(context);
+    registerCommand("azureStorage.createBlockBlob", async (context: IActionContext, parent: BlobContainerTreeItem) => {
+        const blobPath: string = await showBlobPathInputBox(parent);
+        const dirNames: string[] = blobPath.split('/').slice(0, -1);
```
EricJizbaMSFT: Are there `path` methods we could use instead? I get nervous about off-by-one errors, etc. whenever I see hard-coded zeros, ones, and slashes

src/commands/blob/blobContainerActionHandlers.ts
wwlorey: I thought to fix 574 it would have to expand the new blob when its created. Right now it just displays the root directory and you have to expand it yourself

EricJizbaMSFT: Oh I thought the issue was more about getting the correct tree nodes and didn't even think about the expansion state

src/utils/blobUtils.ts
```diff
@@ -119,6 +108,77 @@ export async function getExistingProperties(parent: BlobTreeItem | BlobContainer
...
+                return localize('aBlobWithThisPathAndNameAlreadyExists', 'A blob with this path and name already exists');
+            }
+
+            return undefined;
+        }
+    });
+}
+
+export async function findBlobTreeItem(uri: vscode.Uri, filePath: string, containerTreeItem: BlobContainerTreeItem, context: IActionContext, endSearchEarly?: boolean): Promise<AzureStorageBlobTreeItem> {
+    let treeItem: BlobTreeItem | BlobDirectoryTreeItem | BlobContainerTreeItem = containerTreeItem;
```
EricJizbaMSFT: I think `ext.tree.findTreeItem` would suffice for your scenario instead of pulling this code out into `findBlobTreeItem`

src/utils/blobUtils.ts
```diff
@@ -119,6 +108,77 @@ export async function getExistingProperties(parent: BlobTreeItem | BlobContainer
...
+
+    return treeItem;
+}
+
+export async function getBlobResourceType(filePath: string, containerTreeItem: BlobContainerTreeItem, context: IActionContext): Promise<vscode.FileType> {
+    let treeItem: AzureStorageBlobTreeItem = await findBlobTreeItem(vscode.Uri.parse(containerTreeItem.fullId), filePath, containerTreeItem, context);
+    return treeItem instanceof BlobDirectoryTreeItem || treeItem instanceof BlobContainerTreeItem ? vscode.FileType.Directory : vscode.FileType.File;
+}
+
+export async function createBlobDirectory(fullDirPath: string, parentPath: string, parentTreeItem: BlobDirectoryTreeItem | BlobContainerTreeItem, context: IActionContext): Promise<void> {
```
EricJizbaMSFT: Most of the code you pulled out into this function doesn't apply to your scenario. You already know the exact dirName and you're only creating one directory at a time. Just call createChild directly. Something like this:
```
await dirParentTreeItem.createChild(<IBlobContainerCreateChildContext>{ ...context, childType: BlobDirectoryTreeItem.contextValue, childName: dirName });
```

Issue #596, opened 2020-02-04
Duplicated errors pop up when renaming a blob in File Explorer


v-dantch:
**OS**: Windows
**Build Version**: [20200203.1](https://dev.azure.com/ms-azuretools/AzCode/_build/results?buildId=14328&view=artifacts&type=publishedArtifacts)

**Repro Steps**:
1. Create a storage account --> Create a blob container --> Open this blob container in File Explorer.
2. New a file --> Rename this file.
3. Pop up an error  --> Check it.

**Expect:**
Pop up one error "Renaming folders or files not supported".

**Actual:**
Duplicated errors pop up.
![image](https://user-images.githubusercontent.com/34729092/73723106-18710c00-46dd-11ea-9ef5-67325f8073e1.png)

**More Info:**
1. This issue also reproduces when renaming a folder in File Explorer.
2. This issue doesn't reproduce on VS Code(not insiders).

PR #609, opened 2020-02-11
Fix multiple rename errors

wwlorey:
There were multiple messages on insiders only. I assume the change on VS Code's side is intended & will makes its way to stable soon

Fixes https://github.com/microsoft/vscode-azurestorage/issues/596

### DIFF
--- a/src/AzureStorageFS.ts
+++ b/src/AzureStorageFS.ts
@@ -303,11 +303,10 @@ export class AzureStorageFS implements vscode.FileSystemProvider, vscode.TextDoc
             let newUriParsed = this.parseUri(newUri);
 
             context.errorHandling.rethrow = true;
-            context.errorHandling.suppressDisplay = true;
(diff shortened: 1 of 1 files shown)


src/AzureStorageFS.ts
```diff
@@ -303,12 +303,10 @@ export class AzureStorageFS implements vscode.FileSystemProvider, vscode.TextDoc
             let newUriParsed = this.parseUri(newUri);
```
EricJizbaMSFT: > I assume the change on VS Code's side is intended & will makes its way to stable soon

I would rather confirm this than just make an assumption. Also - how is this handled for users before they update to the new version of VS Code? Duplicate errors are annoying, but better than no errors IMO

wwlorey: As implemented here, it doesn't throw errors in the current version of VS Code :/ 

Throwing the error using `getFileSystemError` only produces one error on insiders & the current version

Issue #247, opened 2018-11-09
Azure Storage Local Emulator


junalmeida:
Feature Request: Provide a way to access local azure storage emulator.

Cpcrook:
I, too, would appreciate this functionality.  Seems like it _should be_ a feature of this, as the Storage Explorer standalone picks up the local emulator account already.

JustinGrote:
Just add a vscode command that lets specifying the connection string. Preferably it would support '
'UseDevelopmentStorage=true'

cnkcb:
I also prefer 'UseDevelopmentStorage=true', so I can get it to work with the Azurite Extension and just do the whole thing from within VS Code.

wwlorey:
This feature was included in the 0.9.0 release. Thanks for suggesting!

junalmeida:
@wwlorey This is awesome, thank you for taking it in consideration :)

PR #618, opened 2020-02-27
Add support for local storage emulator & attaching via connection strings

### LINKED ISSUE #630: Always show all services for attached storage accounts
Not doing this leads to issues like this:

If the connection is slow, polling for blob containers, file shares, etc. sometimes times out. So sometimes only a subset of the supported nodes are displayed beneath the storage account tree item.

For example: `wwla` supports all 4 resource types but only file shares was loaded.

Screen Shot 2020-03-06 at 4 32 34 PM

wwlorey:
Fixes https://github.com/microsoft/vscode-azurestorage/issues/247
Fixes #630


src/commands/startEmulator.ts
```diff
@@ -0,0 +1,49 @@
...
+        if (selected === 'Install Azurite') {
+            context.telemetry.properties.installAzuriteExtension = 'true';
+            await vscode.commands.executeCommand('extension.open', 'Azurite.azurite');
+        } else {
+            throw new UserCancelledError();
+        }
+    }
+
+    // Wait for emulator to start before refreshing tree view
+    setTimeout(async () => { await ext.tree.refresh(ext.attachedStorageAccountsTreeItem); }, 2000);
```
wwlorey: I wasn't sure if there was a better way to let the emulator attach before refreshing

src/tree/blob/BlobContainerGroupTreeItem.ts
```diff
@@ -7,30 +7,52 @@ import * as azureStorageBlob from "@azure/storage-blob";
...
         light: path.join(getResourcesPath(), 'light', 'AzureBlobContainer.svg'),
         dark: path.join(getResourcesPath(), 'dark', 'AzureBlobContainer.svg')
     };
 
+    public get label(): string {
+        return `Blob Containers${this.active ? '' : ' (stopped)'}`;
+    }
+
+    public get contextValue(): string {
+        return `${BlobContainerGroupTreeItem.contextValue}${this.active ? '' : 'Stopped'}`;
```
nturinski: Should the context value change based on whether or not it's stopped?  I'm not sure if we have any other extensions that employ that behavior.

src/tree/blob/BlobContainerGroupTreeItem.ts
```diff
@@ -7,30 +7,52 @@ import * as azureStorageBlob from "@azure/storage-blob";
...
-    public label: string = "Blob Containers";
     public readonly childTypeLabel: string = "Blob Container";
     public static contextValue: string = 'azureBlobContainerGroup';
-    public contextValue: string = BlobContainerGroupTreeItem.contextValue;
     public iconPath: { light: string | Uri; dark: string | Uri } = {
         light: path.join(getResourcesPath(), 'light', 'AzureBlobContainer.svg'),
         dark: path.join(getResourcesPath(), 'dark', 'AzureBlobContainer.svg')
     };
 
+    public get label(): string {
```
nturinski: Rather than appending this to the label, use the `description` property instead.

src/tree/queue/QueueGroupTreeItem.ts
```diff
@@ -6,29 +6,51 @@
...
     public readonly childTypeLabel: string = "Queue";
     public static contextValue: string = 'azureQueueGroup';
-    public contextValue: string = QueueGroupTreeItem.contextValue;
     public iconPath: { light: string | Uri; dark: string | Uri } = {
         light: path.join(getResourcesPath(), 'light', 'AzureQueue.svg'),
         dark: path.join(getResourcesPath(), 'dark', 'AzureQueue.svg')
     };
 
-    async loadMoreChildrenImpl(clearCache: boolean): Promise<QueueTreeItem[]> {
+    public get label(): string {
```
nturinski: Same here

src/AzureStorageFS.ts
```diff
@@ -37,12 +37,12 @@ export class AzureStorageFS implements vscode.FileSystemProvider, vscode.TextDoc
     readonly onDidChangeFile: vscode.Event<vscode.FileChangeEvent[]> = this._emitter.event;
 
     static idToUri(resourceId: string, filePath?: string): vscode.Uri {
-        let matches: RegExpMatchArray | null = resourceId.match(/(\/subscriptions\/[^\/]+\/resourceGroups\/[^\/]+\/providers\/Microsoft.Storage\/storageAccounts\/[^\/]+\/[^\/]+\/[^\/]+)\/?(.*)/i);
+        let matches: RegExpMatchArray | null = resourceId.match(/((\/.*)?\/subscriptions\/[^\/]+\/resourceGroups\/[^\/]+\/providers\/Microsoft.Storage\/storageAccounts\/[^\/]+\/[^\/]+\/[^\/]+)\/?(.*)/i);
```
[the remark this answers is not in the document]
wwlorey: The `AttachedStorageAccountsTreeItem` ID was being prepended to the resource ID so I changed this function to handle it. Do you know of a way to prevent that from happening without getting rid of the tree item's ID?

package.json
```diff
@@ -673,6 +730,18 @@
 				{
 					"command": "azureStorage.openQueue",
 					"when": "never"
+				},
+				{
+					"command": "azureStorage.attachStorageAccount",
```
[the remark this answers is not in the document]
wwlorey: `attachStorageAccount` prompts you to choose between `attachEmulator` and `attachWithConnectionString` so I included the latter two in the command palette

src/tree/blob/BlobContainerGroupTreeItem.ts
wwlorey: I did that so context menu options wouldn't appear for stopped containers

src/commands/selectStorageAccountNodeForCommand.ts
```diff
@@ -26,6 +28,10 @@ export async function selectStorageAccountTreeItemForCommand(
     //   a storage account treeItem
     //   a blob container treeItem
 
+    if (treeItem?.parent?.parent instanceof AttachedStorageAccountTreeItem) {
```
wwlorey: It felt like it was out of the scope of this PR to add support for static sites after creating the new `AttachedStorageAccountTreeItem`

src/commands/startEmulator.ts
```diff
@@ -0,0 +1,81 @@
...
+        // This is a long running command, don't await it
+        // tslint:disable-next-line: no-floating-promises
+        cpUtils.executeCommand(ext.outputChannel, undefined, `azurite-${emulatorType}`, childProcWrapper);
+        const startTime: number = Date.now();
+        ext.outputChannel.show();
+
+        if (childProcWrapper.childProc) {
+            childProcWrapper.childProc.addListener('close', () => {
+                // This process shouldn't end prematurely unless an error ocurred
+                emulatorError = new Error(localize('failedToStartEmulatorUsingAzuriteCLI', 'Failed to start emulator using the Azurite CLI. Check the output window for more details.'));
```
nturinski: Is `emulatorError` supposed to be set to another value at some point?

I would assume that you would be listening for it in the childProcess or something.

src/commands/startEmulator.ts
```diff
@@ -0,0 +1,81 @@
...
+        }
+    } else {
+        const installAzurite: vscode.MessageItem = { title: localize('installAzurite', 'Install Azurite') };
+        const message: string = localize('mustInstallAzurite', 'You must install Azurite to use the Storage Emulator.');
+
+        const result: vscode.MessageItem = await ext.ui.showWarningMessage(message, { modal: true }, installAzurite);
+        if (result === installAzurite) {
+            context.telemetry.properties.installAzuriteExtension = 'true';
+            await vscode.commands.executeCommand('extension.open', 'Azurite.azurite');
+        } else {
```
nturinski: This is unnecessary now since `ext.ui` handles the canceled case.

src/commands/startEmulator.ts
```diff
@@ -0,0 +1,81 @@
...
+        return false;
+    }
+}
+
+export enum EmulatorType {
+    blob = 'blob',
+    queue = 'queue'
+}
+
+export interface IChildProcWrapper {
```
nturinski: Feels like this should live in `cpUtils.ts`

src/tree/AttachedStorageAccountTreeItem.ts
```diff
@@ -0,0 +1,172 @@
...
+export class AttachedStorageAccountTreeItem extends AzExtParentTreeItem {
+    public iconPath: { light: string | Uri; dark: string | Uri } = {
+        light: path.join(getResourcesPath(), 'light', 'AzureStorageAccount.svg'),
+        dark: path.join(getResourcesPath(), 'dark', 'AzureStorageAccount.svg')
+    };
+    public childTypeLabel: string = 'resource type';
+    public autoSelectInTreeItemPicker: boolean = true;
+    public id: string = this.storageAccount.id;
+    public label: string = this.storageAccount.name;
+    public static contextValue: string = `${StorageAccountTreeItem.contextValue}${attachedAccountSuffix}`;
```
nturinski: I'd put some sort of separator, like a hyphen or something.  Could make the regexp grabbing both TreeItems easier.

Also you're only using `attachedAccountSuffix` once.  Making a `const` is kind of overkill.

src/tree/AttachedStorageAccountTreeItem.ts
```diff
@@ -0,0 +1,172 @@
...
+        } else {
+            if (blobContainersActive) {
+                groupTreeItems.push(this._blobContainerGroupTreeItem);
+            }
+
+            if (queuesActive) {
+                groupTreeItems.push(this._queueGroupTreeItem);
+            }
+
+            const shareClient: azureStorageShare.ShareServiceClient = this.root.createShareServiceClient();
```
nturinski: For clarity, I would make two more functions

`isTableActive` and `isFileShareActive` and put all the logic in there.  Then do the `.push` in `LoadMoreChildren`.  I think it's easier to decipher what's going on.

src/tree/AttachedStorageAccountTreeItem.ts
```diff
@@ -0,0 +1,172 @@
...
+            if (await this.taskResolvesBeforeTimeout(shareClient.getProperties())) {
+                groupTreeItems.push(this._fileShareGroupTreeItem);
+            }
+
+            const tableService: azureStorage.TableService = this.root.createTableService();
+            const tableTask: Promise<void> = new Promise((resolve, reject) => {
+                // Getting table service properties will succeed even when tables aren't supported, so attempt to list tables instead
+                // tslint:disable-next-line:no-any
+                tableService.listTablesSegmented(<azureStorage.TableService.ListTablesContinuationToken><unknown>undefined, (err?: any) => {
```
nturinski: Do you know how fast this API call timeouts? The around this seems kind of complex.  If we can just do a try/catch and await it, that'd be preferable.

src/tree/AttachedStorageAccountTreeItem.ts
```diff
@@ -0,0 +1,172 @@
...
+
+    public hasMoreChildrenImpl(): boolean {
+        return false;
+    }
+
+    public async getConnectionString(): Promise<string> {
+        return this.connectionString;
+    }
+
+    public async getActualWebsiteHostingStatus(): Promise<WebsiteHostingStatus> {
```
nturinski: Do you need to have this because of `blobContainerNode`?  Otherwise this never gets used, correct?

src/tree/AttachedStorageAccountTreeItem.ts
```diff
@@ -0,0 +1,172 @@
...
+                return new azureStorage.QueueService(this.connectionString).withFilter(new azureStorage.ExponentialRetryPolicyFilter());
+            },
+            createTableService: () => {
+                return new azureStorage.TableService(this.connectionString).withFilter(new azureStorage.ExponentialRetryPolicyFilter());
+            }
+        };
+    }
+
+    // tslint:disable-next-line:no-any
+    private async taskResolvesBeforeTimeout(promise: any): Promise<boolean> {
```
nturinski: Out of curiosity... did you find this somewhere?

src/tree/AttachedStorageAccountsTreeItem.ts
```diff
@@ -0,0 +1,222 @@
...
+            }
+        }
+    }
+
+    private async getAttachedAccounts(): Promise<AttachedStorageAccountTreeItem[]> {
+        if (!this._attachedAccounts) {
+            try {
+                this._attachedAccounts = await this._loadPersistedAccountsTask;
+            } catch {
+                this._attachedAccounts = [];
```
nturinski: How much work would it be to only detach persistedAccounts that failed rather than all of them?  I think it'd be kind of annoying if I had 4 attached accounts, and only one failed but all of them got deleted.

src/tree/AttachedStorageAccountsTreeItem.ts
```diff
@@ -0,0 +1,222 @@
...
+
+                persistedAccounts.push(await this.createTreeItem(connectionString, account.name));
+            }));
+        }
+
+        return persistedAccounts;
+    }
+
+    // tslint:disable-next-line:no-reserved-keywords
+    private async createTreeItem(connectionString: string, name: string): Promise<AttachedStorageAccountTreeItem> {
```
nturinski: I think you can get rid of this tslint disabler

src/tree/queue/QueueGroupTreeItem.ts
```diff
@@ -18,17 +20,39 @@ export class QueueGroupTreeItem extends AzureParentTreeItem<IStorageRoot> {
...
     public static contextValue: string = 'azureQueueGroup';
-    public contextValue: string = QueueGroupTreeItem.contextValue;
     public iconPath: { light: string | Uri; dark: string | Uri } = {
         light: path.join(getResourcesPath(), 'light', 'AzureQueue.svg'),
         dark: path.join(getResourcesPath(), 'dark', 'AzureQueue.svg')
     };
 
-    async loadMoreChildrenImpl(clearCache: boolean): Promise<QueueTreeItem[]> {
+    public get description(): string {
+        return this.active ? '' : 'stopped';
```
nturinski: nit: No real preference, but you can make 'Stopped' a constant in this case.

src/AzureStorageFS.ts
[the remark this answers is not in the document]
wwlorey: Yeah I originally tried changing the `AttachedStorageAccountsTreeItem` id to work with this but the label gets used if id isn't provided or is an empty string. Is it alright keeping these FS changes?

src/commands/startEmulator.ts
wwlorey: So I only use `emulatorError` to see if the process ended early then check it below. There wasn't a good way to see if the process was running so I check if it isn't running instead

src/tree/AttachedStorageAccountTreeItem.ts
wwlorey: Yeah this is pretty gross. And it doesn't help that some of the parameters are mistyped in the tables SDK.

The call still hadn't resolved after 30 sec when tables weren't supported

src/tree/AttachedStorageAccountTreeItem.ts
wwlorey: Yeah I kept this in to avoid errors when loading the $web container

src/tree/AttachedStorageAccountTreeItem.ts
wwlorey: Yeah it's from here: https://italonascimento.github.io/applying-a-timeout-to-your-promises/

In the 'Implementing the timeout' section

src/tree/AttachedStorageAccountsTreeItem.ts
wwlorey: Hmm we could do it, there's just several spots where an error could happen in `loadPersistedAccounts` so we'd have to account for that. Is this something you think we should do for this release?

src/AzureStorageFS.ts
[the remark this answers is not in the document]
nturinski: I'm worried about the impact of you changing the RegExp in AzureStorageFS since it touches all the other treeItems too.

Couldn't you construct an id that's a similar structure to an AzureAccountTreeItem so that this wouldn't be an issue?

src/tree/AttachedStorageAccountsTreeItem.ts
nturinski: @EricJizbaMSFT I'd like to hear your thoughts.  I personally think it's something we should make an issue for and handle next release.

src/AzureStorageFS.ts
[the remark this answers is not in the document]
wwlorey: I tried extending `AzureAccountTreeItemBase` but adding the label class member still prepended it to the id. Is there an alternate way to label the tree item?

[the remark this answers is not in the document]
EricJizbaMSFT: > I'm worried about the impact of you changing the RegExp in AzureStorageFS since it touches all the other treeItems too.

I can jive with this

> Couldn't you construct an id that's a similar structure to an AzureAccountTreeItem so that this wouldn't be an issue?

but would not suggest doing this. It feels hacky to pretend attached tree items are azure tree items - they're not. I'd keep the attached ids as simple as possible and mitigate risk of the above comment by splitting out the logic here - something like this:
```
let idRegExp: RegExp;
if (resourceId.startsWith('/attachedStorageAccounts')) {
    idRegExp = // new regexp based on attached id structure
} else {
    idRegExp = /(\/subscriptions\/[^\/]+\/resourceGroups\/[^\/]+\/providers\/Microsoft.Storage\/storageAccounts\/[^\/]+\/[^\/]+\/[^\/]+)\/?(.*)/i;
}

let matches: RegExpMatchArray | null = resourceId.match(idRegExp);
...
```

src/tree/AttachedStorageAccountsTreeItem.ts
```diff
@@ -0,0 +1,221 @@
...
+            return <IPersistedAccount>{
+                fullId: treeItem.fullId,
+                name: treeItem.storageAccount.name,
+            };
+        });
+        await ext.context.globalState.update(this._serviceName, JSON.stringify(value));
+    }
+
+    private getAttachedAccountId(name: string): string {
+        return `/subscriptions/attached/resourceGroups/attached/providers/Microsoft.Storage/storageAccounts/${name}`;
```
EricJizbaMSFT: Can you remove all the hard-coded junk in this id? As mentioned elsewhere, it doesn't feel relevant for attached accounts and I think it's better to keep it as simple as possible.

src/tree/AttachedStorageAccountsTreeItem.ts
EricJizbaMSFT: I'm fine with the current behavior as-is. `loadPersistedAccounts` is unlikely to fail and this behavior has existed in Cosmos for a long time with no complaints

src/tree/IStorageRoot.ts
```diff
@@ -17,3 +17,12 @@ export interface IStorageRoot extends ISubscriptionContext {
     createQueueService(): azureStorage.QueueService;
     createTableService(): azureStorage.TableService;
 }
+
```
EricJizbaMSFT: IStorageRoot can be simplified so you don't have duplicate code. I think this'll work:
```
export interface IStorageRoot extends ISubscriptionContext, IAttachedStorageRoot { }
```

src/utils/keytar.ts
```diff
@@ -0,0 +1,14 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+// tslint:disable-next-line: no-implicit-dependencies
```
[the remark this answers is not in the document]
EricJizbaMSFT: I get this error in VS Code in the storage repo with your changes, but not the cosmos repo. Do you see that, too? Any ideas?

```
[12:02:40 PM] File change detected. Starting incremental compilation...

src/utils/keytar.ts:7:29 - error TS2307: Cannot find module 'keytar'.

7 import * as keytarType from 'keytar';
                              ~~~~~~~~

[12:02:40 PM] Found 1 error. Watching for file changes.
```

package.json
```diff
@@ -297,6 +303,40 @@
...
+				"command": "azureStorage.attachStorageAccount",
+				"title": "Attach Storage Account...",
+				"category": "Azure Storage",
+				"icon": {
+					"light": "resources/light/Connect.svg",
+					"dark": "resources/dark/Connect.svg"
+				}
+			},
+			{
+				"command": "azureStorage.attachEmulator",
```
EricJizbaMSFT: I feel like it would simplify this PR quite a bit if you just always displayed the emulator in the tree, (aka no need to attach/detach/persist). That's what Storage Explorer does

nturinski: Yeah, I agree with this especially since you provide ways to install/start the service.

src/commands/selectStorageAccountNodeForCommand.ts
nturinski: I think that @fiveisprime mentioned that we are going to deprioritize static website deployments from the Storage Extension from now on.

src/commands/startEmulator.ts
```diff
@@ -0,0 +1,74 @@
...
+import { delay } from '../utils/delay';
+import { localize } from '../utils/localize';
+
+const emulatorTimeoutInSeconds: number = 3;
+export const emulatorTimeoutInMs: number = emulatorTimeoutInSeconds * 1000;
+
+export async function startEmulator(context: IActionContext, emulatorType: EmulatorType): Promise<void> {
+    if (!!vscode.extensions.getExtension('Azurite.azurite')) {
+        // Use the Azurite extension
+        await vscode.commands.executeCommand(`azurite.start_${emulatorType}`);
```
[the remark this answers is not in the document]
nturinski: I personally think being able to start it from the extension is a pretty nice convenience.

src/commands/startEmulator.ts
```diff
@@ -0,0 +1,74 @@
...
+import { ext } from "../extensionVariables";
+import { cpUtils } from '../utils/cpUtils';
+import { delay } from '../utils/delay';
+import { localize } from '../utils/localize';
+
+const emulatorTimeoutInSeconds: number = 3;
+export const emulatorTimeoutInMs: number = emulatorTimeoutInSeconds * 1000;
+
+export async function startEmulator(context: IActionContext, emulatorType: EmulatorType): Promise<void> {
+    if (!!vscode.extensions.getExtension('Azurite.azurite')) {
```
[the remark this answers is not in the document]
wwlorey: Don't we just need to check if the extension is installed before we use it to start the emulator? If Azurite isn't installed `getExtension` returns undefined

src/utils/keytar.ts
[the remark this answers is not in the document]
wwlorey: Weird, I don't get that. It still threw that after an `npm install`?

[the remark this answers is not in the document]
EricJizbaMSFT: Lol yep that fixed it thanks

src/commands/selectStorageAccountNodeForCommand.ts
EricJizbaMSFT: We can discuss during standup or triage. In any case, doesn't affect this PR

src/commands/startEmulator.ts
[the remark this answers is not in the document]
EricJizbaMSFT: Yeah nevermind. As long as the Azurite extension has the command listed as an activation event it should be fine

I was thinking of the more "official" method of interacting between extensions using `exports` like [here](https://github.com/microsoft/vscode-azureappservice/blob/f6591d861d0aeb2471a2bac1ee3898b33dc051d2/src/explorer/CosmosDBTreeItem.ts#L132)

src/tree/AttachedStorageAccountTreeItem.ts
```diff
@@ -0,0 +1,187 @@
...
+    public get root(): IAttachedStorageRoot {
+        return this._root;
+    }
+
+    public async loadMoreChildrenImpl(): Promise<AzureTreeItem<IStorageRoot>[]> {
+        const blobContainersActive: boolean = await this.isBlobContainerActive();
+        const queuesActive: boolean = await this.isQueueActive();
+        let groupTreeItems: AzureTreeItem<IStorageRoot>[] = [];
+
+        if (this.connectionString === AttachedStorageAccountsTreeItem.emulatorConnectionString) {
```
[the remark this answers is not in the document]
wwlorey: I updated https://github.com/microsoft/vscode-azurestorage/issues/630 to track this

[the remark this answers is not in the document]
EricJizbaMSFT: Okay but I don't want to track it separately if it will simplify this open PR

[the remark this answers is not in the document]
wwlorey: For the emulated account, I still only display blob containers & queues since it was easy to do so. Also I messed around with an 'unsupported' label but it looked too clunky

src/tree/AttachedStorageAccountsTreeItem.ts
```diff
@@ -0,0 +1,201 @@
...
+            }));
+        }
+
+        persistedAccounts.push(await this.createTreeItem(AttachedStorageAccountsTreeItem.emulatorConnectionString, this._emulatorAccountName));
+        return persistedAccounts;
+    }
+
+    private async createTreeItem(connectionString: string, name: string): Promise<AttachedStorageAccountTreeItem> {
+        let storageAccountWrapper: StorageAccountWrapper = new StorageAccountWrapper(<StorageAccount>{
+            id: name,
```
EricJizbaMSFT: I think this `StorageAccountWrapper` opens us up to risk of bugs. For example, "Open in Storage Explorer" would fail when trying to use this id. You already disabled that scenario in package.json, but theoretically typescript could've warned us about this if we did the types right

`AttachedStorageAccountTreeItem` shouldn't need `StorageAccountWrapper` at all. Instead, `IStorageRoot` and/or `IAttachedStorageRoot` should be modified to have just the needed properties. Maybe `name` and `id` without `storageAccount`? In which case `id` might be undefined if it's an attached account... I haven't figured out the exact details but hopefully that gives you enough info to work from.

An additional note: All uses of `IStorageRoot` inside tree items like `BlobContainerGroupTreeItem` on down should be changed to `IAttachedStorageRoot` since it can't guarantee it's a child of the subscription tree item at that point

src/utils/taskUtils.ts
```diff
@@ -0,0 +1,26 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.md in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+// Source: https://italonascimento.github.io/applying-a-timeout-to-your-promises/
+// tslint:disable-next-line:no-any
+export async function taskResolvesBeforeTimeout(promise: any): Promise<boolean> {
```
EricJizbaMSFT: As long as you have `taskResolvesBeforeTimeout`, you're going to have bugs like https://github.com/microsoft/vscode-azurestorage/issues/630. I'm strongly in favor of removing `taskResolvesBeforeTimeout` and all the `isActive` stuff you added. 

You're trying to circumvent the case where it takes 30+ seconds to load a service that's not supported? I get that's a little annoying, but the service isn't going to work for them anyways. I'd much rather prioritize the case where the service _is_ supported, aka don't risk bugs like #630

wwlorey: So I haven't found a good way around this - when a service isn't supported and you try and open it, it loads indefinitely *and* freezes the extension. The only way I've been able to recover the extension is by restarting VS Code. 

I know the `taskResolvesBeforeTimeout` stuff isn't ideal. Do you know of a less error-prone way to prevent this prob?

src/tree/AttachedStorageAccountsTreeItem.ts
wwlorey: > All uses of IStorageRoot inside tree items like BlobContainerGroupTreeItem on down should be changed to IAttachedStorageRoot since it can't guarantee it's a child of the subscription tree item at that point

This part is a little scary to me. It would touch ~22 files and I think we may have to make `IAttachedStorageRoot ` extend `ISubscriptionContext` to do so. Is it sufficient to protect against `IStorageRoot` only features in package.json or should I go ahead and make the switch?

EricJizbaMSFT: > This part is a little scary to me. It would touch ~22 files

I agree, but the fact that inaccurate types are used in ~22 file makes me more nervous than modifying ~22 files. FWIW - I would expect these to be very minor changes in each file - let me know if not.

> I think we may have to make IAttachedStorageRoot extend ISubscriptionContext to do so

I'm not thrilled about it, but that's what Cosmos did, too: https://github.com/microsoft/vscode-cosmosdb/blob/master/src/tree/AttachedAccountsTreeItem.ts#L346

I would still call it "inaccurate types", but at least we would have a consistent error we could check for in telemetry.

src/utils/taskUtils.ts
wwlorey: Adjusting the retry policy did the trick 🙌

src/tree/blob/BlobContainerGroupTreeItem.ts
```diff
@@ -20,25 +23,54 @@ export class BlobContainerGroupTreeItem extends AzureParentTreeItem<IStorageRoot
...
         if (clearCache) {
             this._continuationToken = undefined;
         }
 
-        let containersResponse: azureStorageBlob.ListContainersSegmentResponse = await this.listContainers(this._continuationToken);
+        let containersResponse: azureStorageBlob.ListContainersSegmentResponse;
+        try {
+            containersResponse = await this.listContainers(this._continuationToken);
+        } catch {
+            throw new Error(localize('blobServiceNotActive', 'Blob service is not currently active.'));
```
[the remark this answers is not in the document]
wwlorey: I didn't think the ENOTFOUND error message looked great in the tree view compared to the 'not active' message we could throw.

Screen Shot 2020-03-11 at 4 14 54 PM

Do you have strong feelings either way?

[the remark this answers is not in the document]
EricJizbaMSFT: That's for a blob storage account in azure, right? The "not active" message doesn't make sense to me for accounts in azure. How about something like this:  "This storage account does not support queues."

src/tree/blob/BlobContainerGroupTreeItem.ts
```diff
@@ -25,20 +28,39 @@ export class BlobContainerGroupTreeItem extends AzureParentTreeItem<IStorageRoot
...
+        } catch (error) {
+            if (this.root.isEmulated && parseError(error).errorType === 'ECONNREFUSED') {
+                return [new GenericTreeItem(this, {
+                    contextValue: 'startBlobEmulator',
+                    label: 'Start Blob Emulator',
+                    commandId: 'azureStorage.startBlobEmulator',
+                    includeInTreeItemPicker: false
+                })];
+            } else {
+                throw new Error(localize('storageAccountDoesNotSupportBlobs', 'This storage account does not support blobs.'));
```
EricJizbaMSFT: Bruh
> Any time you do stuff like this you need to check parseError(error).errorType otherwise you could be hiding a bunch of unrelated errors

This needs to be:
```
} else if (parseError(error).errorType === 'ENOTFOUND') {
    throw new Error(localize('storageAccountDoesNotSupportBlobs', 'This storage account does not support blobs.'));
} else {
    throw error;
}
```

src/tree/queue/QueueGroupTreeItem.ts
```diff
@@ -82,6 +102,22 @@ export class QueueGroupTreeItem extends AzureParentTreeItem<IStorageRoot> {
         return contextValue === QueueTreeItem.contextValue;
     }
 
+    public async isActive(): Promise<boolean> {
```
EricJizbaMSFT: throw this code in the 🗑

src/tree/blob/BlobContainerGroupTreeItem.ts
wwlorey: Oof my bad!

PR #621, opened 2020-02-28
Support file uploads in 'Upload to Azure Storage...'

wwlorey:
Fixes https://github.com/microsoft/vscode-azurestorage/issues/159

### DIFF
--- a/src/commands/uploadToAzureStorage.ts
+++ b/src/commands/uploadToAzureStorage.ts
@@ -42,14 +42,10 @@ export async function uploadToAzureStorage(actionContext: IActionContext, target
     let parentDirectory: string;
     let destinationName: string;
 
-    try {
-        await fse.readdir(resourcePath);
-
-        // resourcePath is a directory
+    if ((await fse.stat(resourcePath)).isDirectory()) {
         filePathsToUpload = await listFilePathsWithAzureSeparator(resourcePath);
         parentDirectory = resourcePath;
-    } catch {
-        // resourcePath is a file
+    } else {
         filePathsToUpload = [resourcePath];
         parentDirectory = path.dirname(resourcePath);
     }
--- a/src/commands/uploadToAzureStorage.ts
+++ b/src/commands/uploadToAzureStorage.ts
@@ -3,6 +3,8 @@
  *  Licensed under the MIT License. See License.md in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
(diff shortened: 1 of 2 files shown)


src/commands/uploadToAzureStorage.ts
```diff
@@ -15,19 +17,19 @@ import { uploadFiles } from '../utils/uploadUtils';
...
+        resourcePath = target.fsPath;
     } else {
-        sourceFolderPath = await selectWorkspaceItem(
+        resourcePath = await selectWorkspaceItem(
             ext.ui,
-            'Select folder to upload',
+            localize('selectResourceToUpload', 'Select resource to upload'),
             {
-                canSelectFiles: false,
+                canSelectFiles: true,
```
EricJizbaMSFT: Windows doesn't support `canSelectFiles` and `canSelectFolders` at the same time. https://github.com/microsoft/vscode/blob/master/src/vs/vscode.d.ts#L1669

Not sure how that affects the experience - just making sure you're aware

src/commands/uploadToAzureStorage.ts
```diff
@@ -36,17 +38,30 @@ export async function uploadToAzureStorage(actionContext: IActionContext, target
...
 
+    try {
+        await fse.readdir(resourcePath);
+
+        // resourcePath is a directory
+        filePathsToUpload = await listFilePathsWithAzureSeparator(resourcePath);
+        parentDirectory = resourcePath;
+    } catch {
+        // resourcePath is a file
+        filePathsToUpload = [resourcePath];
```
EricJizbaMSFT: Would prefer if you preemptively check if it's a file or folder using `fse.stat`. If you want to go the try/catch route - you should at least make sure the error thrown is "ENOTDIR" or whatever

src/commands/uploadToAzureStorage.ts
wwlorey: Ah I didn't see that. As written, it only lets you select folders on windows.

Do you think that behavior is OK? Or would it be better to prompt for a file or folder upload?

EricJizbaMSFT: Since folder was the previous behavior, I think that's fine for this PR. I do think the optimal approach is to have a command for each option - that's what Storage Explorer does:
Screen Shot 2020-02-28 at 10 42 43 AM

If we did that, the "files" option should probably set `canSelectMany` to true

Issue #567, opened 2020-01-14
[Suggestion] Add Log Message and Deleting Notification when deleting a storage account


v-dantch:
**OS**: All
**Build Version:** [20200113.1](https://dev.azure.com/ms-azuretools/AzCode/_build/results?buildId=13678&view=results)
**Regression:** Not a Regression

**Repro Steps:**
1. Create a new storage account --> Check all logs including _log message_ in Output window, _creating notification_ and _creating indication_.
![image](https://user-images.githubusercontent.com/34729092/72329186-44f1c380-36ef-11ea-87fb-8585ad09dfe0.png)

2. Delete a storage account --> Check the logs.

**Actual:**
Only display the deleting indication.
![image](https://user-images.githubusercontent.com/34729092/72329178-428f6980-36ef-11ea-9b2c-8ab92377cd79.png)

**Suggestion:**
Add _log message_ and _deleting notification_ when delating a storage account.

PR #626, opened 2020-02-28
Add log message and notification when deleting storage account

wwlorey:
Fixes https://github.com/microsoft/vscode-azurestorage/issues/567

### DIFF
--- a/src/tree/StorageAccountTreeItem.ts
+++ b/src/tree/StorageAccountTreeItem.ts
@@ -11,10 +11,12 @@ import * as azureStorage from "azure-storage";
 // tslint:disable-next-line:no-require-imports
 import opn = require('opn');
 import * as path from 'path';
+import * as vscode from 'vscode';
 import { commands, MessageItem, Uri, window } from 'vscode';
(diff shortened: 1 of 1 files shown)


src/tree/StorageAccountTreeItem.ts
```diff
@@ -117,14 +119,21 @@ export class StorageAccountTreeItem extends AzureParentTreeItem<IStorageRoot> {
...
             let parsedId = this.parseAzureResourceId(this.storageAccount.id);
             let resourceGroupName = parsedId.resourceGroups;
-            await storageManagementClient.storageAccounts.deleteMethod(resourceGroupName, this.storageAccount.name);
+
+            ext.outputChannel.appendLine(deletingStorageAccountString);
+            await vscode.window.withProgress({ location: vscode.ProgressLocation.Notification, title: deletingStorageAccountString }, async () => {
+                await storageManagementClient.storageAccounts.deleteMethod(resourceGroupName, this.storageAccount.name);
+            });
+
+            ext.outputChannel.appendLine(localize('successfullyDeletedStorageAccount', `Successfully deleted storage account "${this.label}".`));
```
EricJizbaMSFT: I would also do `showInformationMessage` for this message (but don't await it because that causes problems with telemetry)

Issue #87, opened 2018-02-27
Need to know how frequently users hit not supported features (e.g. opening large blobs)


StephenWeatherford:
If you try to open a large file in storage explorer,  it will give an error saying that's not supported, and give you an option to open it in Storage Explorer. I'd like to add telemetry:

1) How often users hit this type of limitation and why (too large, what size, not block blob, etc)
This should be recorded in the telemetry as an error for the specific action taking place
2) Whether the user chooses to open in Storage Explorer or not
Lower priority, probably needs to be a separate telemetry action, or maybe an additional attribute of the attempted action.
3) Size of blobs users are successfully dealing with

wwlorey:
TODO: Check is the second item is supported

PR #628, opened 2020-03-04
Add telemetry for uploading/downloading blobs

wwlorey:
Fixes https://github.com/microsoft/vscode-azurestorage/issues/87

### DIFF
--- a/src/tree/blob/BlobContainerTreeItem.ts
+++ b/src/tree/blob/BlobContainerTreeItem.ts
@@ -201,7 +201,7 @@ export class BlobContainerTreeItem extends AzureParentTreeItem<IStorageRoot> imp
             lastUploadFolder = uri;
             let filePath = uri.fsPath;
 
-            await this.checkCanUpload(filePath, context);
+            await this.checkCanUpload(context, filePath);
 
             let blobPath = await vscode.window.showInputBox({
                 prompt: 'Enter a name for the uploaded block blob (may include a path)',
@@ -419,20 +419,18 @@ export class BlobContainerTreeItem extends AzureParentTreeItem<IStorageRoot> imp
         return undefined;
     }
 
-    private async checkCanUpload(localPath: string, context: IActionContext): Promise<void> {
+    private async checkCanUpload(context: IActionContext, localPath: string): Promise<void> {
         let size = (await fse.stat(localPath)).size;
+        context.telemetry.measurements.blockBlobUploadSize = size;
         if (size > Limits.maxUploadDownloadSizeBytes) {
             context.telemetry.properties.blockBlobTooLargeForUpload = 'true';
-            context.telemetry.properties.invalidBlockBlobUploadSize = size.toString();
             await Limits.askOpenInStorageExplorer(
+                context,
                 `Please use Storage Explorer to upload files larger than ${Limits.maxUploadDownloadSizeMB}MB.`,
                 this.root.storageAccount.id,
                 this.root.subscriptionId,
                 'Azure.BlobContainer',
-                this.container.name,
-                context);
-        } else {
-            context.telemetry.properties.validBlockBlobUploadSize = size.toString();
+                this.container.name);
         }
     }
 }
--- a/src/tree/blob/BlobContainerTreeItem.ts
+++ b/src/tree/blob/BlobContainerTreeItem.ts
@@ -201,7 +201,7 @@ export class BlobContainerTreeItem extends AzureParentTreeItem<IStorageRoot> imp
             lastUploadFolder = uri;
             let filePath = uri.fsPath;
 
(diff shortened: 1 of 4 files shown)


src/tree/blob/BlobContainerTreeItem.ts
```diff
@@ -419,15 +419,20 @@ export class BlobContainerTreeItem extends AzureParentTreeItem<IStorageRoot> imp
         return undefined;
     }
 
-    private async checkCanUpload(localPath: string): Promise<void> {
+    private async checkCanUpload(localPath: string, context: IActionContext): Promise<void> {
```
nturinski: nit: Usually context is the first parameter in a function

src/tree/blob/BlobContainerTreeItem.ts
```diff
@@ -419,15 +419,20 @@ export class BlobContainerTreeItem extends AzureParentTreeItem<IStorageRoot> imp
...
             await Limits.askOpenInStorageExplorer(
                 `Please use Storage Explorer to upload files larger than ${Limits.maxUploadDownloadSizeMB}MB.`,
                 this.root.storageAccount.id,
                 this.root.subscriptionId,
                 'Azure.BlobContainer',
-                this.container.name);
+                this.container.name,
+                context);
+        } else {
+            context.telemetry.properties.validBlockBlobUploadSize = size.toString();
```
EricJizbaMSFT: IMO would be easier to have one property `context.telemetry.properties.blockBlobUploadSize` for both cases. We can always filter based on `blockBlobTooLargeForUpload` after the fact

src/tree/blob/BlobContainerTreeItem.ts
```diff
@@ -419,15 +419,20 @@ export class BlobContainerTreeItem extends AzureParentTreeItem<IStorageRoot> imp
         return undefined;
     }
 
-    private async checkCanUpload(localPath: string): Promise<void> {
+    private async checkCanUpload(localPath: string, context: IActionContext): Promise<void> {
         let size = (await fse.stat(localPath)).size;
         if (size > Limits.maxUploadDownloadSizeBytes) {
+            context.telemetry.properties.blockBlobTooLargeForUpload = 'true';
+            context.telemetry.properties.invalidBlockBlobUploadSize = size.toString();
```
EricJizbaMSFT: nit: If size is a number, you could use `telemetry.measurements` and then you wouldn't have to do `toString`

src/utils/limits.ts
```diff
@@ -13,9 +13,10 @@ export namespace Limits {
     export const maxUploadDownloadSizeMB = 4;
     export const maxUploadDownloadSizeBytes = maxUploadDownloadSizeMB * 1000 * 1000;
 
-    export async function askOpenInStorageExplorer(errorMessage: string, resourceId: string, subscriptionId: string, resourceType: ResourceType, resourceName: string): Promise<void> {
+    export async function askOpenInStorageExplorer(errorMessage: string, resourceId: string, subscriptionId: string, resourceType: ResourceType, resourceName: string, context: IActionContext): Promise<void> {
         const message = "Open container in Storage Explorer";
         if (message === await window.showErrorMessage(errorMessage, message)) {
```
EricJizbaMSFT: Since this call is awaited, we're gonna lose a lot of telemetry for people that never interact with the message. That's why we do things like this: https://github.com/microsoft/vscode-azurefunctions/blob/40858a0b50f134c731c60b2eb3864d802251028c/src/commands/deploy/notifyDeployComplete.ts#L25

src/utils/limits.ts
```diff
@@ -13,9 +13,10 @@ export namespace Limits {
...
+            context.telemetry.properties.openInStorageExplorerAfterUnsuccessfulBlockBlobTransfer = 'true';
```
EricJizbaMSFT: WOOOooowwww that is the longest prop name I've ever seen haha. Please shorten it :) The name doesn't really matter as long as it's unique enough that it doesn't conflict with other properties

src/tree/blob/BlobContainerTreeItem.ts
nturinski: I would make this just `BlockBlobUploadSize`.  We know by `blockBlobTooLargeForUpload` being `true` that it is invalid or valid.

Also, `askOpenInStorageExplorer` throws a `UserCancelledError` whenever it fails.  You could change this to be an InvalidSizeError and then we could just track that `errorType` rather than making a new telemetry property.

src/utils/limits.ts
nturinski: nit: The name of this property is quite long.  How about just `openInStorageExplorer`.  We can infer that there was an error since they can only get here from an error.

PR #633, opened 2020-03-09
Add endpoint suffix to the connection string

### LINKED ISSUE #336: Display "The specified account is disabled" after expanding the servers for attached storage account
**OS:** Windows, Mac
**Build Version:**[20190409.1](https://dev.azure.com/ms-azuretools/AzCode/_build/results?buildId=4832)

**Repro Steps:**
1. Make sure you have installed Azure Storage Explorer.
2. Sign in with Azure China account.
3. In VS Code, right click one storage account and select "Copy Connection String".
4. In Azure Storage Explorer, attach with the copied connection string and "Azure China" environment.
image
5. Try to expand it.

**Expect:**
Succeed to expand the attached storage account and servers.

**Actual:**
Pop up an error when expanding the servers like Blob Containers.
image

More Info:
For Azure Germany account, pop up below error:
image

### LINKED ISSUE #629: Can't attach storage account using connection string copied from the extension
For non-attached storage accounts, we don't include `DefaultEndpointsProtocol` when copying connection strings and attached accounts expect that (or endpoints for each resource type)

nturinski:
Fixes #629 
Fixes #336 

The connection string requires an endpointSuffix in order to connect via Will's PR: https://github.com/microsoft/vscode-azurestorage/pull/618

### DIFF
--- a/src/tree/StorageAccountTreeItem.ts
+++ b/src/tree/StorageAccountTreeItem.ts
@@ -166,7 +166,7 @@ export class StorageAccountTreeItem extends AzureParentTreeItem<IStorageRoot> {
     }
 
     async getConnectionString(): Promise<string> {
-        // storageEndpointSuffix has a '.' prefix that we need to remove
+        // https://github.com/Azure/azure-sdk-for-node/issues/4706
         const storageEndpointSuffix: string = this.root.environment.storageEndpointSuffix.charAt(0) === '.' ? this.root.environment.storageEndpointSuffix.substr(1) : this.root.environment.storageEndpointSuffix;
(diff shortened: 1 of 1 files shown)


src/tree/StorageAccountTreeItem.ts
```diff
@@ -166,7 +166,9 @@ export class StorageAccountTreeItem extends AzureParentTreeItem<IStorageRoot> {
     }
 
     async getConnectionString(): Promise<string> {
-        return `DefaultEndpointsProtocol=https;AccountName=${this.storageAccount.name};AccountKey=${this.key.value};`;
+        // storageEndpointSuffix has a '.' prefix that we need to remove
```
EricJizbaMSFT: You can link to this issue in the comment:
```
https://github.com/Azure/azure-sdk-for-node/issues/4706
```

PR #635, opened 2020-03-11
Update test for storage account advanced create

wwlorey:


test/storageAccountActions.test.ts
```diff
@@ -76,7 +76,7 @@ suite('Storage Account Actions', async function (this: ISuiteCallbackContext): P
         const accountNameAdvanced: string = getRandomHexString().toLowerCase();
         const resourceNameAdvanced: string = getRandomHexString().toLowerCase();
         resourceGroupsToDelete.push(resourceNameAdvanced);
-        await testUserInput.runWithInputs([accountNameAdvanced, '$(plus) Create new resource group', resourceNameAdvanced, 'East US'], async () => {
```
EricJizbaMSFT: We want nightly tests to cover the most common or highest priority cases. Wouldn't that be the "Yes" case?

wwlorey: Yeah that makes sense. Should we do anything to verify that static websites were configured or just check to see if the storage account was created properly?

EricJizbaMSFT: Feel free to improve the test, but "created properly" is the only thing needed for this PR IMO

Issue #640, opened 2020-03-12
The uploading log info don't include the target resource


v-dantch:
**OS**: Windows
**Build Version:** [20200312.6](https://dev.azure.com/ms-azuretools/AzCode/_build/results?buildId=16152&view=results)
**Regression**: Not a Regression

**Repro Steps:**
1. F1 --> Type and select "Azure Storage: Upload to Azure Storage..." --> Browse --> Select a folder.
2. Check the uploading info in Output window.

**Expect:**
The log info includes the uploaded resource and **target** resource.

**Actual:**
Only display the uploaded resource.
![image](https://user-images.githubusercontent.com/34729092/76502669-6bc63f00-647f-11ea-9119-989b4e9bcf29.png)

v-dantch:
Verified this issue today and it has been fixed.
![image](https://user-images.githubusercontent.com/34729092/76593969-ff663280-64b4-11ea-8981-5e01c22d40de.png)

Issue #639, opened 2020-03-12
No log info indicates the uploading has been completed in "Output" window 


v-dantch:
**OS**: Windows
**Build Version:** [20200312.6](https://dev.azure.com/ms-azuretools/AzCode/_build/results?buildId=16152&view=results)
**Regression**: Not a Regression

**Repro Steps:**
1. F1 --> Type and select "Azure Storage: Upload to Azure Storage..." --> Browse --> Select a folder with few files.
2. Check the uploading state in Output window.

**Expect:**
Display log info indicates the uploading state and display "Uploaded ..." when it has been completed.

**Actual:**
No log info indicates the uploading has been completed in "Output" window.
![image](https://user-images.githubusercontent.com/34729092/76501918-15a4cc00-647e-11ea-80d7-5e2fbb4bac4a.png)

PR #641, opened 2020-03-12
Update 'Upload to Azure Storage...' log messages

wwlorey:
Old
Screen Shot 2020-03-12 at 11 02 18 AM

New
Screen Shot 2020-03-12 at 11 01 39 AM

Fixes https://github.com/microsoft/vscode-azurestorage/issues/639
Fixes https://github.com/microsoft/vscode-azurestorage/issues/640


src/utils/uploadUtils.ts
```diff
@@ -40,4 +41,6 @@ export async function uploadFiles(
 
         transferProgress.reportToNotification(sourceFileIndex, notificationProgress);
     }
+
+    ext.outputChannel.appendLine(localize('finishedUpload', 'Finished upload to "{0}"', destTreeItem.label));
```
nturinski: Typically we just say "Uploaded..." rather than prefixing it with finished

Issue #636, opened 2020-03-12
Pop up an error when opening an attached account in Storage Explorer


v-dantch:
**OS**: Windows
**Build Version:** [20200312.6](https://dev.azure.com/ms-azuretools/AzCode/_build/results?buildId=16152&view=results)
**Regression**: Not a Regression

**Repro Steps:**
1. Drill down from "Attached Storage Accounts" --> "Local Emulator" -->"Blob Containers" --> Click "Start Blob Emulator".
2. Prompt to install Azurite --> Install.
![image](https://user-images.githubusercontent.com/34729092/76496141-564b1800-6473-11ea-82d6-c0c8e7424dea.png)
3. Create a local blob --> Select "Open in Storage Explorer" from the context menu.

**Expect:**
Not sure if it is by design.
1. No error pops up
2. Remove the un-supported actions from the context menu for attached resources.

**Actual:**
Pop up an error "Cannot retrieve storage account id for an attached account".
![image](https://user-images.githubusercontent.com/34729092/76498521-0458c100-6478-11ea-9c98-84a6da4cec03.png)

**More Info:**
This issue also reproduces when downloading a blob file.
![image](https://user-images.githubusercontent.com/34729092/76506659-7a642480-6486-11ea-9e5f-75345d229a67.png)

wwlorey:
As discussed here https://github.com/microsoft/vscode-azurestorage/pull/642 it's not worth changing that many context values to fix this

PR #642, opened 2020-03-12
Update context menu items for attached storage accounts

wwlorey:
Fixes https://github.com/microsoft/vscode-azurestorage/issues/636

### DIFF
--- a/package.json
+++ b/package.json
@@ -427,18 +427,18 @@
 				},
 				{
 					"command": "azureStorage.refresh",
-					"when": "view == azureStorage && viewItem == azureStorageAccount",
+					"when": "view == azureStorage && viewItem =~ /azureStorageAccount/",
 					"group": "9_refresh"
 				},
 				{
 					"$comment": "========= azureBlobContainerGroup =========",
 					"command": "azureStorage.createBlobContainer",
-					"when": "view == azureStorage && viewItem == azureBlobContainerGroup",
+					"when": "view == azureStorage && viewItem =~ /azureBlobContainerGroup/",
 					"group": "2_create"
 				},
 				{
 					"command": "azureStorage.refresh",
-					"when": "view == azureStorage && viewItem == azureBlobContainerGroup",
+					"when": "view == azureStorage && viewItem =~ /azureBlobContainerGroup/",
 					"group": "9_refresh"
 				},
 				{
@@ -449,7 +449,7 @@
 				},
 				{
 					"command": "azureStorage.createBlockBlob",
-					"when": "view == azureStorage && viewItem == azureBlobContainer",
+					"when": "view == azureStorage && viewItem == azureBlobContainer || viewItem == azureBlobContainer-attached",
 					"group": "2_create@1"
 				},
 				{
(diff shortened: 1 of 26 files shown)


package.json
```diff
@@ -449,7 +449,7 @@
 				},
 				{
 					"command": "azureStorage.createBlockBlob",
-					"when": "view == azureStorage && viewItem == azureBlobContainer",
+					"when": "view == azureStorage && viewItem == azureBlobContainer || viewItem == azureBlobContainer-attached",
```
nturinski: Why not just use the regexp for this?

package.json
```diff
@@ -484,22 +484,22 @@
...
 					"command": "azureStorage.refresh",
-					"when": "view == azureStorage && viewItem == azureBlobContainer",
+					"when": "view == azureStorage && viewItem == azureBlobContainer || viewItem == azureBlobContainer-attached",
 					"group": "9_refresh"
 				},
 				{
 					"$comment": "========= azureBlob =========",
 					"command": "azureStorage.deleteBlob",
-					"when": "view == azureStorage && viewItem == azureBlob"
+					"when": "view == azureStorage && viewItem == azureBlob || viewItem == azureBlob-attached"
```
nturinski: Same with these

EricJizbaMSFT:
Yes I agree with Nathan. The amount of changes are not worth the benefit it provides

package.json
wwlorey: I'd have to exclude the matches on `azureBlobContainerGroup` so I decided to explicitly check

wwlorey:
I'd be fine with putting it off. It is nice to only have the context menu options that apply but it's a lot of changes

src/tree/AttachedStorageAccountsTreeItem.ts
```diff
@@ -65,7 +63,7 @@ export class AttachedStorageAccountsTreeItem extends AzureParentTreeItem {
     }
 
     public isAncestorOfImpl(contextValue: string): boolean {
-        return contextValue !== StorageAccountTreeItem.contextValue;
+        return !!contextValue.match(/-attached/);
```
wwlorey: Just wanted to bring this up so everybody's aware of it:

Without this change to `isAncestorOfImpl`, you can choose an attached or emulated account for commands like `Create Storage Account`

Screen Shot 2020-03-12 at 4 16 42 PM

And that leads to an error like this

Screen Shot 2020-03-12 at 4 18 10 PM

I don't think that error is a dealbreaker but worth mentioning

EricJizbaMSFT: I feel like you could probably do a one-line fix as a separate PR to fix that

EricJizbaMSFT: How about this:
```
return contextValue !== StorageAccountTreeItem.contextValue && contextValue !== SubscriptionTreeItem.contextValue;
```

wwlorey: Ah that fixes it!

wwlorey:
Closing since the benefit this provides isn't worth this many changes

PR #643, opened 2020-03-12
Fix nightly tests after adding support for attached accounts

wwlorey:


src/tree/AzureAccountTreeItem.ts
```diff
@@ -9,9 +9,17 @@ import { AttachedStorageAccountsTreeItem } from "./AttachedStorageAccountsTreeIt
 import { SubscriptionTreeItem } from "./SubscriptionTreeItem";
 
 export class AzureAccountTreeItem extends AzureAccountTreeItemBase {
+    private _isTestAccount: boolean;
+
     public constructor(testAccount?: {}) {
         super(undefined, testAccount);
-        ext.attachedStorageAccountsTreeItem = new AttachedStorageAccountsTreeItem(this);
+
+        if (testAccount) {
```
EricJizbaMSFT: As a general rule, you should modify the test to work with the production code, not the other way around. What's the reason you did this instead of modifying the test? It's possible I would know a quick fix for ya

wwlorey: Yeah I was struggling to figure out why the tests were failing. Do you have any leads?

EricJizbaMSFT: Actually I bet the prompt you mentioned in https://github.com/microsoft/vscode-azurestorage/pull/642#discussion_r391955086 was causing problems. See if this PR fixed the tests: https://github.com/microsoft/vscode-azurestorage/pull/644

wwlorey: Ah yep that's the right direction, thanks!

PR #644, opened 2020-03-12
Fix ancestor check for AttachedStorageAccountsTreeItem

wwlorey:


src/tree/AttachedStorageAccountsTreeItem.ts
```diff
@@ -65,7 +74,16 @@ export class AttachedStorageAccountsTreeItem extends AzureParentTreeItem {
...
+        return contextValue !== StorageAccountTreeItem.contextValue &&
+            contextValue !== SubscriptionTreeItem.contextValue &&
+            contextValue !== BlobContainerGroupTreeItem.contextValue &&
+            contextValue !== FileShareGroupTreeItem.contextValue &&
+            contextValue !== QueueGroupTreeItem.contextValue &&
+            contextValue !== TableGroupTreeItem.contextValue &&
+            contextValue !== BlobContainerTreeItem.contextValue &&
+            contextValue !== FileShareTreeItem.contextValue &&
+            contextValue !== QueueTreeItem.contextValue &&
+            contextValue !== TableTreeItem.contextValue;
```
wwlorey: Is this too extra?

EricJizbaMSFT: Yeah we don't want this. People should be able to do things like "Create file share..." from the command palette even for attached accounts, but this would prevent that

PR #699, opened 2020-07-08
Added changes to work against AzureStack Environment

yatna:
These changes allow Storage extension to work with Azure Stack Environments

### DIFF
--- a/package.json
+++ b/package.json
@@ -799,11 +799,12 @@
         "webpack-cli": "^3.1.2"
     },
     "dependencies": {
-        "@azure/storage-blob": "^12.0.0",
+        "@azure/storage-blob": "12.0.2",
         "@azure/storage-file-share": "^12.0.0",
         "@types/mime": "^2.0.1",
         "azure-arm-resource": "^3.1.1-preview",
-        "azure-arm-storage": "^5.1.1-preview",
+        "azure-arm-storage3": "npm:azure-arm-storage@^3.0.0",
+        "azure-arm-storage": "npm:azure-arm-storage@^5.1.1-preview",
         "azure-storage": "^2.10.0",
         "fs-extra": "^8.0.0",
         "glob": "^7.1.2",
--- a/src/tree/StorageAccountTreeItem.ts
+++ b/src/tree/StorageAccountTreeItem.ts
@@ -2,20 +2,18 @@
  *  Copyright (c) Microsoft Corporation. All rights reserved.
  *  Licensed under the MIT License. See License.md in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
-
 import * as azureStorageBlob from '@azure/storage-blob';
 import * as azureStorageShare from '@azure/storage-file-share';
 import { StorageManagementClient } from 'azure-arm-storage';
 import { StorageAccountKey } from 'azure-arm-storage/lib/models';
 import * as azureStorage from "azure-storage";
-// tslint:disable-next-line:no-require-imports
-import opn = require('opn');
 import * as path from 'path';
 import * as vscode from 'vscode';
 import { commands, MessageItem, Uri, window } from 'vscode';
 import { AzureParentTreeItem, AzureTreeItem, AzureWizard, createAzureClient, DialogResponses, IActionContext, ISubscriptionContext, UserCancelledError } from 'vscode-azureextensionui';
 import { getResourcesPath, staticWebsiteContainerName } from '../constants';
 import { ext } from "../extensionVariables";
+import { getStorageManagementClient, ifStack } from '../utils/environmentUtils';
 import { localize } from '../utils/localize';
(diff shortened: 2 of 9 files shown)


wwlorey:
Hi @yatna, thanks for submitting.

Have those comments made in your fork been addressed yet? If not, could you take a look at those? Also, I see a lot of use of the `any` type. I assume this was at least partly to avoid making changes to the [vscode-azureextensionui](https://github.com/microsoft/vscode-azuretools/tree/master/ui) package. But it would be preferable to use proper typing and update that package accordingly. If you have any questions about that process, please feel free to reach out.

EricJizbaMSFT:
Just want to say we will be switching over to the new [azure-sdk-for-js](https://github.com/Azure/azure-sdk-for-js) soon (could be in master as soon as this month, but will need a bit of time before actually releasing). The code being touched by this PR seems very related, so might be best to wait until that's done

Tracking issue: https://github.com/microsoft/vscode-azuretools/issues/510

src/tree/StorageAccountTreeItem.ts
```diff
@@ -2,20 +2,18 @@
  *  Copyright (c) Microsoft Corporation. All rights reserved.
  *  Licensed under the MIT License. See License.md in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
-
```
wwlorey, line 5: Please revert whitespace changes like this

src/tree/StorageAccountTreeItem.ts
```diff
@@ -30,6 +28,9 @@ import { IStorageRoot } from './IStorageRoot';
 import { QueueGroupTreeItem } from './queue/QueueGroupTreeItem';
 import { TableGroupTreeItem } from './table/TableGroupTreeItem';
 
+// tslint:disable-next-line:no-require-imports
+import opn = require('opn');
```
wwlorey, line 32: Also revert this change

package.json
```diff
@@ -799,11 +799,12 @@
         "webpack-cli": "^3.1.2"
     },
     "dependencies": {
-        "@azure/storage-blob": "^12.0.0",
+        "@azure/storage-blob": "12.0.2",
```
wwlorey, line 802: Do your changes require something from v12.0.2? If not, please revert

src/tree/StorageAccountTreeItem.ts
```diff
@@ -128,7 +129,13 @@ export class StorageAccountTreeItem extends AzureParentTreeItem<IStorageRoot> {
         const result = await ext.ui.showWarningMessage(message, { modal: true }, DialogResponses.deleteResponse, DialogResponses.cancel);
         if (result === DialogResponses.deleteResponse) {
             const deletingStorageAccount: string = localize('deletingStorageAccount', 'Deleting storage account "{0}"...', this.label);
-            let storageManagementClient = createAzureClient(this.root, StorageManagementClient);
+            let storageManagementClient: StorageManagementClient;
+            if (ifStack()) {
+                // tslint:disable-next-line: no-unsafe-any
```
wwlorey, line 134: While disabling tslint rules like this fixes lint errors, it's still preferable to avoid the `any` type (this makes code more robust and our lives as devs easier)

Can you update `createAzureClient` to also accept the storage management client type you need?

package.json
yatna: Yes, 12.0.2 is required. I can add keep 2 versions, both ^12.0.0 and 12.0.2, like I did for azure arm storage, if required

wwlorey: Keeping just 12.0.2 is fine. Go ahead and put the `^` in front to match how the old version was

EricJizbaMSFT:
Hey guys - maybe I wasn't clear enough in my above comment. We're changing to the new "azure-sdk-for-js" within the next couple of weeks and we don't plan to release an update to this extension until after that happens. @yatna I appreciate the effort you've already put in to this PR, but I don't want us to spend more time on throwaway code. If you're trying to get Azure Stack support in for some deadline - let us know and we can discuss options. Otherwise, you're welcome to follow  [this issue](https://github.com/microsoft/vscode-azuretools/issues/510) and file a new PR after that work is done.

src/utils/environmentUtils.ts
```diff
@@ -0,0 +1,95 @@
...
+        try {
+            // tslint:disable-next-line: non-literal-require
+            const data = require(settingsPath);
+            // tslint:disable-next-line: no-unsafe-any
+            baseUrl = data["azure.ppe"].resourceManagerEndpointUrl;
+            // tslint:disable-next-line: no-unsafe-any
+            accountType = data["azure.cloud"];
+        } catch (err) {
+            console.log("Error parsing settings.json. Fallings back to default endpoints.");
+        } finally {
```
bganapa, line 34: target_azurestack_api_version - true

Issue #686, opened 2020-06-09
Extension limits upload size to be max 4MB, can't upload larger files


jmew:
As a data scientist, I am almost always dealing with big data/files. Some examples are uploading datasets for training (this could range from gigabytes to hundreds of gigabytes), or uploading your trained model (which could be from hundreds of MB to a gigabyte). It would be nice to be able to upload these files from the extension rather than having to install storage explorer. 

As well, in the codespaces scenario, VS Code is no longer running on the local machine,  its all running on a VM in the cloud, such that you can access the codespace in any browser. So having to install the storage explorer may not always be feasible depending on a data scientist's environment.

wwlorey:
Fixed for uploads in https://github.com/microsoft/vscode-azurestorage/pull/729 and downloads in https://github.com/microsoft/vscode-azurestorage/pull/779

PR #703, opened 2020-07-15
Remove 4MB upload/download limit for blobs

wwlorey:
After removing the limit, I uploaded a 490MB file and it worked fine (it just took a long time of course). I assume its still worth switching over to azcopy for this functionality though.

Related to https://github.com/microsoft/vscode-azurestorage/issues/686

### DIFF
--- a/src/tree/blob/BlobContainerTreeItem.ts
+++ b/src/tree/blob/BlobContainerTreeItem.ts
@@ -18,7 +18,6 @@ import { TransferProgress } from '../../TransferProgress';
 import { createBlobContainerClient, createBlockBlobClient, createChildAsNewBlockBlob, doesBlobExist, IBlobContainerCreateChildContext, loadMoreBlobChildren } from '../../utils/blobUtils';
 import { throwIfCanceled } from '../../utils/errorUtils';
 import { listFilePathsWithAzureSeparator } from '../../utils/fs';
-import { Limits } from '../../utils/limits';
 import { uploadFiles } from '../../utils/uploadUtils';
 import { ICopyUrl } from '../ICopyUrl';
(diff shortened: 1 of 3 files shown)


src/utils/limits.ts
```diff
@@ -8,21 +8,15 @@ import { IActionContext, UserCancelledError } from "vscode-azureextensionui";
 import { ResourceType } from "../storageExplorerLauncher/ResourceType";
```
wwlorey, line 8: I'll rename this file to something more appropriate if this change is accepted

nturinski:
While I think a 4MB limit is ridiculous, I think we should still have one if we're not using azcopy just because of how long it can potentially take.

That being said, is the storage release supposed to have azcopy implementation in it?  In that case, I don't know if we really need this PR yet.

wwlorey:
Jing asked me to remove the limit as a first step to implementing azcopy. If we'd rather wait to do this until azcopy is implemented I can close this PR

src/tree/blob/BlobContainerTreeItem.ts
```diff
@@ -418,19 +415,4 @@ export class BlobContainerTreeItem extends AzureParentTreeItem<IStorageRoot> imp
 
         return undefined;
     }
-
-    private async checkCanUpload(context: IActionContext, localPath: string): Promise<void> {
-        let size = (await fse.stat(localPath)).size;
-        context.telemetry.measurements.blockBlobUploadSize = size;
```
EricJizbaMSFT, line 424: Can you make sure we have the same telemetry before and after your changes

wwlorey:
Closing since we'll use AzCopy if the file size is > 4MB and still need this code to check that

Issue #713, opened 2020-07-22
Can't upload files to File Share from the Az Storage VS Code Extension


jmew:
From the Az Storage ext, you can't upload files to file share (You can only do it form Storage Explorer but this information isnt surfaced to the user)

![image](https://user-images.githubusercontent.com/15910920/88205619-ec513500-cc01-11ea-8e97-431af8356237.png)

This is in contrast to Storage Blobs which has the option to "Upload block blob"

![image](https://user-images.githubusercontent.com/15910920/88205749-14d92f00-cc02-11ea-87ee-2694c915e7aa.png)

PR #715, opened 2020-07-23
Rename `uploadBlockBlob` ➡️  `uploadFile`

wwlorey:
Support uploading file share files as well as blobs

Fixes https://github.com/microsoft/vscode-azurestorage/issues/713

### DIFF
--- a/src/commands/uploadFile.ts
+++ b/src/commands/uploadFile.ts
@@ -4,9 +4,10 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { stat } from "fs-extra";
-import { basename } from "path";
+import { basename, dirname } from "path";
 import { OpenDialogOptions, Uri, window } from "vscode";
 import { DialogResponses, IActionContext } from "vscode-azureextensionui";
+import { ext } from "../extensionVariables";
 import { BlobContainerTreeItem } from "../tree/blob/BlobContainerTreeItem";
 import { FileShareTreeItem } from "../tree/fileShare/FileShareTreeItem";
 import { doesBlobExist } from "../utils/blobUtils";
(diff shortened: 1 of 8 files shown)


src/commands/uploadFile.ts
```diff
@@ -28,58 +36,55 @@ export async function uploadBlockBlob(context: IActionContext, treeItem: BlobCon
...
+            prompt: localize('enterNameForFile', 'Enter a name for the uploaded file'),
+            value: basename(localFilePath),
+            validateInput: treeItem instanceof BlobContainerTreeItem ? BlobContainerTreeItem.validateBlobName : validateFileName
         });
-        if (blobPath) {
-            if (await doesBlobExist(treeItem, blobPath)) {
-                const result = await vscode.window.showWarningMessage(
-                    `A blob with the name "${blobPath}" already exists. Do you want to overwrite it?`,
+        if (remoteFilePath) {
+            if (treeItem instanceof BlobContainerTreeItem ? await doesBlobExist(treeItem, remoteFilePath) : await doesFileExist(remoteFilePath, treeItem, '', treeItem.shareName)) {
```
EricJizbaMSFT, line 51: Just double checking - how come you're passing `''` as the directory path?

src/commands/uploadFile.ts
```diff
@@ -28,58 +36,55 @@ export async function uploadBlockBlob(context: IActionContext, treeItem: BlobCon
...
+                await window.showWarningMessage(
+                    localize('fileAlreadyExists', `A file with the name "${remoteFilePath}" already exists. Do you want to overwrite it?`),
                     { modal: true },
-                    DialogResponses.yes, DialogResponses.cancel);
-                if (result !== DialogResponses.yes) {
```
EricJizbaMSFT, line 48: You removed this check, but you can only do that if you use `ext.ui.showWarningMessage` instead of `window.showWarningMessage`. Double check that cancel actually cancels

src/tree/blob/BlobDirectoryTreeItem.ts
```diff
@@ -63,7 +64,7 @@ export class BlobDirectoryTreeItem extends AzureParentTreeItem<IStorageRoot> {
         return children;
     }
 
-    public async createChildImpl(context: ICreateChildImplContext & Partial<IExistingBlobContext> & IBlobContainerCreateChildContext): Promise<AzExtTreeItem> {
+    public async createChildImpl(context: ICreateChildImplContext & Partial<IExistingFileContext> & IBlobContainerCreateChildContext): Promise<AzExtTreeItem> {
```
EricJizbaMSFT, line 67: How come this method references `IExistingFileContext`, but doesn't seem to use `localFilePath` or `remoteFilePath`?

src/commands/uploadFile.ts
wwlorey: In most other places where `doesFileExist` is called, there are separate variables for the file name and directory path. But in this case, `remoteFilePath` includes the name and path.

I did it this way to save steps but to match the pattern I'll split em up

src/tree/blob/BlobDirectoryTreeItem.ts
wwlorey: That's a good question.. Looks like `Partial<IExistingBlobContext>` wasn't needed in the first place, I'll remove it

PR #724, opened 2020-07-30
Remove references to old file share SDK

wwlorey:
When onboarding the new SDK, file uploads for empty files would hang indefinitely so we reverted to the old SDK for those few special cases. [That issue](https://github.com/Azure/azure-sdk-for-js/issues/6904) was fixed in v12.1.1 of the new SDKs. I only ran into the problem for file shares but it looks like they shipped a [similar fix](https://github.com/XiaoningLiu/azure-sdk-for-js/commit/269aeeb117334a173b505bda2f28a4abd76bf3a9#diff-1ac4ced9d9574e91d2b0b5da1658680dR6) for blobs, so I bumped both SDK versions here.

### DIFF
--- a/package.json
+++ b/package.json
@@ -804,8 +804,8 @@
         "webpack-cli": "^3.1.2"
     },
     "dependencies": {
-        "@azure/storage-blob": "^12.0.0",
-        "@azure/storage-file-share": "^12.0.0",
+        "@azure/storage-blob": "^12.1.1",
+        "@azure/storage-file-share": "^12.1.1",
         "@types/mime": "^2.0.1",
         "azure-arm-resource": "^3.1.1-preview",
         "azure-arm-storage": "^5.1.1-preview",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -804,8 +804,8 @@
         "webpack-cli": "^3.1.2"
     },
     "dependencies": {
-        "@azure/storage-blob": "^12.0.0",
-        "@azure/storage-file-share": "^12.0.0",
+        "@azure/storage-blob": "^12.1.1",
+        "@azure/storage-file-share": "^12.1.1",
         "@types/mime": "^2.0.1",
```
EricJizbaMSFT, line 809: Aw man the PR title made me think we could remove the whole "azure-storage" dependency. Not yet?

wwlorey: Lol sorry to disappoint. We're still using it for tables & queues :'(

PR #725, opened 2020-07-30
Add function 'warnFileAlreadyExists'

wwlorey:
A lot of these smaller PRs will be preparing for AzCopy changes

### DIFF
--- a/src/utils/uploadUtils.ts
+++ b/src/utils/uploadUtils.ts
@@ -5,7 +5,7 @@
 
 import * as path from 'path';
 import * as vscode from 'vscode';
-import { parseError, TelemetryProperties } from "vscode-azureextensionui";
+import { DialogResponses, parseError, TelemetryProperties } from "vscode-azureextensionui";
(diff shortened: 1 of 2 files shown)


src/utils/uploadUtils.ts
```diff
@@ -44,3 +44,12 @@ export async function uploadFiles(
 
     ext.outputChannel.appendLog(localize('finishedUpload', 'Uploaded to "{0}".', destTreeItem.label));
 }
+
+export async function warnFileAlreadyExists(filePath: string): Promise<void> {
+    await ext.ui.showWarningMessage(
```
EricJizbaMSFT, line 49: If you're going to make this shared code, I'd like to clean it up a bit.
1. Remove "DialogResponse.cancel" because it's included by default
1. Remove "Do you want to overwrite it?" from the message and change "DialogResponses.yes" to an "Overwrite" button. As a general rule, it's better to have the button be the specific action you're about to perform than a "yes" or "no", which requires slightly more thought when reading the prompt

PR #728, opened 2020-07-31
Update to new azure-sdk-for-js

EricJizbaMSFT:
Means people (aka us) will have to use VS Code insiders until 1.48 goes stable

### DIFF
--- a/test/nightly/global.resource.test.ts
+++ b/test/nightly/global.resource.test.ts
@@ -3,8 +3,8 @@
  *  Licensed under the MIT License. See LICENSE.md in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import { ResourceManagementClient } from 'azure-arm-resource';
-import StorageManagementClient from 'azure-arm-storage';
+import { ResourceManagementClient } from '@azure/arm-resources';
+import { StorageManagementClient } from '@azure/arm-storage';
 import * as vscode from 'vscode';
 import { TestAzureAccount } from 'vscode-azureextensiondev';
 import { AzExtTreeDataProvider, AzureAccountTreeItem, createAzureClient, ext } from '../../extension.bundle';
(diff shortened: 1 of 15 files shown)


test/nightly/global.resource.test.ts
```diff
@@ -37,7 +37,7 @@ suiteTeardown(async function (this: Mocha.Context): Promise<void> {
 
 export async function beginDeleteResourceGroup(resourceGroup: string): Promise<void> {
     const client: ResourceManagementClient = createAzureClient(testAccount.getSubscriptionContext(), ResourceManagementClient);
-    if (await client.resourceGroups.checkExistence(resourceGroup)) {
+    if ((await client.resourceGroups.checkExistence(resourceGroup)).body) {
```
wwlorey, line 40: Does just calling `checkExistence` not work here?

EricJizbaMSFT: Nope - it has to do with how the new sdk handles response objects. The old sdk usually had two separate methods (i.e. `checkExistence` and `checkExistenceWithHttpOperationRepsonse`) so you chose up front if you wanted the whole response. The new sdk kind of mashes those together and always returns an object with the result and the response. For results that are already an object, they just stick a "_response" property on top of that. For results that are a primitive (like a boolean in this case), they have to create a new object to contain both the result and "_response"
```
/**
 * Contains response data for the checkExistence operation.
 */
export type ResourceGroupsCheckExistenceResponse = {
  /**
   * The parsed response body.
   */
  body: boolean;

  /**
   * The underlying HTTP response.
   */
  _response: msRest.HttpResponse & {
      /**
       * The response body as text (string format)
       */
      bodyAsText: string;

      /**
       * The response body as parsed JSON or XML
       */
      parsedBody: boolean;
    };
};
```

PR #729, opened 2020-07-31
Add AzCopy utilities and support single blob uploads

wwlorey:
I tried to only include the minimum needed for single blob uploads for this first PR. As we discussed in triage, if AzCopy isn't installed we just open a link to the download page.


src/utils/uploadUtils.ts
```diff
@@ -34,7 +34,7 @@ export async function uploadFiles(
         ext.outputChannel.appendLog(localize('uploadingFile', 'Uploading "{0}" to "{1}"...', sourceFilePath, destTreeItem.label));
 
         try {
-            await destTreeItem.uploadLocalFile(sourceFilePath, destFilePath, true /* suppressLogs */);
+            await destTreeItem.uploadLocalFile(sourceFilePath, destFilePath, false /* useAzCopy */, true /* suppressLogs */);
```
wwlorey, line 37: I'm waiting to put the PR out for multiple file uploads

webpack.config.js
```diff
@@ -18,7 +18,17 @@ let DEBUG_WEBPACK = !!process.env.DEBUG_WEBPACK;
...
+    externals: {
+        './getCoreNodeModule': 'commonjs getCoreNodeModule',
+
+        // @azure-tools/azcopy-node
+        '@azure-tools/azcopy-darwin': true,
+        '@azure-tools/azcopy-linux': true,
+        '@azure-tools/azcopy-win32': true,
+        '@azure-tools/azcopy-win64': true,
+        '@storage-explorer/macos-keychain': true,
+        'keytar': true
```
wwlorey, line 30: The AzCopy executable is supplied by us when we create an `AzCopyClient`, so I marked the exes as externals here. And macos-keychain & keytar are only invoked in `IRemoteAuthLocation` which we don't use.

EricJizbaMSFT: > As we discussed in triage, if AzCopy isn't installed we just open a link to the download page.

As you mentioned, I thought we discussed it's better to start by making the user install azcopy. We can _consider_ shipping azcopy with our vsix in a separate PR, but I don't want to do it in the first PR.

wwlorey: My b - I wasn't very clear with my comment. We provide a path to AzCopy when we create an `AzCopyClient`, not the executable itself. We still make the user to download AzCopy and add it to their path

EricJizbaMSFT: I'm confused - if we still require users to download AzCopy - why does this PR even have to mention modules like "azcopy-win64"?

wwlorey: Not listing those packages as externals was causing the webpack process to fail since they are expected by @azure-tools/azcopy-node

EricJizbaMSFT: Well if "@azure-tools/azcopy-node" expects them to be there and we don't include them, that might lead to problems. Would be better if "@azure-tools/azcopy-node" was a standalone package

wwlorey: Providing the executable path overrides the use of those packages. It would be nice if there was a safer way to handle this though

EricJizbaMSFT: @MRayermannMSFT can we remove packages like "@azure-tools/azcopy-win32" from being explicit dependencies of "@azure-tools/azcopy-node"?

MRayermannMSFT: @EricJizbaMSFT, would you be ok if they were `peerDependencies`?

https://nodejs.org/en/blog/npm/peer-dependencies/

MRayermannMSFT: Also, what about keytar and macos-keychain? Can those stay as dependencies?

EricJizbaMSFT: My ideal scenario would be something like [simple-git](https://www.npmjs.com/package/simple-git) which is basically just a lightweight wrapper for the cli. Will's PR is using `azcopy` in the user's path, so I wouldn't expect to need a dependency like "azcopy-win32". Correct me if I'm wrong, but a peer dependency would still put some sort of requirement on those packages. And yes, it would be great if it didn't depend on "keytar" or "macos-keychain", either.

I just want the interface between us and the cli (the interface should be OS-agnostic, right?) to be completely separate from installing the CLI itself (which is much more complicated for us to handle because it depends on the OS).

MRayermannMSFT: The way azcopy-node is being used by the storage extension is atypical from the point of view of how azcopy-node is designed. azcopy-node is supposed to be a library that helps an application call into AzCopy. This includes sharing authentication secrets between AzCopy and the calling application when invoking AzCopy using AAD credentials. The design also assumes that the calling application ships all of its dependencies, including AzCopy. Typically not shipping your dependencies is considered the less than ideal experience.

Since the storage extension neither supports AAD scenarios nor does it ship all of its dependencies, that makes it atypical from the type of application azcopy-node was designed to be used with. From my point of view, I don't understand what is wrong with using webpack.externals to work around this.

> a peer dependency would still put some sort of requirement on those packages

If I'm using package `foo`, and it has `peerDependency` `bar`, and I do not install `bar`, then I'll get a warning from NPM. And if I try to use `foo` without installing `bar`, `foo` is free to handle that scenario as it pleases. Is that what your understanding of peer dependencies is?

At the end of the day, we have two concerns.

Our main concern is making sure people understand where to get the "right" AzCopy executable. The last thing we want is for someone to install azcopy-node, realize they need an executable, and then:

- install a malicious one from NPM (would be bad because malicious exe's are bad)
- get one not from NPM (would be bad if the user is on Linux and is relying on AAD credential scenarios, as the NPM version of the Linux exe uses different credential storing APIs than the regular version)
- install an incompatible version (would be bad depending on if the incompatible version causes issues that lead to data loss or things of that nature)

By making the exe's peer dependencies, this concern is kinda of assuaged, but I'd have to think more if it's really enough.

Our secondary concern is making it possible for people to simply install azopy-node and start using it right away. By making the exe modules `dependencies` and by making every module needed to support all authentication scenarios `dependencies`, the user just has to install 1 package, and everything needed is installed for them as well.

If all of the packages we are discussing were peer dependencies, then this concern would no longer be addressed.

> I just want the interface between us and the cli (the interface should be OS-agnostic, right?) to be completely separate from installing the CLI itself (which is much more complicated for us to handle because it depends on the OS).

The interface between you and azcopy-ndoe is OS-agnostic. Unfortunately the interface between azcopy-node and AzCopy is not. There's nothing we can do about that.

EricJizbaMSFT: Talked with Matt offline and still unsure how we want to handle this. When picking an npm module, I always recommend we choose ones with good docs and few dependencies (ideally 0-2). Unfortunately, this module does not fit that bill and isn't even open source. @wwlorey - how complicated is it to just spawn a child process and call azcopy ourselves? How many calls to azcopy would we have for our scenarios? I'm wondering if `@azure-tools/azcopy-node` is overkill for us. If so, we wouldn't have to bug the other team to open source, add docs, and continue discussing the dependency concerns

src/utils/azCopyUtils.ts
```diff
@@ -0,0 +1,146 @@
...
+    fromTo: FromToOption,
+    notificationProgress?: Progress<{
+        message?: string | undefined;
+        increment?: number | undefined;
+    }>,
+    throwIfCanceled?: () => void,
+): Promise<void> {
+    await validateAzCopyInstalled();
+    const copyClient: AzCopyClient = new AzCopyClient({ exe: ext.azCopyExePath });
+    const copyOptions: ICopyOptions = { fromTo, overwriteExisting: "true", recursive: true, followSymLinks: true };
```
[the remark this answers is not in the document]
wwlorey: Here's its definition in "@azure-tools/azcopy-node": 
```json
overwriteExisting?: "true" | "false" | "prompt";
```

src/utils/azCopyUtils.ts
```diff
@@ -0,0 +1,146 @@
...
+
+export async function azCopyBlobTransfer(
+    src: ILocalLocation,
+    dst: IRemoteSasLocation,
+    transferProgress: TransferProgress,
+    notificationProgress?: Progress<{
+        message?: string | undefined;
+        increment?: number | undefined;
+    }>,
+    throwIfCanceled?: () => void
```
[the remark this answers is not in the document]
wwlorey: Yeah those are used in AzCopy code that hasn't been submitted yet. I'll take them out until code that uses them is introduced

webpack.config.js
wwlorey: Our use of AzCopy is pretty limited so it looks like implementing whatever functionality we need from it on our end would be reasonable (as long as the storage extension isn't trying to become Storage Explorer 😜)

wwlorey:
Per offline discussion: We'll ship the AzCopy executables with our extension. I will investigate to see how easy it is to port over only what we need from "@azure-tools/azcopy-node". Otherwise, we'll take that package as a dependency

src/commands/uploadFile.ts
```diff
@@ -56,10 +52,15 @@ export async function uploadFile(context: IActionContext, treeItem: BlobContaine
                     const result = await treeItem.treeDataProvider.findTreeItem(id, context);
                     if (result) {
                         // A treeItem for this file already exists, no need to do anything with the tree, just upload
-                        await treeItem.uploadLocalFile(localFilePath, remoteFilePath);
+                        await treeItem.uploadLocalFile(localFilePath, remoteFilePath, await shouldUseAzCopy(context, localFilePath));
                         return;
                     }
                 } catch (err) {
+                    const parsedError: IParsedError = parseError(err);
+                    if (parsedError.message === 'AzCopy Transfer Failed') {
```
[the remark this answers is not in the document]
wwlorey: I didn't realize that linked issue had been closed. I went ahead and removed the try/catch

src/tree/StorageAccountTreeItem.ts
```diff
@@ -149,6 +151,18 @@ export class StorageAccountTreeItem extends AzureParentTreeItem<IStorageRoot> {
             storageAccountId: this.storageAccount.id,
             isEmulated: false,
             primaryEndpoints: this.storageAccount.primaryEndpoints,
+            generateSasToken: () => {
```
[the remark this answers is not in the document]
wwlorey: I moved the `generateSasToken` configuration to azCopyUtils for now - it'll probably change again when I refactor the folder structure

src/tree/blob/BlobContainerTreeItem.ts
```diff
@@ -305,27 +307,32 @@ export class BlobContainerTreeItem extends AzureParentTreeItem<IStorageRoot> imp
         }
     }
 
-    public async uploadLocalFile(filePath: string, blobPath: string, suppressLogs: boolean = false): Promise<void> {
-        const blobFriendlyPath: string = `${this.friendlyContainerName}/${blobPath}`;
-        const blockBlobClient: azureStorageBlob.BlockBlobClient = createBlockBlobClient(this.root, this.container.name, blobPath);
-
+    public async uploadLocalFile(filePath: string, blobPath: string, useAzCopy: boolean = false, suppressLogs: boolean = false): Promise<void> {
```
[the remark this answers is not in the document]
wwlorey: Added `IUploadLocalFileOptions` to take care of this

src/utils/azCopyUtils.ts
```diff
@@ -0,0 +1,146 @@
...
+        resourceUri = shareClient.url;
+    }
+
+    const sasToken: string = treeItem.root.generateSasToken();
+    // Ensure path begins with '/' to transfer properly
+    const path: string = destinationPath[0] === '/' ? destinationPath : `/${destinationPath}`;
+    return { type: "RemoteSas", sasToken, resourceUri, path, useWildCard: false };
+}
+
+export async function azCopyBlobTransfer(
```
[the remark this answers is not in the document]
wwlorey: It looks like uploadLocalFile is used in several places so it looks like it's best to keep that code in BlobContainerTreeItem.ts. Would moving the AzCopy code to a src/azCopy folder be better than having it as a util file?

.gitignore
```diff
@@ -20,7 +20,7 @@
 x64/
 x86/
 bld/
-[Bb]in/
+/[Bb]in/
```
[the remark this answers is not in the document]
wwlorey: The node_modues folder and some of its sub-folders that live in the resources folder were being excluded. So I added the slashes to only ignore what's in the root of the repo

src/commands/azCopy/azCopyTransfer.ts
```diff
@@ -0,0 +1,70 @@
...
+    dst: IRemoteSasLocation,
+    transferProgress: TransferProgress,
+    fromTo: FromToOption,
+): Promise<void> {
+    const exe: string = getAzCopyExePath();
+    const copyClient: AzCopyClient = new AzCopyClient({ exe });
+    const copyOptions: ICopyOptions = { fromTo, overwriteExisting: "true", recursive: true, followSymLinks: true };
+    let jobId: string = await startAndWaitForCopy(copyClient, src, dst, copyOptions, transferProgress);
+    let finalTransferStatus = (await copyClient.getJobInfo(jobId)).latestStatus;
+    if (!finalTransferStatus || finalTransferStatus.JobStatus === 'Failed') {
```
nturinski, line 33: Should we only be checking for "Failed"?  I see there's a couple of other failure states, like "Cancelling" and "Cancelled".

We also don't account for "CompletedWithErrors" or "CompletedWithSkipped".  We should probably report those to the user if there were some errors while transferring a bunch of files.

src/commands/azCopy/azCopyTransfer.ts
```diff
@@ -0,0 +1,70 @@
...
+    options: ICopyOptions,
+    transferProgress: TransferProgress,
+): Promise<string> {
+    let jobId: string = await copyClient.copy(src, dst, options);
+    let status: TransferStatus | undefined;
+    let finishedWork: number;
+    while (!status || status.StatusType !== 'EndOfJob') {
+        status = (await copyClient.getJobInfo(jobId)).latestStatus;
+        // tslint:disable-next-line: strict-boolean-expressions
+        finishedWork = status && (src.useWildCard ? status.TransfersCompleted : status.BytesOverWire) || 0;
```
nturinski, line 51: Sorry, could you explain what's going on here?  What does a wildCard have to do with TransferCompleted or BytesOverWire?  Is it based on the fact that there's multiple files being sent?  But shouldn't we still care about how many bytes gets transferred?

src/commands/azCopy/azCopyTransfer.ts
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    return jobId;
+}
+
+function getAzCopyExePath(): string {
+    const nodeModulesPath: string = join(getResourcesPath(), 'azCopy', 'node_modules', '@azure-tools');
+    if (platform() === "win32") {
+        return (process.arch.toLowerCase() === "x64" || process.env.hasOwnProperty("PROCESSOR_ARCHITEW6432")) ?
+            join(nodeModulesPath, 'azcopy-win64', 'dist', 'bin', 'azcopy_windows_amd64.exe') :
```
nturinski, line 63: I think it makes sense to make dist/bin constants.  Could probably make more things constants, but those seem like the most obvious ones.

src/tree/AttachedStorageAccountTreeItem.ts
```diff
@@ -117,6 +117,10 @@ class AttachedStorageRoot extends AttachedAccountRoot {
         throw new Error(localize('cannotRetrieveStorageAccountIdForAttachedAccount', 'Cannot retrieve storage account id for an attached account.'));
     }
 
+    public generateSasToken(): string {
```
nturinski, line 120: I was thinking, but is this true?  Can't you attach Azure Storage Accounts? Like, if you gave me the connection string to your account, technically I should be able to use AzCopy, right?

src/tree/blob/BlobContainerTreeItem.ts
```diff
@@ -3,20 +3,21 @@
...
 import { ProgressLocation, Uri } from 'vscode';
 import { AzExtTreeItem, AzureParentTreeItem, AzureTreeItem, DialogResponses, GenericTreeItem, IActionContext, ICreateChildImplContext, parseError, TelemetryProperties, UserCancelledError } from 'vscode-azureextensionui';
 import { AzureStorageFS } from '../../AzureStorageFS';
+import { createAzCopyDestination, createAzCopyLocalSource } from '../../commands/azCopy/azCopyLocations';
+import { azCopyBlobTransfer } from '../../commands/azCopy/azCopyTransfer';
 import { IExistingFileContext } from '../../commands/uploadFile';
 import { getResourcesPath, staticWebsiteContainerName } from "../../constants";
 import { ext } from "../../extensionVariables";
 import { TransferProgress } from '../../TransferProgress';
-import { createBlobContainerClient, createBlockBlobClient, createChildAsNewBlockBlob, IBlobContainerCreateChildContext, loadMoreBlobChildren } from '../../utils/blobUtils';
```
nturinski, line 19: You can move `createBlockBlobClient` out of `blobUtils.ts` now since it's only being used in `BlobTreeItem.ts`

src/tree/blob/BlobContainerTreeItem.ts
```diff
@@ -306,26 +307,19 @@ export class BlobContainerTreeItem extends AzureParentTreeItem<IStorageRoot> imp
     }
 
     public async uploadLocalFile(filePath: string, blobPath: string, suppressLogs: boolean = false): Promise<void> {
-        const blobFriendlyPath: string = `${this.friendlyContainerName}/${blobPath}`;
-        const blockBlobClient: azureStorageBlob.BlockBlobClient = createBlockBlobClient(this.root, this.container.name, blobPath);
-
         // tslint:disable-next-line: strict-boolean-expressions
         const totalBytes: number = (await fse.stat(filePath)).size || 1;
+        const transferProgress: TransferProgress = new TransferProgress(totalBytes, blobPath);
```
nturinski, line 312: nit: Doesn't seem like there's any need to move this.  Would prefer it be closer to where it's being used (L322)

webpack.config.js
```diff
@@ -18,7 +18,17 @@ let DEBUG_WEBPACK = !!process.env.DEBUG_WEBPACK;
 let config = dev.getDefaultWebpackConfig({
     projectRoot: __dirname,
     verbosity: DEBUG_WEBPACK ? 'debug' : 'normal',
-    externals: { './getCoreNodeModule': 'commonjs getCoreNodeModule' },
+    externals: {
+        './getCoreNodeModule': 'commonjs getCoreNodeModule',
+
+        // @azure-tools/azcopy-node
```
nturinski, line 24: Why is this commented out?

src/tree/AttachedStorageAccountTreeItem.ts
wwlorey: I believe storage explorer has SAS tokens working for attached accounts. But I didn't look into this yet to keep the PR smaller.

Do you think it's worth digging into now or should I create an issue to track it?

src/tree/blob/BlobContainerTreeItem.ts
wwlorey: It looks like `createBlockBlobClient` is also used in `blobUtils.ts` right now. Is ok to keep it in the utils file?

webpack.config.js
wwlorey: I originally put this here so that it was clear that these externals were related to @azure-tools/azcopy-node. But since the externals already have 'azcopy' in the title I'll remove this lol

src/commands/azCopy/azCopyTransfer.ts
wwlorey: This line is related to multiple file transfers which I forgot to remove. I'll reintroduce it when I submit that PR

src/tree/blob/BlobContainerTreeItem.ts
nturinski: Oh whoops, didn't notice that.  Yeah, keep it!

src/commands/azCopy/azCopyTransfer.ts
wwlorey: I added `getOsSpecificAzCopyExePath` to get rid of those repeated `join` calls. Hopefully that didn't complicate things more 😅

src/tree/AttachedStorageAccountTreeItem.ts
nturinski: Well, the only problem is if we remove our old method of upload/download, this effectively breaks Attached Accounts.  Let's make this another issue, but make sure that it's completed before release.

src/commands/azCopy/azCopyTransfer.ts
```diff
@@ -0,0 +1,92 @@
...
+
+async function azCopyTransfer(
+    src: ILocalLocation,
+    dst: IRemoteSasLocation,
+    transferProgress: TransferProgress,
+    fromTo: FromToOption,
+): Promise<void> {
+    const exe: string = getAzCopyExePath();
+    const copyClient: AzCopyClient = new AzCopyClient({ exe });
+    const copyOptions: ICopyOptions = { fromTo, overwriteExisting: "true", recursive: true, followSymLinks: true };
```
nturinski, line 31: I noticed that we're overwriting.  Is there a confirmation dialog considering this is a destructive action?

nturinski:
So I installed from .vsix to do some spot checking and I haven't been able to upload a blob... (or have I?)

image

Any idea?

wwlorey:
> So I installed from .vsix to do some spot checking and I haven't been able to upload a blob... (or have I?)

My bad not testing with the vsix 🥵 I specified that the exe packages shouldn't be webpacked and updated `getAzCopyExePath` to also check the "dist" folder for node modules (that's where they end up after webpacking)

I also fixed the success message after an upload failure. I forgot to handle errors that happen in `callWithTelemetryAndErrorHandling`

.azure-pipelines/common/build.yml
```diff
@@ -1,4 +1,9 @@
 steps:
+- script: |
+    npm install --force
+    sudo chmod u+x node_modules/@azure-tools/azcopy-darwin/dist/bin/azcopy_darwin_amd64 \
```
EricJizbaMSFT, line 4: I don't think you should need chmod. We have exes in other node modules (for example, "opn") where the executable bit already exists on the exe. Only caveat is that we can't use copy-webpack-plugin for these (but I don't think you are): https://github.com/microsoft/vscode-azuretools/pull/403

.gitignore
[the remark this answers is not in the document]
EricJizbaMSFT: Seems like you can revert these changes now that you're not committing these modules to git?

src/commands/azCopy/azCopyTransfer.ts
```diff
@@ -0,0 +1,118 @@
...
+        // tslint:disable-next-line: strict-boolean-expressions
+        finishedWork = status?.BytesOverWire || 0;
+        transferProgress.reportToOutputWindow(finishedWork);
+        await delay(1000);
+    }
+
+    return jobId;
+}
+
+async function getAzCopyExePath(platform: NodeJS.Platform): Promise<string> {
```
EricJizbaMSFT, line 85: We don't _have_ to pass in a path to azcopy. If the packages like "@azure-tools/azcopy-darwin" are in the typical "node_modules" location, I think "@azure-tools/azcopy-node" is supposed to find the appropriate one. Can you try not specifying a path to azcopy?

.azure-pipelines/common/build.yml
wwlorey: According to Matt, @azure-tools/azcopy-node requires the exe packages to be executable. I originally tried not running chmod (and retried with the most recent changes) and AzCopy fails to transfer

src/commands/azCopy/azCopyTransfer.ts
wwlorey: You're right - I originally passed in a path when we were going the resources folder route. But it works without getAzCopyExePath now that we're npm installing the exe packages

package.json
```diff
@@ -767,6 +767,7 @@
     },
     "scripts": {
         "vscode:prepublish": "npm run webpack-prod",
+        "postinstall": "scripts/postinstall.sh",
```
wwlorey, line 770: Since we're not listing the exe packages as dependencies and force installing them on the build server, I added this script to force install each exe package when you run `npm install` locally. Just to make the dev process run smoother. But this causes the [windows build to fail](https://dev.azure.com/ms-azuretools/AzCode/_build/results?buildId=24472&view=logs&j=2d2b3007-3c5c-5840-9bb0-2b1ea49925f3&t=0382ecff-9843-5774-f6e6-db1050292abf). Do you guys have any ideas on how to improve this?

wwlorey: Per offline discussion: We don't need to force install all the exe packages for local debugging.

src/commands/azCopy/azCopyTransfer.ts
```diff
@@ -0,0 +1,79 @@
...
+    let finalTransferStatus = (await copyClient.getJobInfo(jobId)).latestStatus;
+    context.telemetry.properties.jobStatus = finalTransferStatus?.JobStatus;
+    if (!finalTransferStatus || finalTransferStatus.JobStatus !== 'Completed') {
+        // tslint:disable-next-line: strict-boolean-expressions
+        let message: string = localize('azCopyTransfer', 'AzCopy Transfer: "{0}".', finalTransferStatus?.JobStatus || 'Failed');
+        if (finalTransferStatus?.FailedTransfers || finalTransferStatus?.SkippedTransfers) {
+            message += localize('checkOutputWindow', ' Check the output window for a list of incomplete transfers.');
+
+            if (finalTransferStatus.FailedTransfers) {
+                ext.outputChannel.appendLog(localize('failedTransfers', 'Failed transfers: {0}', finalTransferStatus.FailedTransfers.join(', ')));
```
nturinski, line 43: Have you happened to have a chance to see what this looks like?  I'm imagining that this is the entire file path in some cases, so it might be better to append a newline character to each file.

src/commands/azCopy/azCopyTransfer.ts
```diff
@@ -0,0 +1,92 @@
...
+    const copyClient: AzCopyClient = new AzCopyClient({ exe });
+    const copyOptions: ICopyOptions = { fromTo, overwriteExisting: "true", recursive: true, followSymLinks: true };
+    let jobId: string = await startAndWaitForCopy(copyClient, src, dst, copyOptions, transferProgress);
+    let finalTransferStatus = (await copyClient.getJobInfo(jobId)).latestStatus;
+    if (!finalTransferStatus || finalTransferStatus.JobStatus !== 'Completed') {
+        // tslint:disable-next-line: strict-boolean-expressions
+        let message: string = localize('azCopyTransfer', `AzCopy Transfer: "${finalTransferStatus?.JobStatus || 'Failed'}".`);
+        message += finalTransferStatus?.ErrorMsg ? ` ${finalTransferStatus.ErrorMsg}` : '';
+
+        if (finalTransferStatus?.JobStatus === 'CompletedWithSkipped') {
```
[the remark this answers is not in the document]
nturinski: I think we should do a regex that's like CompletedWith... and throw a warning for anything that's not just "Completed."

src/commands/azCopy/azCopyTransfer.ts
wwlorey: Here's what it's looking like now:
Screen Shot 2020-08-27 at 4 45 39 PM

I also changed it to display the destination path instead of the entire object which wouldn't display properly

wwlorey:
I noticed that `chmod` in the "postinstall" script would throw an error because the linux exe package doesn't exist (when run on mac). So I changed the script to swallow chmod errors

nturinski:
Rather than swallowing the error, why don't you make a specific task for Linux and Mac?

https://github.com/microsoft/vscode-azurestorage/blob/master/.azure-pipelines/main.yml

wwlorey:
> Rather than swallowing the error, why don't you make a specific task for Linux and Mac?
> 
> https://github.com/microsoft/vscode-azurestorage/blob/master/.azure-pipelines/main.yml

Do you know if there's a way to do platform specific commands in the package.json scripts? We only run into the problem here: https://github.com/microsoft/vscode-azurestorage/pull/729/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R770

nturinski:
So you should be able to define which build scripts to run in yml files.  For example, look at the build.yml, it has the commands it runs.  So you can run the npm script to run the installs that way.

EricJizbaMSFT:
@nturinski you've got it covered - @wwlorey feel free to merge

Issue #719, opened 2020-07-27
[Suggestion] It is better to change the error message to "File name cannot be empty"


v-mengwe:
**OS:** Win10
**Build Version:** 20200726.1

**Repro Steps:**

1. Drill down a Storage account --> File Shares.
2. Right click a File Share --> Select "Create Empty File..." --> Press "Enter".
3. Check the error message.

**Actual:**
The error message shows "Filename cannot be empty", miss a space for "Filename".
![image](https://user-images.githubusercontent.com/62053665/88520951-de971780-d026-11ea-9bd5-f37fe20c9496.png)

**Suggestion:**
It is better to change the error message to "File name cannot be empty", like the below screens:
![image](https://user-images.githubusercontent.com/62053665/88521719-e0150f80-d027-11ea-8ae3-f76fabb6a5f4.png)

PR #731, opened 2020-08-03
Rename 'Filename' to 'File name' in prompts

wwlorey:
Fixes https://github.com/microsoft/vscode-azurestorage/issues/719


wwlorey:
Yeah I agree. I like this change just because it keeps the wording consistent with other resource types

Issue #720, opened 2020-07-27
Add support for Azure Stack


yatna:
**Current Status** - Storage accounts within subscriptions don't show up.

**Approach** - The problem arises mainly because Azure Stack doesn't support the same API version of azure-arm-storage and azure storage-blob We plan to dynamically load the correct library versions depending on environment. Here the steps -

1. Install all required versions
```
"azure-arm-storage": "npm:azure-arm-storage@^5.1.1-preview",
"azure-arm-storage3": "npm:azure-arm-storage@^3.0.0",
```
2. Determine the environment by looking at the settings.json
3. Use the correct version of library dynamically using if...else... in all files.

**Here is the working PR for reference - https://github.com/microsoft/vscode-azurestorage/pull/699**

**Maintenance** - We only load a different library when the environment is Azure Stack, so functionality for Azure won't be affected in any way. If the version AzureStack supports changes, we only need to update the package.json file.

alexweininger:
Done in https://github.com/microsoft/vscode-azurestorage/pull/905

PR #732, opened 2020-08-06
[Work In Progress] Added changes to work against AzureStack Environment

yatna:
In reference to #720 

We are waiting for @azure/arm-storage package to be published for 2017-10-01 API version, hence WIP.

### DIFF
--- a/package.json
+++ b/package.json
@@ -805,7 +805,7 @@
     },
     "dependencies": {
         "@azure/arm-resources": "^3.0.0",
-        "@azure/arm-storage1": "https://sdkautomation.blob.core.windows.net/sdkautomation-pipeline/Azure/azure-rest-api-specs/10334/azure-sdk-for-js/Azure/azure-sdk-for-js/@azure/arm-storage/azure-arm-storage-15.0.0.tgz",
+        "@azure/arm-storage-profile-2019-03-01-hybrid": "^1.0.0",
         "@azure/arm-storage": "npm:@azure/arm-storage@^15.0.0",
         "@azure/storage-blob": "12.0.2",
         "@azure/storage-file-share": "^12.1.1",
--- a/package.json
+++ b/package.json
@@ -805,8 +805,9 @@
     },
     "dependencies": {
         "@azure/arm-resources": "^3.0.0",
-        "@azure/arm-storage": "^15.0.0",
-        "@azure/storage-blob": "^12.1.1",
+        "@azure/arm-storage1": "https://sdkautomation.blob.core.windows.net/sdkautomation-pipeline/Azure/azure-rest-api-specs/10334/azure-sdk-for-js/Azure/azure-sdk-for-js/@azure/arm-storage/azure-arm-storage-15.0.0.tgz",
+        "@azure/arm-storage": "npm:@azure/arm-storage@^15.0.0",
+        "@azure/storage-blob": "12.0.2",
         "@azure/storage-file-share": "^12.1.1",
         "@types/mime": "^2.0.1",
         "azure-storage": "^2.10.0",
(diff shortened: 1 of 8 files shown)


src/utils/environmentUtils.ts
```diff
@@ -0,0 +1,88 @@
...
+    if (environmentType === "unknown") {
+        const settingsPath = path.join(<string>process.env.APPDATA, 'Code - Insiders', 'User', 'settings.json');
+        let isStack: string = "";
+        try {
+            // tslint:disable-next-line: non-literal-require
+            const data = require(settingsPath);
+            // tslint:disable-next-line: no-unsafe-any
+            baseUrl = data["azure.ppe"].resourceManagerEndpointUrl;
+            // tslint:disable-next-line: no-unsafe-any
+            isStack = data["azure.target_azurestack_api_version"];
```
bganapa, line 28: For now this is ok, This should ideally be target_api_profile:'2019-03-01-hybrid'..
Since AzureStack is a boxed product, there will be a case where one customer environment is with 2017-10-01 api version and another one is with 2019-02-02.

This matters only when we consume features that are specific to a api version in the extension. Rightnow it is being used to list the accounts and it doesnt matter which api version we use for that.

package.json
```diff
@@ -805,8 +805,9 @@
     },
     "dependencies": {
         "@azure/arm-resources": "^3.0.0",
-        "@azure/arm-storage": "^15.0.0",
-        "@azure/storage-blob": "^12.1.1",
+        "@azure/arm-storage1": "https://sdkautomation.blob.core.windows.net/sdkautomation-pipeline/Azure/azure-rest-api-specs/10334/azure-sdk-for-js/Azure/azure-sdk-for-js/@azure/arm-storage/azure-arm-storage-15.0.0.tgz",
+        "@azure/arm-storage": "npm:@azure/arm-storage@^15.0.0",
+        "@azure/storage-blob": "12.0.2",
```
bganapa, line 810: pls note we are going back to a older version of the package here .. @EricJizbaMSFT

package.json
```diff
@@ -805,8 +805,9 @@
     },
     "dependencies": {
         "@azure/arm-resources": "^3.0.0",
-        "@azure/arm-storage": "^15.0.0",
-        "@azure/storage-blob": "^12.1.1",
+        "@azure/arm-storage1": "https://sdkautomation.blob.core.windows.net/sdkautomation-pipeline/Azure/azure-rest-api-specs/10334/azure-sdk-for-js/Azure/azure-sdk-for-js/@azure/arm-storage/azure-arm-storage-15.0.0.tgz",
```
bganapa, line 808: This is a private package. We are working with the azure sdk team to release a package for us, likely to be versioned as 0.0.5

bganapa:
@EricJizbaMSFT Could you please review? Other than the private package, do you have any concern? Do you think we can consume the 2017-10-01 storage package for azure as well?

EricJizbaMSFT:
I don't normally review "Work in Progress" PRs. I did a quick pass and noticed a few themes. Once the PR is finished, I can do a more thorough review.
- Don't duplicate code - instead pull it out into it's own function
- Don't use `any` - instead use a proper type
- We expect tslint rules to be followed the vast majority of the time, not disabled. Google the rule and you should be able to find docs on why it exists and tips on how to fix

Also - to build on what I said [here](https://github.com/microsoft/vscode-azurefunctions/issues/2259), I don't want us to duplicate all of this azure stack code in every extension we own and would much prefer if we came up with a generic solution. For example, the code could be a part of [`createAzureClient`](https://github.com/microsoft/vscode-azuretools/blob/master/ui/src/createAzureClient.ts) inside our "vscode-azureextensionui" npm package.

EricJizbaMSFT:
Closing due to lack of activity. I think we discussed going a different route with this work offline anyways

bganapa:
Thanks for closing. yes. We will revisit this work post ignite

Issue #741, opened 2020-08-21
Investigate generating SAS tokens for attached accounts


wwlorey:
From https://github.com/microsoft/vscode-azurestorage/pull/729#discussion_r474358614

PR #752, opened 2020-08-28
Implement 'generateSasToken' for attached accounts

wwlorey:
Fixes https://github.com/microsoft/vscode-azurestorage/issues/741


src/tree/AttachedStorageAccountTreeItem.ts
```diff
@@ -117,8 +119,21 @@ class AttachedStorageRoot extends AttachedAccountRoot {
         throw new Error(localize('cannotRetrieveStorageAccountIdForAttachedAccount', 'Cannot retrieve storage account id for an attached account.'));
     }
 
-    public generateSasToken(): string {
-        throw new Error(localize('cantGenerateSasToken', 'Cannot generate SAS token for an attached account.'));
+    public generateSasToken(expiresOn: Date, permissions: string, services: string, resourceTypes: string): string {
```
nturinski, line 122: This code is almost duplicated from StorageAccountTreeItem.  Could we share the code somehow instead?

wwlorey: I changed the `generateSasToken` parameters to just `accountSASSignatureValues` so the only thing that's shared now is calling `generateAccountSASQueryParameters`

src/tree/getPropertyFromConnectionString.ts
```diff
@@ -0,0 +1,11 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.md in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+export function getPropertyFromConnectionString(connectionString: string, property: string): string | undefined {
```
nturinski, line 6: Seems like this should be in the utils folder.

PR #753, opened 2020-08-28
Support AzCopy single file uploads

wwlorey:


src/tree/fileShare/FileShareTreeItem.ts
```diff
@@ -136,17 +137,11 @@ export class FileShareTreeItem extends AzureParentTreeItem<IStorageRoot> impleme
         }
 
         const fileSize: number = (await fse.stat(sourceFilePath)).size;
-        const fileClient: azureStorageShare.ShareFileClient = createFileClient(this.root, this.shareName, parentDirectoryPath, path.basename(destFilePath));
         // tslint:disable-next-line: strict-boolean-expressions
         const transferProgress: TransferProgress = new TransferProgress(fileSize || 1, destFilePath);
-        const options: azureStorageShare.FileParallelUploadOptions = {
-            fileHttpHeaders: {
-                // tslint:disable-next-line: strict-boolean-expressions
-                fileContentType: mime.getType(destFilePath) || undefined
```
wwlorey, line 145: As of right now, @azure-tools/azcopy-node doesn't support the `--content-type` setting which would replace this. I reached out to Matt to see if he can add it for us

nturinski: Do you know how not having the content type will affect the file upload?

src/commands/azCopy/azCopyTransfer.ts
```diff
@@ -21,6 +21,16 @@ export async function azCopyBlobTransfer(
     });
 }
 
+export async function azCopyFileTransfer(
+    src: ILocalLocation,
+    dst: IRemoteSasLocation,
+    transferProgress: TransferProgress,
+): Promise<void> {
+    await callWithTelemetryAndErrorHandling('azCopyFileTransfer', async (context: IActionContext) => {
```
nturinski, line 29: Is there any reason you made a new event instead of just passing in the context?  I believe all uploadFile events are already registered.

src/tree/fileShare/FileShareTreeItem.ts
wwlorey: After looking into this more, AzCopy will actually guess the mime type if we don't specify `--content-type`. So this should have the same functionality as it did before

src/commands/azCopy/azCopyTransfer.ts
wwlorey: I refactored so context is passed down to the AzCopy transfer

src/tree/blob/BlobContainerTreeItem.ts
```diff
@@ -201,6 +201,7 @@ export class BlobContainerTreeItem extends AzureParentTreeItem<IStorageRoot> imp
      */
     // tslint:disable-next-line: max-func-body-length
     private async deployStaticWebsiteCore(
+        context: IActionContext,
         properties: TelemetryProperties,
```
nturinski, line 205: You can remove the properties parameter since context will always have the properties property.

PR #754, opened 2020-09-01
Support AzCopy folder uploads

wwlorey:


src/commands/azCopy/azCopyTransfer.ts
```diff
@@ -3,43 +3,31 @@
...
+import { CancellationToken, Progress } from 'vscode';
 import { IActionContext } from "vscode-azureextensionui";
 import { ext } from '../../extensionVariables';
 import { TransferProgress } from "../../TransferProgress";
 import { delay } from "../../utils/delay";
+import { throwIfCanceled } from "../../utils/errorUtils";
 import { localize } from "../../utils/localize";
 
-export async function azCopyBlobTransfer(
+export async function azCopyTransfer(
```
wwlorey, line 15: The resource-specific transfer functions were getting a little ugly with all the repeated code so I refactored

src/commands/azCopy/azCopyTransfer.ts
```diff
@@ -3,43 +3,31 @@
...
     context: IActionContext,
-    src: ILocalLocation,
+    transferType: 'LocalBlob' | 'LocalFile',
```
nturinski, line 17: Probably worth just making this a type.  You reference it enough times throughout the code, you'll need it for download, and it'll make things easier if we ever plan to support tables/queues down the road.

src/commands/azCopy/azCopyLocations.ts
```diff
@@ -17,6 +18,13 @@ export function createAzCopyLocalSource(sourcePath: string): ILocalLocation {
     return { type: "Local", path: sourcePath, useWildCard: false };
 }
 
+export function createAzCopyLocalDirectorySource(path: string): ILocalLocation & { isDirectory?: boolean } {
```
nturinski, line 21: Doesn't seem like `isDirectory` should be optional in this case.

src/commands/azCopy/azCopyLocations.ts
```diff
@@ -17,6 +18,13 @@ export function createAzCopyLocalSource(sourcePath: string): ILocalLocation {
...
+    if (!path.endsWith(sep)) {
+        path += sep;
+    }
+    return { type: "Local", path, useWildCard: true, isDirectory: true };
```
nturinski, line 25: So I thought we were supposed to use `useWildCard` for directories.  It doesn't seem like `isDirectory` is a property that natively exists on the `ILocalLocation` typing either.

src/commands/azCopy/azCopyTransfer.ts
```diff
@@ -3,43 +3,31 @@
...
-    context: IActionContext,
-    src: ILocalLocation,
-    dst: IRemoteSasLocation,
-    transferProgress: TransferProgress,
-    fromTo: FromToOption,
-): Promise<void> {
     const copyClient: AzCopyClient = new AzCopyClient();
-    const copyOptions: ICopyOptions = { fromTo, overwriteExisting: "true", recursive: true, followSymLinks: true };
-    let jobId: string = await startAndWaitForCopy(copyClient, src, dst, copyOptions, transferProgress);
+    const copyOptions: ICopyOptions = { fromTo: transferType, overwriteExisting: "true", recursive: true, followSymLinks: true, excludePath: '.git;.vscode' };
```
nturinski, line 29: Might be out of scope of this PR, but is it possible to prompt users for each item that they are overwriting and give them the following options:

`Do you wish to overwrite "filename"? Yes, Yes to All, No, No to All?`

src/commands/azCopy/azCopyLocations.ts
wwlorey: You right - I added `isDirectory` specifically to make [this check](https://github.com/microsoft/vscode-azurestorage/blob/4eac7659d651df11526ec0a3bc80e5bd3636d92e/src/commands/azCopy/azCopyTransfer.ts#L82) more readable. It seemed better than checking against `useWildCard`. But if just checking `useWildCard` seems fine, I can do that instead

src/commands/azCopy/azCopyTransfer.ts
wwlorey: For the sake of getting core code done, I created an issue to track this: https://github.com/microsoft/vscode-azurestorage/issues/757

src/commands/azCopy/azCopyLocations.ts
nturinski: I'm fine with `isDirectory` but you weren't using it, as far as I can tell.  I think I'd be okay with just using `useWildCard` and maybe have a comment or something saying that this indicates that it's a folder or something.

src/commands/azCopy/azCopyLocations.ts
```diff
@@ -6,15 +6,25 @@
...
 import { AccountSASPermissions, AccountSASSignatureValues, ContainerClient } from "@azure/storage-blob";
 import { ShareClient } from "@azure/storage-file-share";
+import { sep } from "path";
 import { BlobContainerTreeItem } from "../../tree/blob/BlobContainerTreeItem";
 import { FileShareTreeItem } from "../../tree/fileShare/FileShareTreeItem";
 import { createBlobContainerClient } from "../../utils/blobUtils";
 import { createShareClient } from "../../utils/fileUtils";
 
 const threeDaysInMS: number = 1000 * 60 * 60 * 24 * 3;
+let localLocationType: 'Local';
```
nturinski, line 16: nit: `const`

src/commands/azCopy/azCopyLocations.ts
```diff
@@ -6,15 +6,25 @@
...
+let remoteSasLocationType: 'RemoteSas';
```
nturinski, line 17: Don't need this one yet though 😅

src/commands/azCopy/azCopyLocations.ts
wwlorey: Went ahead and removed `isDirectory`

src/commands/azCopy/azCopyLocations.ts
```diff
@@ -6,15 +6,25 @@
...
 
 export function createAzCopyLocalSource(sourcePath: string): ILocalLocation {
-    return { type: "Local", path: sourcePath, useWildCard: false };
+    return { type: localLocationType, path: sourcePath, useWildCard: false };
+}
+
+export function createAzCopyLocalDirectorySource(path: string): ILocalLocation & { isDirectory: true } {
```
nturinski: I think we could just make this a custom type that we wrap `ILocalLocation` around.  You do reference it a few times at least.

src/commands/azCopy/azCopyLocations.ts
```diff
@@ -6,15 +6,25 @@
...
 const threeDaysInMS: number = 1000 * 60 * 60 * 24 * 3;
+let localLocationType: 'Local';
+let remoteSasLocationType: 'RemoteSas';
 
 export function createAzCopyLocalSource(sourcePath: string): ILocalLocation {
-    return { type: "Local", path: sourcePath, useWildCard: false };
+    return { type: localLocationType, path: sourcePath, useWildCard: false };
+}
+
+export function createAzCopyLocalDirectorySource(path: string): ILocalLocation {
```
nturinski, line 23: nit: It's kind of a small thing, but you could actually just call `createAzCopyLocalSource` here instead and set `useWildCard` by a boolean parameter called `isFolder`.  Kind of best of both worlds in this case (for clarity)

In which case, you wouldn't need the localLocationType const either.

src/commands/azCopy/azCopyTransfer.ts
```diff
@@ -3,43 +3,33 @@
...
-export async function azCopyBlobTransfer(
-    context: IActionContext,
-    src: ILocalLocation,
-    dst: IRemoteSasLocation,
-    transferProgress: TransferProgress,
-): Promise<void> {
-    context.errorHandling.rethrow = true;
-    await azCopyTransfer(context, src, dst, transferProgress, 'LocalBlob');
-}
+export type AzCopyTransferType = 'LocalBlob' | 'LocalFile';
```
nturinski, line 15: So this just offers type safety (which we already had from `FromToOption`).  In that case, we should just go ahead and use the `FromToOption` type.

I was thinking of maybe using an enum like this: https://github.com/microsoft/vscode-azuretools/blob/5999c2ad4423e86f22d2c648027242d8816a50e4/appservice/src/createAppService/AppKind.ts#L6-L9

so that you wouldn't have to keep typing 'LocalBlob', but I'll leave that up to you.

src/commands/azCopy/azCopyTransfer.ts
```diff
@@ -3,43 +3,33 @@
...
-    context: IActionContext,
-    src: ILocalLocation,
-    dst: IRemoteSasLocation,
-    transferProgress: TransferProgress,
-    fromTo: FromToOption,
-): Promise<void> {
     const copyClient: AzCopyClient = new AzCopyClient();
-    const copyOptions: ICopyOptions = { fromTo, overwriteExisting: "true", recursive: true, followSymLinks: true };
-    let jobId: string = await startAndWaitForCopy(copyClient, src, dst, copyOptions, transferProgress);
+    const copyOptions: ICopyOptions = { fromTo: transferType, overwriteExisting: "true", recursive: true, followSymLinks: true, excludePath: '.git;.vscode' };
```
nturinski, line 31: This should always the same as this filter right?  https://github.com/microsoft/vscode-azurestorage/pull/754/files#diff-cdda8e0f37050d0b16f295984560111eR21

If so, I'd prefer you make some sort of constant so we don't ever forget to include more folders in the future.

src/commands/azCopy/azCopyTransfer.ts
```diff
@@ -73,20 +63,36 @@ async function azCopyTransfer(
...
     let jobId: string = await copyClient.copy(src, dst, options);
     let status: TransferStatus | undefined;
     let finishedWork: number;
     while (!status || status.StatusType !== 'EndOfJob') {
+        throwIfCanceled(cancellationToken, context.telemetry.properties, 'startAndWaitForCopy');
         status = (await copyClient.getJobInfo(jobId)).latestStatus;
-        // tslint:disable-next-line: strict-boolean-expressions
-        finishedWork = status?.BytesOverWire || 0;
+        if (src.useWildCard) {
+            // Directory transfers always have `useWildCard` set
```
nturinski, line 85: nit: Can you move the comment above the if statement?  I dunno why, but I always expect the comment to refer to the code below it.

src/commands/uploadToAzureStorage.ts
```diff
@@ -38,26 +40,30 @@ export async function uploadToAzureStorage(actionContext: IActionContext, target
...
-        const transferProgress = new TransferProgress(filePathsToUpload.length);
-        await uploadFiles(actionContext, treeItem, parentDirectory, '', filePathsToUpload, actionContext.telemetry.properties, transferProgress, notificationProgress, cancellationToken);
+    await vscode.window.withProgress({ cancellable: true, location: vscode.ProgressLocation.Notification, title: `Uploading to ${destinationName} from ${resourcePath}` }, async (notificationProgress, cancellationToken) => {
+        if ((await fse.stat(resourcePath)).isDirectory()) {
+            const message: string = localize('uploadWillOverwrite', 'Uploading "{0}" will overwrite any existing resources with the same name.', resourcePath);
+            await ext.ui.showWarningMessage(message, { modal: true }, { title: localize('upload', 'Upload') });
+
+            // AzCopy recognizes folders as a resource when uploading to file shares. So only count folders in that case
+            const numResources: number = await getNumResourcesInDirectory(resourcePath, treeItem instanceof FileShareTreeItem);
```
nturinski, line 59: This is used the exact same way in `BlobContainerTreeItem`.  Seems like you could actually move it into `uploadFiles` instead (and then get rid of `getNumResources` from the utils file).

src/commands/uploadToAzureStorage.ts
```diff
@@ -38,26 +40,30 @@ export async function uploadToAzureStorage(actionContext: IActionContext, target
...
+            const transferProgress = new TransferProgress(numResources);
+            await uploadFiles(actionContext, treeItem, resourcePath, destinationPath, transferProgress, notificationProgress, cancellationToken);
+        } else {
+            if (treeItem instanceof BlobContainerTreeItem ? await doesBlobExist(treeItem, destinationPath) : await doesFileExist(path.basename(destinationPath), treeItem, path.dirname(destinationPath), treeItem.shareName)) {
```
nturinski, line 63: I think you could consolidate this logic since it exists in [uploadFile](https://github.com/microsoft/vscode-azurestorage/blob/master/src/commands/uploadFile.ts#L47),  Or, both call `uploadLocalFile` so you could also put the check in the respective tree item so you don't even need this ternary statement.

src/commands/azCopy/azCopyTransfer.ts
nturinski: Yeah, I think it's probably not worth doing that then.  Though where/how did you decide on these files/folders to ignore? I figure users wouldn't want to upload other things, such as `.env` files that could contain environment variables with secrets, etc.

src/commands/azCopy/azCopyTransfer.ts
```diff
@@ -73,20 +61,37 @@ async function azCopyTransfer(
...
     while (!status || status.StatusType !== 'EndOfJob') {
+        throwIfCanceled(cancellationToken, context.telemetry.properties, 'startAndWaitForCopy');
         status = (await copyClient.getJobInfo(jobId)).latestStatus;
-        // tslint:disable-next-line: strict-boolean-expressions
-        finishedWork = status?.BytesOverWire || 0;
+
+        // Directory transfers always have `useWildCard` set
+        if (src.useWildCard) {
+            // tslint:disable: strict-boolean-expressions
+            finishedWork = status?.TransfersCompleted || 0;
```
nturinski, line 86: Would be a little cleaner to do this `finishedWork = ( src.useWildCard ? status?.TransfersCompleted : status.BytesOverWrite ) || 0;`

src/commands/uploadToAzureStorage.ts
```diff
@@ -38,26 +38,25 @@ export async function uploadToAzureStorage(actionContext: IActionContext, target
...
-    if ((await fse.stat(resourcePath)).isDirectory()) {
-        filePathsToUpload = await listFilePathsWithAzureSeparator(resourcePath);
-        parentDirectory = resourcePath;
-    } else {
-        filePathsToUpload = [resourcePath];
-        parentDirectory = path.dirname(resourcePath);
-    }
-
+    let destinationPath: string = path.basename(resourcePath);
     if (treeItem instanceof BlobContainerTreeItem) {
```
nturinski, line 43: So since you're checking if the treeItem is a `BlobContainerTreeItem` in `uploadFiles`, I think you could just pass the `resourcePath` and set the `destinationPath` when you set the `transferType`.  Seems like it's easier to manage setting blob/file specific things all in one spot.

As for destinationName, I think I'd just make the treeItems have a `get name()` function that returns what we expect as the name.  While the label is actually technically the same thing as the name, it'd probably be better to be explicit about it since we could decide to change the label at some point.

After that, you can change any `treeItem.container.name` or `treeItem.shareName` to all just be `treeItem.name`

src/commands/azCopy/azCopyTransfer.ts
wwlorey: A while ago we were ignoring all folders/files starting with `.` and we narrowed it down to just .git and .vscode based on some user filed issues. There are probably more folders that would be worth ignoring though. Should I file an issue to track that?

src/commands/uploadToAzureStorage.ts
wwlorey: After refactoring, the only time we need the container/share name in `uploadToAzureStorage.ts` is in a notification title. So I just used `treeItem.label` there since we don't need it to be the exact name.

I also made the `destPath` parameter in `uploadLocalFile` and `uploadFiles` optional for other cases when there is a set destination path we want to use

src/commands/azCopy/azCopyTransfer.ts
nturinski: I don't know if we need an issue-- maybe a quick google search to see if there is an official list of folders to ignore.  I think GitHub usually has things to include in the .gitignore, I'd assume it's safe to copy that.  @EricJizbaMSFT opinions?

wwlorey: I searched around and had trouble finding a concise list of what we should ignore

src/commands/uploadToAzureStorage.ts
```diff
@@ -38,26 +35,15 @@ export async function uploadToAzureStorage(actionContext: IActionContext, target
     }
 
     let treeItem: BlobContainerTreeItem | FileShareTreeItem = await ext.tree.showTreeItemPicker([BlobContainerTreeItem.contextValue, FileShareTreeItem.contextValue], actionContext);
-    let filePathsToUpload: string[];
-    let parentDirectory: string;
-    let destinationName: string;
+    await vscode.window.withProgress({ cancellable: true, location: vscode.ProgressLocation.Notification, title: `Uploading to ${treeItem.label} from ${resourcePath}` }, async (notificationProgress, cancellationToken) => {
```
nturinski, line 38: Might as well localize 😇

src/commands/uploadToAzureStorage.ts
```diff
@@ -38,26 +35,15 @@ export async function uploadToAzureStorage(actionContext: IActionContext, target
...
-    if (treeItem instanceof BlobContainerTreeItem) {
-        destinationName = treeItem.container.name;
-    } else {
-        destinationName = treeItem.shareName;
-    }
-
-    await vscode.window.withProgress({ cancellable: true, location: vscode.ProgressLocation.Notification, title: `Uploading to ${destinationName} from ${parentDirectory}` }, async (notificationProgress, cancellationToken) => {
-        const transferProgress = new TransferProgress(filePathsToUpload.length);
-        await uploadFiles(actionContext, treeItem, parentDirectory, '', filePathsToUpload, actionContext.telemetry.properties, transferProgress, notificationProgress, cancellationToken);
+            // AzCopy recognizes folders as a resource when uploading to file shares. So only count folders in that case
```
nturinski, line 43: I'm not sure, but should this comment be moved to `uploadFiles` right above `getNumResourcesInDirectory`?

nturinski (approved):
Nice! I think this PR is look pretty spiffy and cleaned up. 👍

src/commands/uploadToAzureStorage.ts
wwlorey: I think a comment still go right above the `uploadFiles` call since that's where we set `countFoldersAsResources`. I updated the comment so it hopefully makes more sense

src/commands/azCopy/azCopyTransfer.ts
EricJizbaMSFT: You're not gonna find an "official list" for something as generic as Azure Storage. IMO, we should copy _all_ folders by default, but provide settings for users to ignore folders if they want. I'm a little surprised we're ignoring ".git" and ".vscode", but perhaps that was mostly for the static site scenario (which is no longer a priority because of static web apps).

PR #759, opened 2020-09-02
Remove unneeded error handling & fix failed/skipped transfers listing

wwlorey:
We no longer need additional error handling when calling `azCopyTransfer`.

Also, `finalTransferStatus?.FailedTransfers` tends to be `[]` instead of null so not explicitly checking the length led to failed transfers being output when there were none.

Here's an example of errors for the same failed file transfer:

| Old  | New  |
|---|---|
| Screen Shot 2020-09-02 at 2 08 14 PM  | Screen Shot 2020-09-02 at 2 07 11 PM  |

### DIFF
--- a/src/commands/azCopy/azCopyTransfer.ts
+++ b/src/commands/azCopy/azCopyTransfer.ts
@@ -47,6 +47,9 @@ export async function azCopyTransfer(
                     ext.outputChannel.appendLog(`\t${skippedTransfer.Dst}`);
                 }
             }
+        } else {
+            // Add an additional error log since we don't have any more info about the failure
+            ext.outputChannel.appendLog(localize('couldNotUpload', 'Could not upload "{0}"', src.path));
         }
         message += finalTransferStatus?.ErrorMsg ? ` ${finalTransferStatus.ErrorMsg}` : '';
 
--- a/src/commands/azCopy/azCopyTransfer.ts
+++ b/src/commands/azCopy/azCopyTransfer.ts
@@ -24,7 +24,6 @@ export async function azCopyTransfer(
     }>,
     cancellationToken?: CancellationToken
 ): Promise<void> {
-    context.errorHandling.rethrow = true;
     const copyClient: AzCopyClient = new AzCopyClient();
     const copyOptions: ICopyOptions = { fromTo, overwriteExisting: "true", recursive: true, followSymLinks: true, excludePath: '.git;.vscode' };
     let jobId: string = await startAndWaitForCopy(context, copyClient, src, dst, copyOptions, transferProgress, notificationProgress, cancellationToken);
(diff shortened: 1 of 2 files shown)


nturinski:
Wait, I can't tell from your screenshots.  Is the upload not actually failing?  The new one seems like it has less information than the old one.

wwlorey:
The old failure showed the "Failed transfers:" header even though `finalTransferStatus?.FailedTransfers` was an empty list. So the new failure doesn't show that. Its possible for AzCopy to fail with an empty failed transfers list if the config is screwed up or something along those lines

And I removed "Could not upload file ..." from the output because if was originally in a try/catch we no longer need. Do you think that message is still helpful to have?

nturinski:
Are these two screenshots for the same exact scenario though?  I think the output regarding which file hasn't uploaded is useful.

wwlorey:
They are from the same scenario. I'll add that log back in and re-upload the screenshots to make it more clear

wwlorey:
Ok new screenshots coming your way:

| Old | New |
| --- | --- |
|Screen Shot 2020-09-02 at 2 08 14 PM  | Screen Shot 2020-09-02 at 3 59 25 PM |

The big differences between old and new are

* "Failed transfers:" isn't displayed when it doesn't need to be, and 
* an error notification appears when the error occurs

src/commands/azCopy/azCopyTransfer.ts
```diff
@@ -24,25 +24,26 @@ export async function azCopyTransfer(
...
         let message: string = localize('azCopyTransfer', 'AzCopy Transfer: "{0}".', finalTransferStatus?.JobStatus || 'Failed');
-        if (finalTransferStatus?.FailedTransfers || finalTransferStatus?.SkippedTransfers) {
+        if (finalTransferStatus?.FailedTransfers?.length || finalTransferStatus?.SkippedTransfers?.length) {
             message += localize('checkOutputWindow', ' Check the output window for a list of incomplete transfers.');
 
-            if (finalTransferStatus.FailedTransfers) {
+            if (finalTransferStatus.FailedTransfers?.length) {
                 ext.outputChannel.appendLog(localize('failedTransfers', 'Failed transfers:'));
                 for (let failedTransfer of finalTransferStatus.FailedTransfers) {
                     ext.outputChannel.appendLog(`\t${failedTransfer.Dst}`);
```
nturinski, line 43: How come this doesn't seem to get printed off in the screenshots?  Is this only for multiple files?

src/commands/azCopy/azCopyTransfer.ts
```diff
@@ -24,25 +24,26 @@ export async function azCopyTransfer(
...
     cancellationToken?: CancellationToken
 ): Promise<void> {
-    context.errorHandling.rethrow = true;
     const copyClient: AzCopyClient = new AzCopyClient();
     const copyOptions: ICopyOptions = { fromTo, overwriteExisting: "true", recursive: true, followSymLinks: true, excludePath: '.git;.vscode' };
     let jobId: string = await startAndWaitForCopy(context, copyClient, src, dst, copyOptions, transferProgress, notificationProgress, cancellationToken);
     let finalTransferStatus = (await copyClient.getJobInfo(jobId)).latestStatus;
     context.telemetry.properties.jobStatus = finalTransferStatus?.JobStatus;
     if (!finalTransferStatus || finalTransferStatus.JobStatus !== 'Completed') {
+        ext.outputChannel.appendLog(localize('couldNotUpload', 'Could not upload "{0}"', src.path));
```
nturinski, line 33: So I didn't realize from the screenshot, but this looks like this line would be redundant for the error handling that we have below.

src/commands/azCopy/azCopyTransfer.ts
wwlorey: Sometimes AzCopy doesn't return any `FailedTransfers` even when a transfer does fail

src/commands/azCopy/azCopyTransfer.ts
wwlorey: I went ahead and removed it

src/commands/azCopy/azCopyTransfer.ts
nturinski: Meh, that's 💩... I would like to have the message you just removed if something failed, but there's 0 failedTransfers in the array.

wwlorey: That message is back in business

nturinski: Hoh boy...

src/commands/azCopy/azCopyTransfer.ts
```diff
@@ -33,21 +32,24 @@ export async function azCopyTransfer(
...
-            if (finalTransferStatus.SkippedTransfers) {
+            if (finalTransferStatus.SkippedTransfers?.length) {
                 ext.outputChannel.appendLog(localize('skippedTransfers', 'Skipped transfers:'));
                 for (let skippedTransfer of finalTransferStatus.SkippedTransfers) {
                     ext.outputChannel.appendLog(`\t${skippedTransfer.Dst}`);
                 }
             }
+        } else {
+            // Add an additional error log since we don't have any more info about the failure
+            ext.outputChannel.appendLog(localize('couldNotUpload', 'Could not upload "{0}"', src.path));
```
nturinski, line 52: Okay, I think I like this 👍

nturinski (approved):

Issue #685, opened 2020-06-09
Extension does not allow multiple file/folder uploads


jmew:
As a data scientist, I would like to use Azure storage to store either 1 – my data sets (a folder with multiple subfolders with gigabytes of data), and 2 – my trained model (multiple files), such that if you want to create an API in azure based on that ML model, you can just grab it from azure storage.

wwlorey:
Fixed by https://github.com/microsoft/vscode-azurestorage/pull/775

Issue #757, opened 2020-09-02
Add better overwriting prompts for AzCopy transfers


wwlorey:
From https://github.com/microsoft/vscode-azurestorage/pull/754#discussion_r481499088

PR #760, opened 2020-09-02
Support uploading multiple files/folders via uploadToAzureStorage

wwlorey:
Related to https://github.com/microsoft/vscode-azurestorage/issues/685

I'll add context menu items to better expose this functionality in a different PR

### DIFF
--- a/src/commands/uploadToAzureStorage.ts
+++ b/src/commands/uploadToAzureStorage.ts
@@ -40,7 +40,7 @@ export async function uploadToAzureStorage(actionContext: IActionContext, target
     const title: string = localize('uploading', 'Uploading resources to "{0}"', treeItem.label);
     await vscode.window.withProgress({ cancellable: true, location: vscode.ProgressLocation.Notification, title }, async (notificationProgress, cancellationToken) => {
         for (const resourceUri of resourceUris) {
-            const resourcePath: string = resourceUri.path;
+            const resourcePath: string = resourceUri.fsPath;
             if ((await fse.stat(resourcePath)).isDirectory()) {
                 if (!multiResourceUpload) {
                     await showUploadWarning(localize('uploadWillOverwrite', 'Uploading "{0}" will overwrite any existing resources with the same name.', resourcePath));
--- a/src/commands/uploadToAzureStorage.ts
+++ b/src/commands/uploadToAzureStorage.ts
@@ -12,7 +12,6 @@ import { BlobContainerTreeItem } from '../tree/blob/BlobContainerTreeItem';
 import { FileShareTreeItem } from '../tree/fileShare/FileShareTreeItem';
 import { localize } from '../utils/localize';
 import { uploadFiles } from '../utils/uploadUtils';
-import { selectWorkspaceItems } from '../utils/workspaceUtils';
 
 export async function uploadToAzureStorage(actionContext: IActionContext, target?: vscode.Uri): Promise<void> {
     let resourceUris: vscode.Uri[];
@@ -23,16 +22,13 @@ export async function uploadToAzureStorage(actionContext: IActionContext, target
 
         resourceUris = [vscode.Uri.file(target.fsPath)];
     } else {
-        resourceUris = await selectWorkspaceItems(
-            ext.ui,
-            localize('selectResourceToUpload', 'Select resources to upload'),
-            {
-                canSelectFiles: true,
-                canSelectFolders: true,
-                canSelectMany: true,
-                defaultUri: vscode.workspace.workspaceFolders && vscode.workspace.workspaceFolders.length > 0 ? vscode.workspace.workspaceFolders[0].uri : undefined,
-                openLabel: localize('select', 'Select')
-            });
+        resourceUris = await ext.ui.showOpenDialog({
+            canSelectFiles: true,
+            canSelectFolders: true,
+            canSelectMany: true,
+            defaultUri: vscode.workspace.workspaceFolders && vscode.workspace.workspaceFolders.length > 0 ? vscode.workspace.workspaceFolders[0].uri : undefined,
+            openLabel: localize('select', 'Select')
+        });
     }
 
     const multiResourceUpload: boolean = resourceUris.length > 1;
--- a/src/commands/uploadToAzureStorage.ts
+++ b/src/commands/uploadToAzureStorage.ts
@@ -53,7 +53,8 @@ export async function uploadToAzureStorage(actionContext: IActionContext, target
                 // AzCopy recognizes folders as a resource when uploading to file shares. So only set `countFoldersAsResources=true` in that case
                 await uploadFiles(actionContext, treeItem, resourcePath, undefined, notificationProgress, cancellationToken, basename(resourcePath), treeItem instanceof FileShareTreeItem, multiResourceUpload);
             } else {
-                await treeItem.uploadLocalFile(actionContext, resourcePath, undefined, multiResourceUpload);
+                const destPath: string = basename(resourcePath);
+                await treeItem.uploadLocalFile(actionContext, resourcePath, destPath, multiResourceUpload);
             }
         }
     });
--- a/src/commands/uploadToAzureStorage.ts
+++ b/src/commands/uploadToAzureStorage.ts
@@ -47,8 +47,7 @@ export async function uploadToAzureStorage(actionContext: IActionContext, target
             const resourcePath: string = resourceUri.path;
             if ((await fse.stat(resourcePath)).isDirectory()) {
                 if (!multiResourceUpload) {
-                    const message: string = localize('uploadWillOverwrite', 'Uploading "{0}" will overwrite any existing resources with the same name.', resourcePath);
-                    await ext.ui.showWarningMessage(message, { modal: true }, { title: localize('upload', 'Upload') });
+                    await showUploadWarning(localize('uploadWillOverwrite', 'Uploading "{0}" will overwrite any existing resources with the same name.', resourcePath));
                 }
 
                 // AzCopy recognizes folders as a resource when uploading to file shares. So only set `countFoldersAsResources=true` in that case
--- a/src/commands/uploadToAzureStorage.ts
+++ b/src/commands/uploadToAzureStorage.ts
@@ -4,6 +4,7 @@
  *--------------------------------------------------------------------------------------------*/
 
 import * as fse from 'fs-extra';
+import { basename } from 'path';
 import * as vscode from 'vscode';
 import { IActionContext } from 'vscode-azureextensionui';
 import { ext } from '../extensionVariables';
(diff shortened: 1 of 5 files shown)


src/commands/uploadToAzureStorage.ts
```diff
@@ -4,47 +4,66 @@
...
-
-            // AzCopy recognizes folders as a resource when uploading to file shares. So only set `countFoldersAsResources=true` in that case
-            await uploadFiles(actionContext, treeItem, resourcePath, undefined, notificationProgress, cancellationToken, undefined, treeItem instanceof FileShareTreeItem);
-        } else {
-            await treeItem.uploadLocalFile(actionContext, resourcePath);
+        for (const resourceUri of resourceUris) {
+            const resourcePath: string = resourceUri.path;
+            if ((await fse.stat(resourcePath)).isDirectory()) {
+                if (!multiResourceUpload) {
+                    const message: string = localize('uploadWillOverwrite', 'Uploading "{0}" will overwrite any existing resources with the same name.', resourcePath);
```
EricJizbaMSFT, line 50: I assume you meant to use "showUploadWarning" here

src/tree/blob/BlobContainerTreeItem.ts
```diff
@@ -302,16 +302,22 @@ export class BlobContainerTreeItem extends AzureParentTreeItem<IStorageRoot> imp
...
     }
 
-    public async uploadLocalFile(context: IActionContext, filePath: string, blobPath?: string, suppressLogs: boolean = false): Promise<void> {
-        const destFolder: string = path.basename(filePath);
-        blobPath = blobPath !== undefined ? blobPath : await getBlobPath(this, destFolder);
-        if (await doesBlobExist(this, blobPath)) {
-            await warnFileAlreadyExists(blobPath);
+    public async uploadLocalFile(context: IActionContext, filePath: string, blobPath?: string, suppressLogsAndPrompts: boolean = false): Promise<void> {
+        if (blobPath === undefined) {
+            const destFolder: string = path.basename(filePath);
```
EricJizbaMSFT, line 307: The name of this variable confused me. It's the _file_ name, not the _folder_, right? How about naming it `fileName`

src/tree/blob/BlobContainerTreeItem.ts
```diff
@@ -302,16 +302,22 @@ export class BlobContainerTreeItem extends AzureParentTreeItem<IStorageRoot> imp
         }
     }
 
-    public async uploadLocalFile(context: IActionContext, filePath: string, blobPath?: string, suppressLogs: boolean = false): Promise<void> {
-        const destFolder: string = path.basename(filePath);
-        blobPath = blobPath !== undefined ? blobPath : await getBlobPath(this, destFolder);
-        if (await doesBlobExist(this, blobPath)) {
-            await warnFileAlreadyExists(blobPath);
+    public async uploadLocalFile(context: IActionContext, filePath: string, blobPath?: string, suppressLogsAndPrompts: boolean = false): Promise<void> {
```
EricJizbaMSFT, line 305: You can double check this, but I don't think `suppressLogs` was ever set to `true` in the past and I'm not sure why we would want to start suppressing them now. Feels like we should always show the logs and change `suppressLogsAndPrompts` to just `suppressPrompts`

src/utils/workspaceUtils.ts
```diff
@@ -13,7 +13,7 @@ export function getSingleRootWorkspace(): vscode.WorkspaceFolder | undefined {
     return vscode.workspace.workspaceFolders && vscode.workspace.workspaceFolders.length === 1 ? vscode.workspace.workspaceFolders[0] : undefined;
 }
 
-export async function selectWorkspaceItem(ui: IAzureUserInput, placeHolder: string, options: vscode.OpenDialogOptions, getSubPath?: (f: vscode.WorkspaceFolder) => string | undefined | Promise<string | undefined>): Promise<string> {
+export async function selectWorkspaceItems(ui: IAzureUserInput, placeHolder: string, options: vscode.OpenDialogOptions, getSubPath?: (f: vscode.WorkspaceFolder) => string | undefined | Promise<string | undefined>): Promise<vscode.Uri[]> {
     let folder: IAzureQuickPickItem<string | undefined> | undefined;
     if (vscode.workspace.workspaceFolders) {
```
EricJizbaMSFT, line 18: I know it was like this before your PR - but seems pretty weird to me that we prompt to upload your entire workspace first. That kind of logic makes sense for Functions/App Service when selecting a folder to deploy, but I find it unlikely someone would want to upload their whole workspace to storage. I guess I would prefer if we went straight to `showOpenDialog`

src/commands/uploadToAzureStorage.ts
```diff
@@ -4,47 +4,66 @@
...
-                canSelectMany: false,
+                canSelectMany: true,
                 defaultUri: vscode.workspace.workspaceFolders && vscode.workspace.workspaceFolders.length > 0 ? vscode.workspace.workspaceFolders[0].uri : undefined,
                 openLabel: localize('select', 'Select')
             });
     }
 
+    const multiResourceUpload: boolean = resourceUris.length > 1;
+    if (multiResourceUpload) {
+        await showUploadWarning(localize('uploadWillOverwrite', 'Uploading multiple files/folders will overwrite any existing resources with the same name.'));
```
EricJizbaMSFT, line 40: I'm not convinced this warning is enough. Part of my concern is that the multi-resource case doesn't prompt for a name, whereas the single-resource case did. Probably not a blocker for this PR, but would like to discuss the experience with Matt/Jing before releasing.

Alternatives include prompting for name for each resource (could get gross if you select a bunch) or adding some sort of logic that warns for each resource with "Yes" and "Yes to All" buttons

nturinski: I think that's being tracked here: https://github.com/microsoft/vscode-azurestorage/issues/757

EricJizbaMSFT: Okay cool - but just want to re-emphasize that this PR exacerbates the problem because it no longer prompts for the name in Azure

src/commands/uploadToAzureStorage.ts
```diff
@@ -4,47 +4,66 @@
...
-            const message: string = localize('uploadWillOverwrite', 'Uploading "{0}" will overwrite any existing resources with the same name.', resourcePath);
-            await ext.ui.showWarningMessage(message, { modal: true }, { title: localize('upload', 'Upload') });
-
-            // AzCopy recognizes folders as a resource when uploading to file shares. So only set `countFoldersAsResources=true` in that case
-            await uploadFiles(actionContext, treeItem, resourcePath, undefined, notificationProgress, cancellationToken, undefined, treeItem instanceof FileShareTreeItem);
-        } else {
-            await treeItem.uploadLocalFile(actionContext, resourcePath);
+        for (const resourceUri of resourceUris) {
+            const resourcePath: string = resourceUri.path;
+            if ((await fse.stat(resourcePath)).isDirectory()) {
```
nturinski, line 48: `ENOENT: no such file or directory, stat 'C:\c:\Users\Nathan\Desktop\Code\temp\functions-js\api'` getting this error on Windows when I try to upload anything (not just folders).

src/tree/blob/BlobContainerTreeItem.ts
wwlorey: I think the output channel gets a little busy/repetitive when we don't suppress logs.

| don't suppress logs | suppress logs |
| - | -| 
|Screen Shot 2020-09-03 at 12 26 26 PM|Screen Shot 2020-09-03 at 12 22 56 PM|

Do you think that's too much or should I just always display logs?

src/commands/uploadToAzureStorage.ts
wwlorey: There's some funkiness goin on with `fse.stat` 🧐

Using `resourceUri.path` returns the path starting with capital "C":
Screen Shot 2020-09-03 at 1 24 06 PM

And leads to this error:
Screen Shot 2020-09-03 at 1 24 19 PM


Whereas using `resourceUri.fsPath` returns the path starting with lowercase "c" and worked fine.
Screen Shot 2020-09-03 at 1 22 58 PM

Its possible that `fse.stat` prepends "C:" if it doesn't detect a *lowercase* "c" at the beginning of the path. But in Nathan's example his path already included a lowercase "c". So I dunno why exactly this fixes things...

src/commands/uploadToAzureStorage.ts
```diff
@@ -4,47 +4,61 @@
...
-        if ((await fse.stat(resourcePath)).isDirectory()) {
-            const message: string = localize('uploadWillOverwrite', 'Uploading "{0}" will overwrite any existing resources with the same name.', resourcePath);
-            await ext.ui.showWarningMessage(message, { modal: true }, { title: localize('upload', 'Upload') });
-
-            // AzCopy recognizes folders as a resource when uploading to file shares. So only set `countFoldersAsResources=true` in that case
-            await uploadFiles(actionContext, treeItem, resourcePath, undefined, notificationProgress, cancellationToken, undefined, treeItem instanceof FileShareTreeItem);
-        } else {
-            await treeItem.uploadLocalFile(actionContext, resourcePath);
+        for (const resourceUri of resourceUris) {
+            const resourcePath: string = resourceUri.fsPath;
```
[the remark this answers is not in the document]
wwlorey: Ah that makes sense..

src/commands/uploadToAzureStorage.ts
```diff
@@ -4,47 +4,61 @@
...
-            ext.ui,
-            localize('selectResourceToUpload', 'Select resource to upload'),
-            {
-                canSelectFiles: true,
-                canSelectFolders: true,
-                canSelectMany: false,
-                defaultUri: vscode.workspace.workspaceFolders && vscode.workspace.workspaceFolders.length > 0 ? vscode.workspace.workspaceFolders[0].uri : undefined,
-                openLabel: localize('select', 'Select')
-            });
+        resourceUris = await ext.ui.showOpenDialog({
```
nturinski, line 25: I don't know how concerning it is, but having `canSelectFiles` and `canSelectFolders` doesn't work well on Windows.  I think if it detects that both are set, it defaults to just letting you select only folders.  The workaround is just don't use the command palette, but something to consider.

wwlorey: I can tackle that here: https://github.com/microsoft/vscode-azurestorage/issues/768

src/commands/uploadToAzureStorage.ts
nturinski: Another thing I don't like about this message is you get it before even selecting an account, so even if you're creating a new blob, you'll see it.

wwlorey: > Another thing I don't like about this message is you get it before even selecting an account, so even if you're creating a new blob, you'll see it.

Is it alright to wait and address this in the fix for https://github.com/microsoft/vscode-azurestorage/issues/757? The overwrite logic will be different there anyway

nturinski: Yeah, it's fine, just about noting another issue with it.

src/tree/blob/BlobContainerTreeItem.ts
EricJizbaMSFT: I don't think it's too much. Even if it is, that's what "find" is for. The biggest reason I want to display logs is because you're skipping the naming prompt (as Nathan and I have brought up elsewhere). I want users to be able to check the logs to see which name we used since we didn't prompt

src/tree/blob/BlobContainerTreeItem.ts
```diff
@@ -301,16 +301,22 @@ export class BlobContainerTreeItem extends AzureParentTreeItem<IStorageRoot> imp
...
+        if (blobPath === undefined) {
+            const fileName: string = path.basename(filePath);
+
+            if (suppressLogsAndPrompts) {
```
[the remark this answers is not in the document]
wwlorey: > Is this a pattern that was already in storage? This is effectively letting the user skip the naming prompt, right? I'm not sure how I feel about it.

This would be a new pattern. I added it because asking for the blob name for every blob you upload in a multi-resource transfer was a lot of prompts. Storage explorer does the same thing - if you upload multiple blobs it just keeps their file names.

> Also, it seems like you could just pass blobPath as a parameter and do the path.basename(filePath) call before you invoke uploadLocalFile.

Good call - I went ahead and made that change

src/commands/uploadToAzureStorage.ts
```diff
@@ -4,49 +4,68 @@
...
+            canSelectFiles: true,
+            canSelectFolders: true,
+            canSelectMany: true,
+            defaultUri: vscode.workspace.workspaceFolders && vscode.workspace.workspaceFolders.length > 0 ? vscode.workspace.workspaceFolders[0].uri : undefined,
+            openLabel: localize('select', 'Select')
+        });
+    }
+
+    const multiResourceUpload: boolean = resourceUris.length > 1;
```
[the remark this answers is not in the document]
wwlorey: I added comments to help explain. Does that clear things up?

src/tree/blob/BlobContainerTreeItem.ts
wwlorey: I changed `suppressLogsAndPrompts` to just `suppressPrompts` and removed the 'success' log from each individual transfer

src/commands/uploadToAzureStorage.ts
```diff
@@ -4,49 +4,66 @@
...
-            // AzCopy recognizes folders as a resource when uploading to file shares. So only set `countFoldersAsResources=true` in that case
-            await uploadFiles(actionContext, treeItem, resourcePath, undefined, notificationProgress, cancellationToken, undefined, treeItem instanceof FileShareTreeItem);
-            ext.outputChannel.appendLog(localize('success', 'Successfully uploaded to "{0}".', treeItem.label));
-        } else {
-            await treeItem.uploadLocalFile(actionContext, resourcePath);
+    const title: string = multiResourceUpload ?
+        localize('uploadingTo', 'Uploading to "{0}"', treeItem.label) :
+        getUploadingMessage(resourceUris[0].fsPath, treeItem.label);
+    await vscode.window.withProgress({ cancellable: true, location: vscode.ProgressLocation.Notification, title }, async (notificationProgress, cancellationToken) => {
+        for (const resourceUri of resourceUris) {
```
EricJizbaMSFT, line 46: I would add a call to `throwIfCanceled` to the start of this loop. When we have a token, I feel like pretty much every loop should have that call

src/tree/blob/BlobContainerTreeItem.ts
```diff
@@ -301,29 +301,21 @@ export class BlobContainerTreeItem extends AzureParentTreeItem<IStorageRoot> imp
...
-            await warnFileAlreadyExists(blobPath);
-        }
-
-        const blobFriendlyPath: string = `${this.friendlyContainerName}/${blobPath}`;
-        if (!suppressLogs) {
-            ext.outputChannel.show();
-            ext.outputChannel.appendLog(getUploadingMessage(this.label, filePath));
+    public async uploadLocalFile(context: IActionContext, filePath: string, blobPath: string, suppressPrompts: boolean = false): Promise<void> {
+        if (!suppressPrompts) {
+            blobPath = await getBlobPath(this, blobPath);
```
EricJizbaMSFT, line 306: > > Also, it seems like you could just pass blobPath as a parameter and do the path.basename(filePath) call before you invoke uploadLocalFile.
> 
> Good call - I went ahead and made that change

I don't think you can do this because you have other places calling `uploadLocalFile` and they probably still need that logic, right? Double check the "Upload File" command and make sure that's working because I think it might be broken with your latest changes...

wwlorey: I double checked those scenarios still work 👍

src/commands/uploadToAzureStorage.ts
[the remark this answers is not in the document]
wwlorey: Yeah that's considered a single resource upload because its uploading one folder

src/commands/uploadToAzureStorage.ts
wwlorey: Ok I can do that in this PR - it may get a little chonky though 🎅

src/tree/blob/BlobContainerTreeItem.ts
EricJizbaMSFT: Okay well one issue I'm having with "Upload File" is duplicate prompts for the remote file path. Seems like there's a prompt in uploadFile.ts and blobUtils.getBlobPath

wwlorey: I'll clean up & combine `uploadFile` and `uploadToAzureStorage` in this PR to get rid of this weirdness

wwlorey: > Okay well one issue I'm having with "Upload File" is duplicate prompts for the remote file path. Seems like there's a prompt in uploadFile.ts and blobUtils.getBlobPath

Just pushed a fix for this

src/tree/blob/BlobContainerTreeItem.ts
```diff
@@ -301,29 +301,21 @@ export class BlobContainerTreeItem extends AzureParentTreeItem<IStorageRoot> imp
...
-            await warnFileAlreadyExists(blobPath);
-        }
-
-        const blobFriendlyPath: string = `${this.friendlyContainerName}/${blobPath}`;
-        if (!suppressLogs) {
-            ext.outputChannel.show();
-            ext.outputChannel.appendLog(getUploadingMessage(this.label, filePath));
+    public async uploadLocalFile(context: IActionContext, filePath: string, blobPath: string, suppressPrompts: boolean = false): Promise<void> {
+        if (!suppressPrompts) {
+            blobPath = blobPath !== undefined ? blobPath : await getBlobPath(this, blobPath);
```
EricJizbaMSFT, line 306: I still think there's something weird here because `blobPath` can never be undefined (based on the type). We might not need the call to `getBlobPath` at all. But I think it's fine to clean that up as a part of your future PR to clean up uploadFile vs. uploadToAzureStorage

EricJizbaMSFT (approved):

Issue #763, opened 2020-09-03
[Suggestion] Keep the message consistently when uploading a file to file share


v-mengwe:
**OS:** All
**Build Version:** 20200903.1

**Repro Steps:**

1. F1 -> Type and select "Upload to Azure Storage..." --> Select one file to upload --> Select a subscription --> Select a storage account --> Select "File Shares" --> Select a file share --> Press "Enter".
2. Check the output logs and notification.

**Actual:**
The message is inconsistent between the notification and the output logs.
![image](https://user-images.githubusercontent.com/62053665/92102435-32a8cf00-ee11-11ea-8bee-70e3ac0c30a4.png)

**Suggestion:**
Keep the message consistently between the notification and the output logs.

**More Info:**
This issue also reproduces for upload a folder to file share.

v-mengwe:
Verified on the latest build 20200909.18, this issue reproduces for now.
![image](https://user-images.githubusercontent.com/62053665/92676620-8d33b500-f354-11ea-9c46-4205bce89141.png)

1. This issue also reproduces when uploading a file to blob container.
2. This issue doesn't reproduce when uploading a folder to file share/blob container.

wwlorey:
This is by design. Since we support multiple file uploads now, it could get cumbersome to display all the file names in the notification

PR #766, opened 2020-09-03
Make log and notification consistent when uploading single file

wwlorey:
Fixes https://github.com/microsoft/vscode-azurestorage/issues/763

| Old | New |
|-|-|
|old1| new1|

### DIFF
--- a/src/tree/blob/BlobContainerTreeItem.ts
+++ b/src/tree/blob/BlobContainerTreeItem.ts
@@ -312,7 +312,7 @@ export class BlobContainerTreeItem extends AzureParentTreeItem<IStorageRoot> imp
         const blobFriendlyPath: string = `${this.friendlyContainerName}/${blobPath}`;
         if (!suppressLogs) {
             ext.outputChannel.show();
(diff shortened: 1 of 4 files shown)


src/tree/blob/BlobContainerTreeItem.ts
```diff
@@ -311,7 +312,8 @@ export class BlobContainerTreeItem extends AzureParentTreeItem<IStorageRoot> imp
         const blobFriendlyPath: string = `${this.friendlyContainerName}/${blobPath}`;
         if (!suppressLogs) {
             ext.outputChannel.show();
-            ext.outputChannel.appendLog(`Uploading ${filePath} as ${blobFriendlyPath}`);
+            uploadingMessage = uploadingMessage || localize('uploadingBlob', 'Uploading "{0}" as "{1}"', filePath, blobFriendlyPath);
```
nturinski, line 315: Why don't we just change the uploading message to be `localize('uploading', 'Uploading to "{0}" from "{1}"', treeItem.label, resourcePath);` instead of making it a parameter?  I'm not super attached to the old message personally.  

Also we can probably just make it a constant that we use in all 3 files.

nturinski (approved):

Issue #623, opened 2020-02-28
Split 'Upload to Azure Storage...' into 'Upload Folder...' and 'Upload Files...'


wwlorey:
From https://github.com/microsoft/vscode-azurestorage/pull/621#discussion_r385861300

PR #775, opened 2020-09-04
Split 'uploadToAzureStorage' into 'uploadFiles' and 'uploadFolder'

wwlorey:
I dialed back the functionality from https://github.com/microsoft/vscode-azurestorage/pull/760 to match what storage explorer does.

Screen Shot 2020-09-04 at 1 45 37 PM

This keeps the code simpler and still lets users upload whatever combo of files/folders they need to. (If you want to upload multiple folders you can upload one folder containing all the other folders you want to upload)

Fixes https://github.com/microsoft/vscode-azurestorage/issues/623 once and for all 😤

### DIFF
--- a/src/commands/uploadFile.ts
+++ b/src/commands/uploadFile.ts
@@ -4,51 +4,65 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { basename } from "path";
-import { OpenDialogOptions, Uri, window } from "vscode";
+import { Uri, window } from "vscode";
 import { IActionContext } from "vscode-azureextensionui";
+import { ext } from "../extensionVariables";
 import { BlobContainerTreeItem } from "../tree/blob/BlobContainerTreeItem";
 import { FileShareTreeItem } from "../tree/fileShare/FileShareTreeItem";
-import { localize } from "../utils/localize";
-import { validateFileName } from "../utils/validateNames";
+import { getBlobPath } from "../utils/blobUtils";
+import { getFileName } from "../utils/fileUtils";
+import { upload } from "../utils/uploadUtils";
 
-let lastUploadFolder: Uri;
+let lastUriUpload: Uri;
 
 export interface IExistingFileContext extends IActionContext {
     localFilePath: string;
     remoteFilePath: string;
 }
 
-export async function uploadFile(context: IActionContext, treeItem: BlobContainerTreeItem | FileShareTreeItem): Promise<void> {
-    const uris: Uri[] | undefined = await window.showOpenDialog(
-        <OpenDialogOptions>{
-            canSelectFiles: true,
-            canSelectFolders: false,
-            canSelectMany: false,
-            defaultUri: lastUploadFolder,
-            filters: {
-                "All files": ['*']
-            },
-            openLabel: "Upload"
-        }
-    );
-    if (uris && uris.length) {
-        const uri: Uri = uris[0];
-        lastUploadFolder = uri;
-        const localFilePath: string = uri.fsPath;
-        const remoteFilePath = await window.showInputBox({
-            prompt: localize('enterNameForFile', 'Enter a name for the uploaded file'),
-            value: basename(localFilePath),
-            validateInput: treeItem instanceof BlobContainerTreeItem ? BlobContainerTreeItem.validateBlobName : validateFileName
-        });
-        if (remoteFilePath) {
-            const id: string = `${treeItem.fullId}/${remoteFilePath}`;
-            const result = await treeItem.treeDataProvider.findTreeItem(id, context);
-            if (result) {
-                // A treeItem for this file already exists, no need to do anything with the tree, just upload
-                await treeItem.uploadLocalFile(context, localFilePath, remoteFilePath);
-                return;
+export async function uploadFiles(context: IActionContext, treeItem?: BlobContainerTreeItem | FileShareTreeItem, uris?: Uri[]): Promise<void> {
+    if (!uris) {
+        uris = await window.showOpenDialog(
+            {
+                canSelectFiles: true,
+                canSelectFolders: false,
+                canSelectMany: true,
+                defaultUri: lastUriUpload,
+                filters: {
+                    "All files": ['*']
+                },
+                openLabel: upload
             }
-            await treeItem.createChild(<IExistingFileContext>{ ...context, remoteFilePath, localFilePath });
+        );
+    }
+
+    if (!(treeItem instanceof BlobContainerTreeItem) && !(treeItem instanceof FileShareTreeItem)) {
+        treeItem = <BlobContainerTreeItem | FileShareTreeItem>(await ext.tree.showTreeItemPicker([BlobContainerTreeItem.contextValue, FileShareTreeItem.contextValue], context));
+    }
+
+    if (uris && uris.length) {
+        lastUriUpload = uris[0];
+        for (const uri of uris) {
+            const localFilePath: string = uri.fsPath;
+            let remoteFilePath: string | undefined = basename(localFilePath);
+
+            // Only prompt for upload name if we're uploading a single file
+            if (uris.length === 1) {
+                remoteFilePath = treeItem instanceof BlobContainerTreeItem ?
+                    await getBlobPath(treeItem, remoteFilePath) :
+                    await getFileName(treeItem, '', treeItem.shareName, remoteFilePath);
+            }
+
+            if (remoteFilePath) {
+                const id: string = `${treeItem.fullId}/${remoteFilePath}`;
+                const result = await treeItem.treeDataProvider.findTreeItem(id, context);
+                if (result) {
+                    // A treeItem for this file already exists, no need to do anything with the tree, just upload
+                    await treeItem.uploadLocalFile(context, localFilePath, remoteFilePath);
+                    continue;
+                }
+                await treeItem.createChild(<IExistingFileContext>{ ...context, remoteFilePath, localFilePath });
+            }
         }
     }
 }
--- a/src/commands/uploadFile.ts
+++ b/src/commands/uploadFile.ts
@@ -15,7 +15,7 @@ import { throwIfCanceled } from "../utils/errorUtils";
 import { getFileName } from "../utils/fileUtils";
 import { getUploadingMessage, upload } from "../utils/uploadUtils";
 
-let lastUriUpload: Uri;
+let lastUriUpload: Uri | undefined;
 
 export interface IExistingFileContext extends IActionContext {
     localFilePath: string;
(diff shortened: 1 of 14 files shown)


src/utils/fileUtils.ts
```diff
@@ -47,12 +47,15 @@ export async function getFileName(parent: AzureParentTreeItem<IStorageRoot>, dir
...
+            }
+
             let nameError = validateFileName(name);
             if (nameError) {
                 return nameError;
-            } else if (await doesFileExist(name, parent, directoryPath, shareName)) {
-                return "A file with this path and name already exists";
             }
+
             return undefined;
```
wwlorey, line 59: `validateInput` for file shares was not always detecting an existing file name and rearranging the function like this fixed it. I assume its because `doesFileExist` was taking a long time to return but I'm not totally sure

src/commands/uploadToAzureStorage.ts
```diff
@@ -3,62 +3,53 @@
  *  Licensed under the MIT License. See License.md in the project root for license information.
```
wwlorey, line 3: I'll rename `uploadToAzureStorage.ts` and `uploadFile.ts` to better match the new function names in a different PR

wwlorey:
Per offline discussion: Add a separate entry point for the file view that lets you select folders & files at the same time

src/commands/uploadFile.ts
```diff
@@ -4,51 +4,95 @@
...
+    context: IActionContext,
+    treeItem?: BlobContainerTreeItem | FileShareTreeItem,
+    uris?: Uri[],
+    notificationProgress?: Progress<{
+        message?: string | undefined;
+        increment?: number | undefined;
+    }>,
+    cancellationToken?: CancellationToken
+): Promise<void> {
+    uris = uris || await window.showOpenDialog(
```
EricJizbaMSFT, line 34: Use `ext.ui` instead of `window`. Biggest difference is how "cancel" is handled

src/commands/uploadFile.ts
```diff
@@ -4,51 +4,95 @@
...
-            const id: string = `${treeItem.fullId}/${remoteFilePath}`;
-            const result = await treeItem.treeDataProvider.findTreeItem(id, context);
-            if (result) {
-                // A treeItem for this file already exists, no need to do anything with the tree, just upload
-                await treeItem.uploadLocalFile(context, localFilePath, remoteFilePath);
-                return;
+    }
+}
+
+async function uploadFilesHelper(
```
EricJizbaMSFT, line 61: All the args for this helper should be required, not optional

src/commands/uploadFile.ts
```diff
@@ -4,51 +4,95 @@
...
 }
 
-export async function uploadFile(context: IActionContext, treeItem: BlobContainerTreeItem | FileShareTreeItem): Promise<void> {
-    const uris: Uri[] | undefined = await window.showOpenDialog(
-        <OpenDialogOptions>{
+export async function uploadFiles(
+    context: IActionContext,
+    treeItem?: BlobContainerTreeItem | FileShareTreeItem,
+    uris?: Uri[],
+    notificationProgress?: Progress<{
```
EricJizbaMSFT, line 28: Can you make a type for this? It's pretty ugly, and I assume annoying to copy everywhere. Something like this:
```
export type StorageProgress = vscode.Progress<{
    message?: string | undefined;
    increment?: number | undefined;
}>;
```

Should hopefully mean you don't have to do multi-line function definitions as much, either (which I also find ugly - but OK when a single line is super long)

src/commands/uploadFile.ts
```diff
@@ -4,51 +4,95 @@
...
+    }>,
+    cancellationToken?: CancellationToken
+): Promise<void> {
+    if (uris && uris.length) {
+        lastUriUpload = uris[0];
+        for (const uri of uris) {
+            throwIfCanceled(cancellationToken, context.telemetry.properties, 'uploadFiles');
+
+            const localFilePath: string = uri.fsPath;
+            let remoteFilePath: string | undefined = basename(localFilePath);
```
EricJizbaMSFT, line 77: I don't think `remoteFilePath` can be `undefined` anymore. That was only because the previous code was using `window.showInputBox` instead of `ext.ui.showInputBox`

src/commands/uploadFile.ts
```diff
@@ -4,51 +4,95 @@
...
+                    await getFileName(treeItem, '', treeItem.shareName, remoteFilePath);
+            }
+
+            if (remoteFilePath) {
+                const id: string = `${treeItem.fullId}/${remoteFilePath}`;
+                const result = await treeItem.treeDataProvider.findTreeItem(id, context);
+                if (result) {
+                    // A treeItem for this file already exists, no need to do anything with the tree, just upload
+                    await treeItem.uploadLocalFile(context, localFilePath, remoteFilePath, notificationProgress, cancellationToken);
+                    continue;
```
EricJizbaMSFT, line 92: You wouldn't even need "continue;" if you just stuck the below line in an `else` case

src/commands/uploadFolder.ts
```diff
@@ -0,0 +1,67 @@
...
+    treeItem?: BlobContainerTreeItem | FileShareTreeItem,
+    uri?: vscode.Uri,
+    notificationProgress?: vscode.Progress<{
+        message?: string | undefined;
+        increment?: number | undefined;
+    }>,
+    cancellationToken?: vscode.CancellationToken,
+    suppressPrompts?: boolean
+): Promise<void> {
+    if (uri?.scheme === 'azurestorage') {
```
EricJizbaMSFT, line 28: Pretty sure we want this in `uploadToAzureStorage` since that's the file system entry point. Otherwise I could right click on a file and I wouldn't get this error

src/commands/uploadToAzureStorage.ts
```diff
@@ -3,69 +3,39 @@
...
-    let resourceUris: vscode.Uri[];
-    if (target) {
-        if (target.scheme === 'azurestorage') {
-            throw new Error(localize('cannotUploadToAzureFromAzureResource', 'Cannot upload to Azure from an Azure resource.'));
+import { getUploadingMessage } from '../utils/uploadUtils';
+import { uploadFiles } from './uploadFile';
+import { uploadFolder } from './uploadFolder';
+
+export async function uploadToAzureStorage(actionContext: IActionContext, _firstSelection: vscode.Uri, uris: vscode.Uri[]): Promise<void> {
+    const folderUris: vscode.Uri[] = [];
```
EricJizbaMSFT, line 18: I would define a variable here like
```
const suppressPrompts: boolean = uris.length > 1;
```

and pass that in to both `uploadFolder` and `uploadFiles`. There's some edge cases that I don't think the current code handles well (like if I select 30 folders and 1 file - I think I still get prompted for the file name - which I don't think we want)

src/commands/uploadToAzureStorage.ts
```diff
@@ -3,69 +3,39 @@
...
-            defaultUri: vscode.workspace.workspaceFolders && vscode.workspace.workspaceFolders.length > 0 ? vscode.workspace.workspaceFolders[0].uri : undefined,
-            openLabel: localize('select', 'Select')
-        });
-    }
-
-    // AzCopy works by uploading either single files or entire folders.
-    // So uploading multiple resources only happens if we're uploading more than one file/folder.
-    const multiResourceUpload: boolean = resourceUris.length > 1;
-    if (multiResourceUpload) {
-        await showUploadWarning(localize('uploadWillOverwrite', 'Uploading multiple files/folders will overwrite any existing resources with the same name.'));
```
EricJizbaMSFT, line 39: I think we're missing a warning message for the "uploadToAzureStorage" entry point after your changes

src/utils/fileUtils.ts
EricJizbaMSFT: Async code as a part of validateInput is a pretty bad experience. "validateInput" is executed every time the user enters a character and the async code can be jarringly slow. That's why we always put it last so that the better "sync" code is more likely to show an error first.

I'm skeptical your change actually fixes the issue and it doesn't seem super related to this goal of the PR. Can we investigate separately?

src/utils/uploadUtils.ts
```diff
@@ -14,55 +13,38 @@ import { ext } from '../extensionVariables';
...
-        { title: localize('overwrite', 'Overwrite') }
-    );
+export function getUploadingMessage(treeItemLabel: string): string {
+    return localize('uploadingTo', 'Uploading to "{0}"', treeItemLabel);
 }
 
-export function getUploadingMessage(sourcePath: string, treeItemLabel: string): string {
-    return localize('uploading', 'Uploading "{0}" to "{1}"', sourcePath, treeItemLabel);
+export function getUploadingMessageWithSource(sourcePath: string, treeItemLabel: string): string {
+    return localize('uploadingFromTo', 'Uploading "{0}" to "{1}"', sourcePath, treeItemLabel);
```
EricJizbaMSFT, line 47: I agree with "uploadingFromTo" - it should be `Uploading from "{0}" to "{1}"`

src/utils/fileUtils.ts
wwlorey: Tracking that issue here: https://github.com/microsoft/vscode-azurestorage/issues/780

src/commands/uploadToAzureStorage.ts
```diff
@@ -3,69 +3,45 @@
...
     }
 
-    let treeItem: BlobContainerTreeItem | FileShareTreeItem = await ext.tree.showTreeItemPicker([BlobContainerTreeItem.contextValue, FileShareTreeItem.contextValue], actionContext);
-    const title: string = multiResourceUpload ?
-        localize('uploadingTo', 'Uploading to "{0}"', treeItem.label) :
-        getUploadingMessage(resourceUris[0].fsPath, treeItem.label);
+    const treeItem: BlobContainerTreeItem | FileShareTreeItem = await ext.tree.showTreeItemPicker([BlobContainerTreeItem.contextValue, FileShareTreeItem.contextValue], actionContext);
+    const suppressPrompts: boolean = uris.length > 1;
+    const title: string = getUploadingMessage(treeItem.label);
+    await showUploadWarning(localize('uploadingToWillOverwrite', 'Uploading to "{0}" will overwrite any existing resources with the same name.', treeItem.label));
```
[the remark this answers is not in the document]
wwlorey: There's only a single prompt in `uploadToAzureStorage` since we pass `suppressPrompts` into `uploadFolder` and `uploadFiles`. If we only show the upload warning when `suppressPrompts = true` we would only warn about overwriting during a single URI upload.

I moved the `suppressPrompts` declaration to hopefully make this easier to read but lemme know if that doesn't help

wwlorey:
Ok hopefully the `uploadToAzureStorage` prompt problem is fixed now. The naming is a bit bizarre but I had trouble coming up with a cleaner way

EricJizbaMSFT (approved):

PR #779, opened 2020-09-08
Use AzCopy for blob downloads

wwlorey:
Related to https://github.com/microsoft/vscode-azurestorage/issues/755

### DIFF
--- a/src/tree/blob/BlobTreeItem.ts
+++ b/src/tree/blob/BlobTreeItem.ts
@@ -1,8 +1,9 @@
 /*
-  *  Copyright (c) Microsoft Corporation. All rights reserved.
-  *  Licensed under the MIT License. See License.md in the project root for license information.
-  **/
+*  Copyright (c) Microsoft Corporation. All rights reserved.
+*  Licensed under the MIT License. See License.md in the project root for license information.
+**/
 
+import { ILocalLocation, IRemoteSasLocation } from "@azure-tools/azcopy-node";
 import * as azureStorageBlob from "@azure/storage-blob";
 import { BlobGetPropertiesResponse, BlockBlobClient } from "@azure/storage-blob";
 import * as path from 'path';
@@ -10,11 +11,14 @@ import * as vscode from 'vscode';
 import { MessageItem, SaveDialogOptions, Uri, window } from 'vscode';
 import { AzureParentTreeItem, AzureTreeItem, DialogResponses, IActionContext, UserCancelledError } from 'vscode-azureextensionui';
 import { AzureStorageFS } from "../../AzureStorageFS";
+import { createAzCopyLocalLocation, createAzCopyRemoteLocation } from "../../commands/azCopy/azCopyLocations";
+import { azCopyTransfer } from "../../commands/azCopy/azCopyTransfer";
 import { getResourcesPath } from "../../constants";
 import { ext } from "../../extensionVariables";
 import { TransferProgress } from "../../TransferProgress";
 import { createBlobClient, createBlockBlobClient } from '../../utils/blobUtils';
 import { Limits } from "../../utils/limits";
+import { localize } from "../../utils/localize";
 import { ICopyUrl } from '../ICopyUrl';
 import { IStorageRoot } from "../IStorageRoot";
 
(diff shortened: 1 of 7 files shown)


src/tree/blob/BlobTreeItem.ts
```diff
@@ -96,38 +100,28 @@ export class BlobTreeItem extends AzureTreeItem<IStorageRoot> implements ICopyUr
...
-
         const client: BlockBlobClient = createBlockBlobClient(this.root, this.container.name, this.blobPath);
         let props: BlobGetPropertiesResponse = await client.getProperties();
-
         context.telemetry.measurements.blobDownloadSize = props.contentLength;
-        if (Number(props.contentLength) > Limits.maxUploadDownloadSizeBytes) {
-            context.telemetry.properties.blobTooLargeForDownload = 'true';
-            message = `Please use Storage Explorer for blobs larger than ${Limits.maxUploadDownloadSizeMB}MB.`;
-        } else if (props.blobType && !props.blobType.toLocaleLowerCase().startsWith("block")) {
+        if (props.blobType && !props.blobType.toLocaleLowerCase().startsWith("block")) {
```
nturinski, line 121: Dose our limitation with block blob still exist now that we're using AzCopy?

wwlorey: Yeah we still don't support page or append blobs

nturinski: I thought the reason we didn't support anything but block blobs was because the API couldn't download/upload them (but we can still list them).  Did you confirm that it doesn't work with the new AzCopy implementation?

src/utils/limits.ts
```diff
@@ -1,28 +1,23 @@
...
 import { IActionContext, UserCancelledError } from "vscode-azureextensionui";
 import { ResourceType } from "../storageExplorerLauncher/ResourceType";
 import { storageExplorerLauncher } from "../storageExplorerLauncher/storageExplorerLauncher";
+import { localize } from "./localize";
 
-export namespace Limits {
-    //  VS Code currently supports at least 256MB, but not 512MB. But it won't open anything larger than 4MB through the APIs.
-    export const maxUploadDownloadSizeMB = 4;
-    export const maxUploadDownloadSizeBytes = maxUploadDownloadSizeMB * 1000 * 1000;
+export async function askOpenInStorageExplorer(context: IActionContext, errorMessage: string, resourceId: string, subscriptionId: string, resourceType: ResourceType, resourceName: string): Promise<void> {
```
nturinski, line 12: We should change the name of this file since all the other limits have been removed.

wwlorey: I was going to change that in a separate PR to make the diff easier to see

src/tree/blob/BlobTreeItem.ts
wwlorey: I thought there was other work required to support those blob types but I'll try it and see if it works

wwlorey: Downloading append blobs with AzCopy works fine but it looks like it would take more work to fully support them. We would need to modify how we're using "@azure/storage-blob" or ditch it for AzCopy when saving an append blob in the editor. Right now that leads to this error: `Error: The blob type is invalid for this operation.` We would probably also want to support choosing the blob type when uploading.

I wasn't able to easily test page blobs - storage explorer kept throwing errors related to authentication when I tried creating one.

Would you be cool with tracking this as a separate issue? I wanna make sure we can get download in for the private build that's going out today

nturinski: Yeah, no problem. I just wanted to make sure that only limitation wasn't just the download/upload portions that blocked this (since Storage Explorer is supposed to support these blobs apparently).  I don't know if we need an issue honestly, not sure if it's part of our plan to support it anyway.

src/utils/limits.ts
nturinski: Actually, I think that this file can be straight up deleted since askOpenInStorageExplorer is only used once.  But yeah, that can wait until a separate PR.

nturinski (approved):

PR #782, opened 2020-09-09
Delete limits.ts

wwlorey:


src/tree/blob/BlobTreeItem.ts
```diff
@@ -121,9 +122,22 @@ export class BlobTreeItem extends AzureTreeItem<IStorageRoot> implements ICopyUr
         if (props.blobType && !props.blobType.toLocaleLowerCase().startsWith("block")) {
             context.telemetry.properties.invalidBlobTypeForDownload = 'true';
             const message: string = localize('pleaseUseSE', 'Please use Storage Explorer for blobs of type "{0}".', props.blobType);
-            await askOpenInStorageExplorer(context, message, this.root.storageAccountId, this.root.subscriptionId, 'Azure.BlobContainer', this.container.name);
+            await this.askOpenInStorageExplorer(context, message, this.root.storageAccountId, this.root.subscriptionId, 'Azure.BlobContainer', this.container.name);
         }
     }
+
+    private async askOpenInStorageExplorer(context: IActionContext, errorMessage: string, resourceId: string, subscriptionId: string, resourceType: ResourceType, resourceName: string): Promise<void> {
```
nturinski, line 129: I don't think we need to pass in the `errorMessage` anymore.  But also, maybe we should leave it in its own file so we're not adding bloat functions to this tree item since it has a decent amount of code in it already.

wwlorey: I went ahead and moved it to its own file. I'd prefer to keep `errorMessage` as a parameter though so its generic enough to be used elsewhere if we need to. What do you think?

nturinski (approved):

src/tree/blob/BlobTreeItem.ts
nturinski: Yeah, I'm fine with it.  Maybe you could just pass a parameter to replace the noun like blob or file share, but either way is gravy.

PR #786, opened 2020-09-10
Add overwrite prompts for resource uploads

wwlorey:
This takes care of all the overwrite prompts before any transfers are made.

Fixes https://github.com/microsoft/vscode-azurestorage/issues/757

### DIFF
--- a/src/commands/uploadFiles.ts
+++ b/src/commands/uploadFiles.ts
@@ -3,6 +3,7 @@
  *  Licensed under the MIT License. See License.md in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
+import { basename } from "path";
 import { CancellationToken, ProgressLocation, Uri, window } from "vscode";
 import { IActionContext } from "vscode-azureextensionui";
 import { NotificationProgress } from "../constants";
@@ -11,7 +12,7 @@ import { BlobContainerTreeItem } from "../tree/blob/BlobContainerTreeItem";
 import { FileShareTreeItem } from "../tree/fileShare/FileShareTreeItem";
 import { throwIfCanceled } from "../utils/errorUtils";
 import { nonNullValue } from "../utils/nonNull";
-import { getRemoteResourceName, getUploadingMessage, OverwriteChoice, RemoteResourceNameMap, upload } from "../utils/uploadUtils";
+import { getUploadingMessage, OverwriteChoice, shouldUploadUri, upload } from "../utils/uploadUtils";
 
 let lastUriUpload: Uri | undefined;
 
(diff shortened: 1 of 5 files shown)


src/utils/blobUtils.ts
```diff
@@ -87,6 +87,22 @@ export async function doesBlobExist(treeItem: BlobContainerTreeItem | BlobDirect
...
 
+export async function doesBlobDirectoryExist(treeItem: BlobContainerTreeItem | BlobDirectoryTreeItem, blobDirectoryName: string): Promise<boolean> {
+    const sep: string = path.posix.sep;
+    if (!blobDirectoryName.endsWith(sep)) {
+        blobDirectoryName = `${blobDirectoryName}${sep}`;
+    }
+
+    const containerClient: azureStorageBlob.ContainerClient = createBlobContainerClient(treeItem.root, treeItem.container.name);
+    const prefix: string | undefined = treeItem instanceof BlobDirectoryTreeItem ? treeItem.dirPath : undefined;
+    let response: AsyncIterableIterator<azureStorageBlob.ContainerListBlobHierarchySegmentResponse> = containerClient.listBlobsByHierarchy(sep, { prefix }).byPage();
```
EricJizbaMSFT, line 98: So this won't be reliable. Unless we list out _all_ blobs with some sort of loop (which I don't want to do because it could take forever), we can't know for sure if the directory exists or not. I assume there's no "exists" method because this is for blobs and directories aren't actually a thing?

What does Storage Explorer do? I don't have a solution for this problem yet, but I'll think about it...

EricJizbaMSFT (changes requested):

src/commands/uploadFiles.ts
```diff
@@ -45,15 +42,23 @@ export async function uploadFiles(
     );
 
     treeItem = treeItem || <BlobContainerTreeItem | FileShareTreeItem>(await ext.tree.showTreeItemPicker([BlobContainerTreeItem.contextValue, FileShareTreeItem.contextValue], context));
-    // tslint:enable: strict-boolean-expressions
+    if (remoteResourceNameMap === undefined) {
+        remoteResourceNameMap = new Map();
+        let overwriteChoice: OverwriteChoice = { choice: undefined };
+        let remoteFileName: string;
+        for (const uri of uris) {
```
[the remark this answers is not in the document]
wwlorey: I just modified this part of the code, is `await Promise.all()` still needed? I'm not familiar with how thats used

src/utils/blobUtils.ts
```diff
@@ -87,6 +87,22 @@ export async function doesBlobExist(treeItem: BlobContainerTreeItem | BlobDirect
     return blobClient.exists();
 }
 
+export async function doesBlobDirectoryExist(treeItem: BlobContainerTreeItem | BlobDirectoryTreeItem, blobDirectoryName: string): Promise<boolean> {
+    const sep: string = path.posix.sep;
+    if (!blobDirectoryName.endsWith(sep)) {
```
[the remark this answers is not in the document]
wwlorey: Yeah it does since its an Azure path

src/utils/blobUtils.ts
wwlorey: I just pushed a change that doesn't involve paging based on this example: https://github.com/Azure/azure-sdk-for-js/blob/cf3ffd7ed1d154753f206c95862100b9dfadfc17/sdk/storage/storage-blob/src/Clients.ts#L8079-L8087

Its still a loop that could take a long time though. I'll keep thinking of what else we could do

wwlorey: Lol ok it turns out you can pass in a specific prefix to search on 🤦‍♂️ The newest version should be a constant time lookup

src/commands/uploadFiles.ts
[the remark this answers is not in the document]
EricJizbaMSFT: As long as you await `shouldUploadUri` (which you're already doing), you don't need `Promise.all`. You could theoretically refactor to use `Promise.all` if you wanted to execute those in parallel, but I don't think that would be super helpful.

src/utils/blobUtils.ts
```diff
@@ -87,6 +87,22 @@ export async function doesBlobExist(treeItem: BlobContainerTreeItem | BlobDirect
...
+        blobDirectoryName = `${blobDirectoryName}${sep}`;
+    }
+
+    const containerClient: azureStorageBlob.ContainerClient = createBlobContainerClient(treeItem.root, treeItem.container.name);
+
+    // tslint:disable-next-line: await-promise
+    for await (let _ of containerClient.listBlobsByHierarchy(sep, { prefix: blobDirectoryName })) {
```
EricJizbaMSFT, line 99: This line with the tslint rule disable looks super weird to me. Can you clean it up and avoid disabling the lint rule?

EricJizbaMSFT: Also it's technically possible for the list operation to return a page with zero blobs, even if there are blobs. I believe the setting is _max_ page size, not _exact_ page size. I would prefer if you did something like this:
- set page size to 1 (for performance reasons)
- If it returned a blob, return true
- If it didn't return a blob, check if there is a next page (I assume this is possible)
    - if yes, check that page (and continue looping as necessary)
    - if no, return false

src/utils/uploadUtils.ts
```diff
@@ -45,8 +56,53 @@ export function getUploadingMessageWithSource(sourcePath: string, treeItemLabel:
...
+    if (overwriteChoice.choice === OverwriteChoice.noToAll) {
+        // No need to check if this resource exists. Don't upload
+        return false;
+    } else if (overwriteChoice.choice === OverwriteChoice.yesToAll) {
+        // Always upload
+        return true;
+    }
+
+    const localResourcePath: string = uri.fsPath;
+    const remoteResourceName: string = basename(localResourcePath);
```
EricJizbaMSFT, line 93: nit: Please make a function like `convertLocalPathToRemotePath` that just does `basename(localPath)` and use that here and in uploadFolder.ts/uploadFile.ts

I know it's a simple function, but I want to be sure that the code checking if a file exists uses the exact same path when actually uploading

src/utils/blobUtils.ts
wwlorey: Yeah looks like the only page setting we have access to is max page size. I also added a check for `blobPrefixes` because there could be a nested directory like `dir1/dir2/` where there's only blobs at the `dir2/` level. `dir1/` wouldn't have any blobs listed directly under it in the response

src/commands/uploadFiles.ts
```diff
@@ -27,33 +25,45 @@ export async function uploadFiles(
...
-            defaultUri: lastUriUpload,
-            filters: {
-                "All files": ['*']
-            },
-            openLabel: upload
-        }
-    );
+    let shouldCheckUris: boolean = false;
+    if (uris === undefined) {
+        // tslint:disable: strict-boolean-expressions
```
nturinski, line 32: I think you can remove this now

src/commands/uploadFolder.ts
```diff
@@ -3,55 +3,52 @@
...
-    uri = uri || (await ext.ui.showOpenDialog({
-        canSelectFiles: false,
-        canSelectFolders: true,
-        canSelectMany: false,
-        defaultUri: vscode.workspace.workspaceFolders && vscode.workspace.workspaceFolders.length > 0 ? vscode.workspace.workspaceFolders[0].uri : undefined,
-        openLabel: upload
-    }))[0];
+    let shouldCheckUri: boolean = false;
+    if (uri === undefined) {
+        // tslint:disable: strict-boolean-expressions
```
nturinski, line 24: Same here, I believe.

EricJizbaMSFT (approved):

Issue #796, opened 2020-09-11
An error occurs when downloading a folder from blob container/file share


v-ruizh:
**OS:** Windows 10
**Build Version:** 20200911.6

**Repro Steps:**
Download a folder from blob container/file share.

**Expect:**
Download this folder successfully.

**Actual:**
An error occurs when downloading a folder from blob container/file share.
![image](https://user-images.githubusercontent.com/45614674/92897372-ba18c280-f44f-11ea-9d41-a54f463e774b.png)

**More Info:**
This issue does not reproduce on Mac.

PR #801, opened 2020-09-11
Don't use platform specific separator for AzCopy remote location

wwlorey:
The path used in `createAzCopyRemoteLocation` is always an Azure path which uses the `/` separator.

Fixes https://github.com/microsoft/vscode-azurestorage/issues/796

### DIFF
--- a/src/commands/azCopy/azCopyLocations.ts
+++ b/src/commands/azCopy/azCopyLocations.ts
@@ -6,7 +6,7 @@
 import { ILocalLocation, IRemoteSasLocation } from "@azure-tools/azcopy-node";
 import { AccountSASPermissions, AccountSASSignatureValues, ContainerClient } from "@azure/storage-blob";
 import { ShareClient } from "@azure/storage-file-share";
-import { sep } from "path";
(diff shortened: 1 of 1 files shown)


src/commands/azCopy/azCopyLocations.ts
```diff
@@ -26,8 +26,8 @@ export function createAzCopyLocalLocation(path: string, isFolder?: boolean): ILo
 }
 
 export function createAzCopyRemoteLocation(treeItem: BlobTreeItem | BlobDirectoryTreeItem | BlobContainerTreeItem | FileTreeItem | DirectoryTreeItem | FileShareTreeItem, path: string, isFolder?: boolean): IRemoteSasLocation {
-    if (isFolder && !path.endsWith(sep)) {
-        path += sep;
+    if (isFolder && !path.endsWith('/')) {
```
EricJizbaMSFT, line 29: You could use `posix.sep` from "path" to accomplish the same thing. IMO, that makes it clear this was intentional instead of accidentally hard-coding a slash

EricJizbaMSFT (approved):

Issue #806, opened 2020-09-14
Suggested string changes for "Upload to Azure Storage"


EricJizbaMSFT:
<img width="257" alt="Screen Shot 2020-09-14 at 11 18 39 AM" src="https://user-images.githubusercontent.com/11282622/93122970-3d1e6080-f67c-11ea-869c-78188d3829c9.png">

Not sure what the units are with 23/23 (for a file) and 4/4 (for a folder). Suggest adding units

<img width="296" alt="Screen Shot 2020-09-14 at 11 18 29 AM" src="https://user-images.githubusercontent.com/11282622/93122920-2b3cbd80-f67c-11ea-9a21-e76ba793a783.png">

Should have a "..." at the end for an in-progress message

<img width="273" alt="Screen Shot 2020-09-14 at 11 18 13 AM" src="https://user-images.githubusercontent.com/11282622/93122904-26780980-f67c-11ea-80b4-980d86e92c23.png">

Should have a "." at the end

PR #819, opened 2020-09-17
Add punctuation and units to AzCopy transfers

wwlorey:
Fixes https://github.com/microsoft/vscode-azurestorage/issues/806

### DIFF
--- a/src/TransferProgress.ts
+++ b/src/TransferProgress.ts
@@ -15,6 +15,7 @@ export class TransferProgress {
     constructor(
         public totalWork?: number,
         private readonly messagePrefix?: string,
+        private readonly units?: 'bytes' | 'files',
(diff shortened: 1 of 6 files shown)


nturinski (approved):

nturinski (commented):
Was there anywhere `new Transfer Progress()` didn't include that unit type?  Why make it an optional param?

wwlorey:
> Was there anywhere `new Transfer Progress()` didn't include that unit type? Why make it an optional param?

I didn't originally include units for blob deletions [here](https://github.com/microsoft/vscode-azurestorage/blob/wwl/string-changes/src/tree/blob/BlobContainerTreeItem.ts#L226-L227) since the output channel already says "blobs" on each line. But might as well add units to the transfer progress notification too

Issue #814, opened 2020-09-15
Add prompt for "Destination Directory" during "Upload to Storage"


EricJizbaMSFT:
If I do "Upload to Storage" from the file view, the only place I can upload to is the root of a blob container or file share. I believe @wwlorey did this because he didn't want to prompt for a remote path 100 times if you select 100 items. However, Storage Explorer gets around this by asking for "Destination Directory". It gives people a lot more flexibility, but is only one prompt (and applies to all 100 items)

PR #822, opened 2020-09-22
Add destination directory prompt when uploading blobs/files

wwlorey:
Fixes https://github.com/microsoft/vscode-azurestorage/issues/814

### DIFF
--- a/src/commands/uploadFolder.ts
+++ b/src/commands/uploadFolder.ts
@@ -10,7 +10,7 @@ import { ext } from '../extensionVariables';
 import { BlobContainerTreeItem } from '../tree/blob/BlobContainerTreeItem';
 import { FileShareTreeItem } from '../tree/fileShare/FileShareTreeItem';
 import { nonNullValue } from '../utils/nonNull';
-import { convertLocalPathToRemotePath, getUploadingMessageWithSource, promptForDestinationDirectory, shouldUploadUri, upload, uploadLocalFolder } from '../utils/uploadUtils';
+import { convertLocalPathToRemotePath, getDestinationDirectory, getUploadingMessageWithSource, shouldUploadUri, upload, uploadLocalFolder } from '../utils/uploadUtils';
 
 export async function uploadFolder(
     actionContext: IActionContext,
@@ -34,7 +34,7 @@ export async function uploadFolder(
 
     // tslint:disable-next-line: strict-boolean-expressions
     treeItem = treeItem || <BlobContainerTreeItem | FileShareTreeItem>(await ext.tree.showTreeItemPicker([BlobContainerTreeItem.contextValue, FileShareTreeItem.contextValue], actionContext));
-    destinationDirectory = destinationDirectory !== undefined ? destinationDirectory : await promptForDestinationDirectory();
+    destinationDirectory = await getDestinationDirectory(destinationDirectory);
 
     if (shouldCheckUri && !(await shouldUploadUri(treeItem, uri, { choice: undefined }, destinationDirectory))) {
         // Don't upload this folder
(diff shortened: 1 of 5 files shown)


src/commands/uploadFolder.ts
```diff
@@ -10,14 +10,15 @@ import { ext } from '../extensionVariables';
...
-import { convertLocalPathToRemotePath, getUploadingMessageWithSource, shouldUploadUri, upload, uploadLocalFolder } from '../utils/uploadUtils';
+import { convertLocalPathToRemotePath, getUploadingMessageWithSource, promptForDestinationDirectory, shouldUploadUri, upload, uploadLocalFolder } from '../utils/uploadUtils';
 
 export async function uploadFolder(
     actionContext: IActionContext,
     treeItem?: BlobContainerTreeItem | FileShareTreeItem,
     uri?: vscode.Uri,
     notificationProgress?: NotificationProgress,
     cancellationToken?: vscode.CancellationToken,
+    destinationDirectory?: string
```
nturinski, line 21: What do you think about making an uploadOptions object?  It seems like you use these 6 properties in uploadFiles, uploadFolders, and uploadFilesHelper.

src/utils/uploadUtils.ts
```diff
@@ -106,8 +106,21 @@ export async function shouldUploadUri(treeItem: BlobContainerTreeItem | FileShar
...
+    let path: string = posix.join(destinationDirectory, basename(localPath));
+    if (path.startsWith(posix.sep)) {
+        // `BlobClient` and `ShareFileClient` treat resource paths that start with "/" differently from those that don't. So remove the leading "/".
+        // This check is done after `posix.join()` because that function removes any duplicate slashes from the beginning of the path.
+        path = path.substr(1);
+    }
+    return path;
+}
+
+export async function promptForDestinationDirectory(): Promise<string> {
```
nturinski, line 119: I think I'd prefer calling this `getDestinationDirectory(dir?: string)` and then returning `destinationDirectory !== undefined ? destinationDirectory : await promptForDestinationDirectory();`

You do this logic twice and if we wanted to have a defaultDirectory setting, it'd be easier to change it here.

src/commands/uploadFolder.ts
wwlorey: I dig it. Since `uploadFiles` and `uploadFolders` are entry points themselves, I'll add everything except `actionContext` and `treeItem` to `uploadOptions`

wwlorey: I tried implementing this and I dig it a little less 😞 It got messy because `uploadFiles` takes multiple URIs while `uploadFolder` takes a single URI. 
Also the list of URIs passed into `uploadFiles` isn't necessarily the same list it passes into `uploadFilesHelper`. 

Should I try making the change anyway just to see what it looks like?

nturinski (approved):

src/tree/fileShare/FileShareTreeItem.ts
```diff
@@ -132,10 +132,12 @@ export class FileShareTreeItem extends AzureParentTreeItem<IStorageRoot> impleme
         // Ensure parent directories exist before creating child files
         let partialParentDirectoryPath: string = '';
         for (let dir of parentDirectories) {
-            partialParentDirectoryPath += `${dir}/`;
-            if (!(await doesDirectoryExist(this, partialParentDirectoryPath, this.shareName))) {
-                const directoryClient: azureStorageShare.ShareDirectoryClient = createDirectoryClient(this.root, this.shareName, partialParentDirectoryPath);
-                await directoryClient.create();
+            if (!!dir) {
```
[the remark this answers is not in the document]
wwlorey: This check was needed when there were paths that started with "/". Splitting a path like "/dir" would produce an array `['', 'dir']` and trying to create a `ShareDirectoryClient` on that empty directory would throw an error.

But since I added logic to remove the leading "/" from the directory path I can remove this check

Issue #817, opened 2020-09-16
First failed transfer abruptly ends all remaining transfers


nturinski:
Repro steps
1. Highlight multiple files in File Explorer
1. Upload To Azure Storage...
1. One file successfully transfers
1. Next file fails
1. Remaining files do not get uploaded

![image](https://user-images.githubusercontent.com/5290572/93369267-ffe4da80-f803-11ea-8e34-11abc79f44e1.png)

wwlorey:
I wasn't able to repro on Mac or Windows with random files or specifically uploading ".log" files.

Does this repro for you regardless of what the second file is?

nturinski:
I should mention that I was using your latest branch wwl/no-to-all

nturinski:
![image](https://user-images.githubusercontent.com/5290572/93371112-b053de00-f806-11ea-9119-3bf09130a969.png)

Here were my inputs and the corresponding output.

wwlorey:
Some context: that failure was happening because of a file lock problem (i.e. the file being uploaded was touched by some other process at the same time). Storage Explorer fails in that case as well. But we want to continue to upload the rest of the selected files even if one fails.

PR #823, opened 2020-09-22
Continue AzCopy upload after an error occurs

wwlorey:
Fixes https://github.com/microsoft/vscode-azurestorage/issues/817

Here's a few examples of what this looks like in action:

#### Multiple failures while uploading
Screen Shot 2020-09-22 at 4 09 01 PM

#### A single failure while uploading
Screen Shot 2020-09-22 at 4 09 41 PM

### DIFF
--- a/src/commands/azCopy/azCopyTransfer.ts
+++ b/src/commands/azCopy/azCopyTransfer.ts
@@ -33,7 +33,7 @@ export async function azCopyTransfer(
         // tslint:disable-next-line: strict-boolean-expressions
         let message: string = localize('azCopyTransfer', 'AzCopy Transfer: "{0}".', finalTransferStatus?.JobStatus || 'Failed');
         if (finalTransferStatus?.FailedTransfers?.length || finalTransferStatus?.SkippedTransfers?.length) {
-            message += localize('checkOutputWindow', ' Check the output window for a list of incomplete transfers.');
+            message += localize('checkOutputWindow', ' Check the [output window](command:{0}) for a list of incomplete transfers.', `${ext.prefix}.showOutputChannel`);
 
             if (finalTransferStatus.FailedTransfers?.length) {
                 ext.outputChannel.appendLog(localize('failedTransfers', 'Failed transfer(s):'));
--- a/src/commands/azCopy/azCopyTransfer.ts
+++ b/src/commands/azCopy/azCopyTransfer.ts
@@ -35,18 +35,14 @@ export async function azCopyTransfer(
         if (finalTransferStatus?.FailedTransfers?.length || finalTransferStatus?.SkippedTransfers?.length) {
             message += localize('checkOutputWindow', ' Check the output window for a list of incomplete transfers.');
 
-            // The optional "s" at the end of "transfer(s)"
-            let s: string;
             if (finalTransferStatus.FailedTransfers?.length) {
-                s = finalTransferStatus.FailedTransfers.length > 1 ? 's' : '';
-                ext.outputChannel.appendLog(localize('failedTransfers', 'Failed transfer{0}:', s));
+                ext.outputChannel.appendLog(localize('failedTransfers', 'Failed transfer(s):'));
                 for (let failedTransfer of finalTransferStatus.FailedTransfers) {
                     ext.outputChannel.appendLog(`\t${failedTransfer.Dst}`);
                 }
             }
             if (finalTransferStatus.SkippedTransfers?.length) {
-                s = finalTransferStatus.SkippedTransfers.length > 1 ? 's' : '';
-                ext.outputChannel.appendLog(localize('skippedTransfers', 'Skipped transfer{0}:', s));
+                ext.outputChannel.appendLog(localize('skippedTransfers', 'Skipped transfer(s):'));
                 for (let skippedTransfer of finalTransferStatus.SkippedTransfers) {
                     ext.outputChannel.appendLog(`\t${skippedTransfer.Dst}`);
                 }
--- a/src/commands/azCopy/azCopyTransfer.ts
+++ b/src/commands/azCopy/azCopyTransfer.ts
@@ -35,14 +35,18 @@ export async function azCopyTransfer(
         if (finalTransferStatus?.FailedTransfers?.length || finalTransferStatus?.SkippedTransfers?.length) {
             message += localize('checkOutputWindow', ' Check the output window for a list of incomplete transfers.');
 
+            // The optional "s" at the end of "transfer(s)"
+            let s: string;
             if (finalTransferStatus.FailedTransfers?.length) {
-                ext.outputChannel.appendLog(localize('failedTransfers', 'Failed transfers:'));
+                s = finalTransferStatus.FailedTransfers.length > 1 ? 's' : '';
+                ext.outputChannel.appendLog(localize('failedTransfers', 'Failed transfer{0}:', s));
                 for (let failedTransfer of finalTransferStatus.FailedTransfers) {
                     ext.outputChannel.appendLog(`\t${failedTransfer.Dst}`);
                 }
             }
             if (finalTransferStatus.SkippedTransfers?.length) {
-                ext.outputChannel.appendLog(localize('skippedTransfers', 'Skipped transfers:'));
+                s = finalTransferStatus.SkippedTransfers.length > 1 ? 's' : '';
+                ext.outputChannel.appendLog(localize('skippedTransfers', 'Skipped transfer{0}:', s));
                 for (let skippedTransfer of finalTransferStatus.SkippedTransfers) {
                     ext.outputChannel.appendLog(`\t${skippedTransfer.Dst}`);
                 }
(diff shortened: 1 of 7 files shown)


src/commands/uploadFiles.ts
```diff
@@ -28,8 +28,8 @@ export async function uploadFiles(
     notificationProgress?: NotificationProgress,
     cancellationToken?: CancellationToken,
     destinationDirectory?: string
-): Promise<void> {
-    let shouldCheckUris: boolean = false;
+): Promise<IParsedError[]> {
+    const calledFromUploadToAzureStorage: boolean = uris !== undefined;
```
nturinski, line 32: You can move this closer to where it's being used now.

src/commands/uploadFiles.ts
```diff
@@ -28,8 +28,8 @@ export async function uploadFiles(
     notificationProgress?: NotificationProgress,
     cancellationToken?: CancellationToken,
     destinationDirectory?: string
-): Promise<void> {
-    let shouldCheckUris: boolean = false;
+): Promise<IParsedError[]> {
```
nturinski, line 31: Hmm, I think this return type is pretty unexpected.  Like as an internal dev, I understand why you're returning `IParsedError[]`, but it seems really unintuitive from an outside perspective.

Maybe we could return a "resolution" object that is like the response of the transaction?  Maybe simulate what the API response is from AzCopy?  I could see it being something like:

```
successfulTransfers: Uri[]
skippedTransfers: Uri[]
failedTransfesr: Uri[]
errors: IParsedError[]
```

Or something like that?

src/commands/uploadFiles.ts
```diff
@@ -83,21 +82,44 @@ async function uploadFilesHelper(
...
+    if (calledFromUploadToAzureStorage) {
+        // `uploadToAzureStorage` will deal with any errors
+        return errors;
+    } else if (errors.length === 1) {
+        throw errors[0];
+    } else if (errors.length > 1) {
+        throw new Error(multipleAzCopyErrorsMessage);
+    } else {
+        // No errors
+        return [];
```
nturinski, line 123: Should be able to just return errors then (since it should also be an empty array)

src/commands/uploadFolder.ts
```diff
@@ -29,30 +30,40 @@ export async function uploadFolder(
...
+        if (notificationProgress && cancellationToken) {
+            // AzCopy recognizes folders as a resource when uploading to file shares. So only set `countFoldersAsResources=true` in that case
+            await uploadLocalFolder(actionContext, treeItem, sourcePath, destPath, notificationProgress, cancellationToken, destPath, treeItem instanceof FileShareTreeItem);
+        } else {
+            const title: string = getUploadingMessageWithSource(sourcePath, treeItem.label);
+            await vscode.window.withProgress({ cancellable: true, location: vscode.ProgressLocation.Notification, title }, async (newNotificationProgress, newCancellationToken) => {
+                await uploadLocalFolder(actionContext, nonNullValue(treeItem), sourcePath, nonNullValue(destPath), newNotificationProgress, newCancellationToken, destPath, treeItem instanceof FileShareTreeItem);
+            });
+        }
+    } catch (error) {
```
nturinski, line 57: uploadFolder is only called when there is one folder, right?  Multiple folders would use uploadFiles?

src/utils/errorUtils.ts
```diff
@@ -4,7 +4,10 @@
  *--------------------------------------------------------------------------------------------*/
 
 import * as vscode from 'vscode';
-import { TelemetryProperties, UserCancelledError } from 'vscode-azureextensionui';
+import { IParsedError, TelemetryProperties, UserCancelledError } from 'vscode-azureextensionui';
+import { localize } from './localize';
+
+export const multipleAzCopyErrorsMessage: string = localize('multipleAzCopyErrors', 'Multiple AzCopy errors occurred while uploading. Check the output window for more details.');
```
nturinski, line 10: Would be worth linking output window text to the output channel.  Here's an example. https://github.com/microsoft/vscode-azuretools/blob/main/appservice/src/deploy/deploy.ts#L70

src/commands/azCopy/azCopyTransfer.ts
```diff
@@ -35,14 +35,18 @@ export async function azCopyTransfer(
         if (finalTransferStatus?.FailedTransfers?.length || finalTransferStatus?.SkippedTransfers?.length) {
             message += localize('checkOutputWindow', ' Check the output window for a list of incomplete transfers.');
 
+            // The optional "s" at the end of "transfer(s)"
+            let s: string;
```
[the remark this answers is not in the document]
wwlorey: Hahaha I'll just change it to transfer(s) to keep it simple 😂🤡 Not sure why I didn't do that in the first place...

src/commands/uploadFolder.ts
wwlorey: `uploadFolder` is called repeatedly by `uploadToAzureStorage` if there's multiple folders

src/commands/uploadFiles.ts
wwlorey: Since we prompt for `uris` right after this declaration, I gotta keep it where it is so we know if any URIs were passed into the function. I originally moved it down to where its first used and was confused when no errors were thrown lol

src/commands/uploadFiles.ts
wwlorey: I added `IAzCopyResolution` which has an `errors` property to hopefully make this more intuitive. Unless there's a reason we need it in `uploadFiles.ts`, I didn't add info about the failed/skipped/succeeded transfers since that is handled in `azCopyTransfer`. It would take some refactoring to bubble that info up

src/commands/uploadFiles.ts
```diff
@@ -83,21 +83,41 @@ async function uploadFilesHelper(
     uris: Uri[],
     notificationProgress: NotificationProgress,
     cancellationToken: CancellationToken,
-    destinationDirectory: string
-): Promise<void> {
+    destinationDirectory: string,
+    calledFromUploadToAzureStorage: boolean
+): Promise<IAzCopyResolution> {
```
nturinski, line 88: In this case, the resolution isn't always from AzCopy.  Maybe we can just call it something generic, like TransferResolution or something?

src/commands/uploadFolder.ts
```diff
@@ -29,30 +31,41 @@ export async function uploadFolder(
...
-        await uploadLocalFolder(actionContext, treeItem, sourcePath, destPath, notificationProgress, cancellationToken, destPath, treeItem instanceof FileShareTreeItem);
-    } else {
-        const title: string = getUploadingMessageWithSource(sourcePath, treeItem.label);
-        await vscode.window.withProgress({ cancellable: true, location: vscode.ProgressLocation.Notification, title }, async (newNotificationProgress, newCancellationToken) => {
-            await uploadLocalFolder(actionContext, nonNullValue(treeItem), sourcePath, nonNullValue(destPath), newNotificationProgress, newCancellationToken, destPath, treeItem instanceof FileShareTreeItem);
-        });
+    try {
+        if (notificationProgress && cancellationToken) {
+            // AzCopy recognizes folders as a resource when uploading to file shares. So only set `countFoldersAsResources=true` in that case
+            await uploadLocalFolder(actionContext, treeItem, sourcePath, destPath, notificationProgress, cancellationToken, destPath, treeItem instanceof FileShareTreeItem);
```
nturinski, line 52: Would it make more sense to put the try/catch in `uploadLocalFolder` since that seems to be what we expect to throw the error in both the if and else case?

wwlorey: I think it would be more intrusive to make that change since we'd have to pass `calledFromUploadToAzureStorage` into `uploadLocalFolder` and also change the return type to `IAzCopyResolution`

nturinski: True enough, though I think we would want all transfer actions to return some sort of resolution eventually.  But right now, I think it's fine as is.

src/commands/azCopy/azCopyTransfer.ts
```diff
@@ -33,7 +33,7 @@ export async function azCopyTransfer(
         // tslint:disable-next-line: strict-boolean-expressions
         let message: string = localize('azCopyTransfer', 'AzCopy Transfer: "{0}".', finalTransferStatus?.JobStatus || 'Failed');
         if (finalTransferStatus?.FailedTransfers?.length || finalTransferStatus?.SkippedTransfers?.length) {
-            message += localize('checkOutputWindow', ' Check the output window for a list of incomplete transfers.');
+            message += localize('checkOutputWindow', ' Check the [output window](command:{0}) for a list of incomplete transfers.', `${ext.prefix}.showOutputChannel`);
```
nturinski, line 36: For reference, the only reason why we use ext.prefix in the tools package is because it's being used by App Service and Functions, so we have to use the registered prefix to open the appropriate channel.

nturinski (approved):
I think that we should eventually move most of this logic in `azCopyTransfer`.  In doing so, we can consolidate a lot of this logic and just return the errors, failed/skipped transfers, etc and let each function handle the errors however it needs to.

This will also let the logic be easily used by the download scenario, where we will have to use similar logic to handle a failed download in a multi-download situation.

But that refactor can wait until after the 🐛-fix

Issue #805, opened 2020-09-14
"Upload to Azure Storage" with expanded folder uploads twice


EricJizbaMSFT:
Repro steps:
1. From the file view, expand a folder
1. Select that folder, its contents, and a file at the root level. Something like this:
    <img width="423" alt="Screen Shot 2020-09-14 at 11 12 26 AM" src="https://user-images.githubusercontent.com/11282622/93122432-4c50de80-f67b-11ea-91d7-e46e00437558.png">
1. Right click and upload to storage

Expected:
It uploads with the same folder structure as local

Actual:
The files inside "TimerTrigger1" are uploaded twice - once under a "TimerTrigger1" path and once at the root.
<img width="236" alt="Screen Shot 2020-09-14 at 11 13 59 AM" src="https://user-images.githubusercontent.com/11282622/93122521-7bffe680-f67b-11ea-9529-186d8ee4edb0.png">

wwlorey:
The URIs passed into `uploadToAzureStorage` from a selection like that include the folder URI as well as each individual file because they were all selected. So the uploaded folder and files are accurate.

Changing this behavior would involve detecting when the selected files are the same as the children of the selected folder. And skipping the file uploads

EricJizbaMSFT:
I wouldn't call this "accurate". Users don't care how VS Code gives us the uris (that's just an engineering detail). Plus - we're the ones picking the remote path, so we still have some control/blame here. I would model our behavior off of the local "Copy" experience within VS Code.

PR #825, opened 2020-09-23
Prevent double-upload when a folder and its contents are selected

wwlorey:
Fixes https://github.com/microsoft/vscode-azurestorage/issues/805

### DIFF
--- a/src/commands/uploadToAzureStorage.ts
+++ b/src/commands/uploadToAzureStorage.ts
@@ -4,13 +4,13 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { stat } from 'fs-extra';
-import { dirname } from 'path';
 import * as vscode from 'vscode';
 import { IActionContext } from 'vscode-azureextensionui';
 import { ext } from '../extensionVariables';
 import { BlobContainerTreeItem } from '../tree/blob/BlobContainerTreeItem';
 import { FileShareTreeItem } from '../tree/fileShare/FileShareTreeItem';
 import { throwIfCanceled } from '../utils/errorUtils';
+import { isSubpath } from '../utils/fs';
 import { localize } from '../utils/localize';
 import { getUploadingMessage, OverwriteChoice, promptForDestinationDirectory, shouldUploadUri } from '../utils/uploadUtils';
 import { uploadFiles } from './uploadFiles';
@@ -21,7 +21,6 @@ export async function uploadToAzureStorage(actionContext: IActionContext, _first
     const destinationDirectory: string = await promptForDestinationDirectory();
     let fileUris: vscode.Uri[] = [];
     let folderUris: vscode.Uri[] = [];
-    const folderPathSet: Set<string> = new Set();
     let overwriteChoice: { choice: OverwriteChoice | undefined } = { choice: undefined };
 
     for (const uri of uris) {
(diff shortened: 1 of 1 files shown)


src/commands/uploadToAzureStorage.ts
```diff
@@ -29,12 +31,21 @@ export async function uploadToAzureStorage(actionContext: IActionContext, _first
...
             if ((await stat(uri.fsPath)).isDirectory()) {
                 folderUris.push(uri);
+                folderPathSet.add(uri.fsPath);
             } else {
                 fileUris.push(uri);
             }
         }
     }
 
+    fileUris = fileUris.filter(fileUri => {
```
EricJizbaMSFT, line 41: What about sub folders? Does that have the same problem where it's uploaded twice? If so, you would need to filter `folderUris` as well

src/commands/uploadToAzureStorage.ts
```diff
@@ -29,12 +31,21 @@ export async function uploadToAzureStorage(actionContext: IActionContext, _first
...
+        if (folderPathSet.has(dirname(fileUri.fsPath))) {
```
EricJizbaMSFT, line 42: Normally we just return a boolean in the `filter` method. Should make it a one-liner. Something like
```
    // This file's containing folder is already being uploaded. So don't upload this file.
    fileUris = fileUris.filter(fileUri => !folderPathSet.has(dirname(fileUri.fsPath)));
```

src/commands/uploadToAzureStorage.ts
```diff
@@ -30,12 +32,17 @@ export async function uploadToAzureStorage(actionContext: IActionContext, _first
...
                 folderUris.push(uri);
+                folderPathSet.add(uri.fsPath);
             } else {
                 fileUris.push(uri);
             }
         }
     }
 
+    // Only upload files and folders if their containing folder isn't already being uploaded.
+    fileUris = fileUris.filter(fileUri => !folderPathSet.has(dirname(fileUri.fsPath)));
```
EricJizbaMSFT, line 43: If I have a directory structure like this:
Screen Shot 2020-09-24 at 12 28 27 PM

Is it possible to select folder "a" and file "b.test.txt" _without_ selecting folder "b"? If so, `b.test.txt` might be uploaded twice because `a/b` technically isn't in `folderPathSet`. That being said, I couldn't figure out a way to make that happen, so I'll go ahead and approve as-is

EricJizbaMSFT (approved):

src/commands/uploadToAzureStorage.ts
wwlorey: That is possible if you <kbd>Cmd</kbd>+<kbd>click</kbd>

Screen Shot 2020-09-24 at 12 40 41 PM

And `b.test.txt` gets double uploaded :(

I could fix this by checking all the parent folders of a file when filtering out files. But I'm not sure if that extra work is worth it since this may not be a common scenario. What do you think?

EricJizbaMSFT: Yeah I think it's worth fixing

This util could be helpful:
https://github.com/microsoft/vscode-azureappservice/blob/b4cec5cc748e23833e191645ea295e3cc0d7f28a/src/utils/pathUtils.ts#L9

EricJizbaMSFT (approved):

Issue #712, opened 2020-07-22
The blob file which created from Azure view doesn't exist in "File Explorer"


v-ruizh:
**OS:** Windows 10
**Build Version:** 20200721.3

**Repro Steps:**
1. Expand one blob container and click "Open in File Explorer", it will open it in a new VS Code instance.
2. Create a file from "File Explorer" then refresh the blob container, the file shows in blob container.
3. Create empty blob or upload a file "one" to blob container from Azure view, then refresh the "File Explorer".
4. Check the result.

**Expect:**
The file "one" shows in "File Explorer".

**Actual:**
There is no file "one" shows in "File Explorer".
![image](https://user-images.githubusercontent.com/45614674/88149808-271f9100-cc33-11ea-8824-e910046333e2.png)

wwlorey:
As Eric said [here](https://github.com/microsoft/vscode-azurestorage/pull/840#discussion_r499943538), its doubtful that the file explorer and azure view would be used at the same time

PR #840, opened 2020-10-05
Refresh directory tree item when refreshing the file explorer

wwlorey:
Fixes https://github.com/microsoft/vscode-azurestorage/issues/712

When you open the file explorer in a new window and create a blob in a different window, that blob won't show up in the file explorer because we weren't refreshing the tree item. We were just loading what children existed when the file explorer was opened. So a refresh before calling `treeItem.loadAllChildren` fixes it

### DIFF
--- a/src/AzureStorageFS.ts
+++ b/src/AzureStorageFS.ts
@@ -102,6 +102,9 @@ export class AzureStorageFS implements vscode.FileSystemProvider, vscode.TextDoc
             context.telemetry.suppressIfSuccessful = true;
             let treeItem: AzureStorageDirectoryTreeItem = await this.lookupAsDirectory(uri, context);
 
+            // https://github.com/microsoft/vscode-azurestorage/issues/712
+            await treeItem.refresh();
+
             const loadingMessage: string = localize('loadingDir', 'Loading directory "{0}"...', treeItem.label);
             let children: AzExtTreeItem[] = await treeItem.loadAllChildren({ ...context, loadingMessage });
             let result: [string, vscode.FileType][] = [];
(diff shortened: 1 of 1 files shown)


src/AzureStorageFS.ts
```diff
@@ -102,6 +102,9 @@ export class AzureStorageFS implements vscode.FileSystemProvider, vscode.TextDoc
             context.telemetry.suppressIfSuccessful = true;
             let treeItem: AzureStorageDirectoryTreeItem = await this.lookupAsDirectory(uri, context);
 
+            // https://github.com/microsoft/vscode-azurestorage/issues/712
```
nturinski, line 105: We typically only leave the issue url if it's a weird work-around we're implementing to get around the bug (but intend to fix it when the fix or bandwidth exists).

I think in this case, it's fine to just refresh the tree. You can leave a comment about refreshing the treeItem to make sure that it's up to date since changes can happen outside of the file explorer, but I don't think it's entirely necessary.

nturinski (approved):

src/AzureStorageFS.ts
```diff
@@ -101,6 +101,7 @@ export class AzureStorageFS implements vscode.FileSystemProvider, vscode.TextDoc
         return await callWithTelemetryAndErrorHandling('readDirectory', async (context) => {
             context.telemetry.suppressIfSuccessful = true;
             let treeItem: AzureStorageDirectoryTreeItem = await this.lookupAsDirectory(uri, context);
+            await treeItem.refresh();
```
EricJizbaMSFT, line 104: Seems like you've been throwing around a lot of `treeItem.refresh();` to fix some bugs lately, but you have to be more careful with it. It will clear the cache of child tree items, which can directly affect performance if there's a bunch of files and if VS Code is calling `readDirectory` a lot. The cache is especially important in the file system because we always do load**All**Children by default, as opposed to the user having to manually click "Load More..."

For this specific bug, I suggest one of two options:
1. Fix it in a more precise manner (aka instead of completely clearing the cache, surgically update the existing cache and make sure the UI reflects that change)
1. "won't-fix" the bug. I don't think there's much value in keeping the Azure/File views in-sync. I expect users to use one or the other, not both.

Issue #795, opened 2020-09-11
Support "Copy URL" and "Delete..." for blob directories


v-mengwe:
**OS:** All
**Build Version:** 20200911.6

**Repro Steps:**

1. Upload a folder to a blob container --> Right click the folder --> Check the context menu.
2. Right click the file under the folder --> Check the context menu.

**Expect:**
Keep the context menu consistently between right click the folder and file.

**Actual:**
The context menu is inconsistent between right click the folder and file.
![image](https://user-images.githubusercontent.com/62053665/92889885-35c34100-f449-11ea-98fe-06b7b93f07af.png)

PR #842, opened 2020-10-06
Add "Copy URL" and "Delete..." context menu items for blob directories

wwlorey:
Fixes https://github.com/microsoft/vscode-azurestorage/issues/795

### DIFF
--- a/package.json
+++ b/package.json
@@ -514,14 +514,15 @@
                     "group": "3_download"
                 },
                 {
-                    "command": "azureStorage.download",
-                    "when": "view == azureStorage && viewItem == azureBlobDirectory",
-                    "group": "3_download"
-                },
-                {
                     "command": "azureStorage.copyUrl",
                     "when": "view == azureStorage && viewItem == azureBlob",
                     "group": "5_cutcopypaste"
+                },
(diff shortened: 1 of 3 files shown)


src/tree/blob/BlobDirectoryTreeItem.ts
```diff
@@ -74,6 +74,14 @@ export class BlobDirectoryTreeItem extends AzureParentTreeItem<IStorageRoot> {
         return child;
     }
 
+    public async copyUrl(): Promise<void> {
```
EricJizbaMSFT, line 77: This class should now have ` implements ICopyUrl {`

package.json
```diff
@@ -224,6 +224,11 @@
                 "title": "Delete...",
                 "category": "Azure Storage"
             },
+            {
+                "command": "azureStorage.deleteBlobDirectory",
```
EricJizbaMSFT, line 228: Since you're not supporting the command palette, you should do the `"when": "never"` thing in the "commandPalette" section of this file. "deleteBlob" already does it

EricJizbaMSFT (approved):

PR #851, opened 2020-10-14
Refactor `shouldUploadUri` to be useful for multi-downloads

wwlorey:
Related to https://github.com/microsoft/vscode-azurestorage/issues/788

### DIFF
--- a/src/commands/uploadToAzureStorage.ts
+++ b/src/commands/uploadToAzureStorage.ts
@@ -12,7 +12,7 @@ import { FileShareTreeItem } from '../tree/fileShare/FileShareTreeItem';
 import { multipleAzCopyErrorsMessage, throwIfCanceled } from '../utils/errorUtils';
 import { isSubpath } from '../utils/fs';
 import { localize } from '../utils/localize';
-import { checkCanOverwrite, convertLocalPathToRemotePath, getUploadingMessage, OverwriteChoice, promptForDestinationDirectory, remoteResourceExists } from '../utils/uploadUtils';
+import { checkCanUpload, convertLocalPathToRemotePath, getUploadingMessage, OverwriteChoice, promptForDestinationDirectory } from '../utils/uploadUtils';
 import { uploadFiles } from './uploadFiles';
 import { uploadFolder } from './uploadFolder';
 
@@ -50,7 +50,7 @@ export async function uploadToAzureStorage(actionContext: IActionContext, _first
         }
 
         const destPath: string = convertLocalPathToRemotePath(folderUri.fsPath, destinationDirectory);
-        if (!hasParent && await checkCanOverwrite(destPath, remoteResourceExists, overwriteChoice, treeItem)) {
+        if (!hasParent && await checkCanUpload(destPath, overwriteChoice, treeItem)) {
             folderUrisToUpload.push(folderUri);
         }
     }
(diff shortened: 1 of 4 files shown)


src/utils/uploadUtils.ts
```diff
@@ -68,27 +68,23 @@ async function showDuplicateResourceWarning(resourceName: string): Promise<Overw
 }
 
 // Pass `overwriteChoice` as an object to make use of pass by reference.
-export async function shouldUploadUri(treeItem: BlobContainerTreeItem | FileShareTreeItem, uri: vscode.Uri, overwriteChoice: { choice: OverwriteChoice | undefined }, destinationDirectory: string): Promise<boolean> {
+export async function checkCanOverwrite(
```
wwlorey, line 71: I'll move this function to a more suitable place in a different PR

src/utils/uploadUtils.ts
```diff
@@ -68,27 +68,23 @@ async function showDuplicateResourceWarning(resourceName: string): Promise<Overw
...
+    destPath: string,
+    destPathExists: (destPath: string, destTreeItem?: BlobContainerTreeItem | FileShareTreeItem) => Promise<boolean>,
```
wwlorey, line 73: I made `destTreeItem` optional since `destPathExists` for downloads will only involve looking at the local file system

EricJizbaMSFT: I like where you're going here, but I think it could be a little better. As you say, there's a problem here because `checkCanOverwrite` is supposed to be agnostic r.e. upload/download, but you have to add parameters specific to the upload case (destTreeItem)

Instead, I propose something more generic like this:
```typescript
export async function checkCanOverwrite(
    destPath: string,
    overwriteChoice: { choice: OverwriteChoice | undefined },
    destPathExists: () => Promise<boolean> // NOTE: I like to put callback parameters last just because of the formatting 
): Promise<boolean> {
```

And then instead of defining a `remoteResourceExists` function and passing it in every time you call `checkCanOverwrite`, you define a `checkCanUpload` function that contains the logic of `remoteResourceExists`:
```typescript
export async function checkCanUpload(
    destPath: string,
    overwriteChoice: { choice: OverwriteChoice | undefined },
    treeItem: BlobContainerTreeItem | FileShareTreeItem
): Promise<boolean> {
    return await checkCanOverwrite(destPath, overwriteChoice, async () => {
        if (treeItem instanceof BlobContainerTreeItem) {
            return await doesBlobExist(treeItem, destPath) || await doesBlobDirectoryExist(treeItem, destPath);
        } else {
            return await doesFileExist(basename(destPath), treeItem, dirname(destPath), treeItem.shareName) || await doesDirectoryExist(treeItem, destPath, treeItem.shareName);
        }
       // NOTE: You'll notice I don't have to throw an error here because `treeItem` is no longer optional
    });
}
```

src/commands/uploadToAzureStorage.ts
```diff
@@ -36,6 +36,7 @@ export async function uploadToAzureStorage(actionContext: IActionContext, _first
 
     let hasParent: boolean;
     let overwriteChoice: { choice: OverwriteChoice | undefined } = { choice: undefined };
+    let destPath: string;
```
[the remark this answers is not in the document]
wwlorey: Are there any problems with re-defining variables in a for loop like in this case? I feel like I remember learning to avoid that when I worked with lower level languages but it may not matter in this case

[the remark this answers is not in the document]
EricJizbaMSFT: Maybe this is what you're referring to:
https://softwareengineering.stackexchange.com/questions/296721/is-it-good-to-define-a-variable-inside-a-loop/296724

EricJizbaMSFT (approved):

Issue #666, opened 2020-05-09
Better handle "Files above 50MB cannot be synchronized with extensions" error


v-mengwe:
**OS:** Win10
**Build Version:** [20200508.1](https://dev.azure.com/ms-azuretools/AzCode/_build/results?buildId=18973&view=artifacts&type=publishedArtifacts)

**Repro Steps:**
1. Left click on a text file larger than 4MB
2. Verify the extension tells you to use Storage Explorer

**Expect:**
The extension tells you to use Storage Explorer, such as:
![image](https://user-images.githubusercontent.com/62053665/81470935-4c9a2280-9220-11ea-8219-80f71fb6dd73.png)

**Actual:**
Another error notification pops up.
![image](https://user-images.githubusercontent.com/62053665/81469963-bca5aa00-921a-11ea-9d93-fd18ac388c3e.png)

wwlorey:
We don't have the 4MB limitation anymore. But VS Code caps file size for editing files at 50MB: https://github.com/microsoft/vscode/blob/3b99352509cb2bf0cfc65fc567cea100e4a8a279/src/vs/workbench/api/browser/mainThreadDocuments.ts#L255

PR #852, opened 2020-10-14
Throw a better error when trying to edit a blob/file larger than 50MB

wwlorey:
Fixes https://github.com/microsoft/vscode-azurestorage/issues/666 😈

### DIFF
--- a/src/AzureStorageFS.ts
+++ b/src/AzureStorageFS.ts
@@ -3,12 +3,14 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import { BlobClient, BlobDownloadResponseModel } from "@azure/storage-blob";
-import { FileDownloadResponseModel, ShareFileClient } from "@azure/storage-file-share";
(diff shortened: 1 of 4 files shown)


wwlorey:
Here's the new error message:

Screen Shot 2020-10-14 at 12 44 04 PM

src/AzureStorageFS.ts
```diff
@@ -54,9 +56,31 @@ export class AzureStorageFS implements vscode.FileSystemProvider, vscode.TextDoc
     }
 
     static async showEditor(treeItem: BlobTreeItem | FileTreeItem): Promise<void> {
-        const uri = this.idToUri(treeItem.fullId);
-        let doc = await vscode.workspace.openTextDocument(uri);
-        await vscode.window.showTextDocument(doc, { preserveFocus: true, preview: false });
+        return await callWithTelemetryAndErrorHandling('showEditor', async (context) => {
```
EricJizbaMSFT, line 59: You should pass in an existing `IActionContext` instead of creating a new telemetry event

EricJizbaMSFT (approved):

Issue #219, opened 2018-08-21
Error: server failed to authenticate


omartin2010:
Issue Type: <b>Bug</b>

I wish I knew how to repro but I don't. Sometimes I get the error saying Error: Server failed to authenticate the request. Make sure the value of Authorization header is formed correctly including the signature. RequestId:.... Time: 2018-08-21T14...
![image](https://user-images.githubusercontent.com/11371870/44407927-45535280-a52d-11e8-92eb-240658bd7d81.png)

Extension version: 0.4.1
VS Code version: Code 1.26.1 (493869ee8e8a846b0855873886fc79d480d342de, 2018-08-16T18:38:57.434Z)
OS version: Windows_NT x64 10.0.17134

<details>
<summary>System Info</summary>

|Item|Value|
|---|---|
|CPUs|Intel(R) Core(TM) i7-8650U CPU @ 1.90GHz (8 x 2112)|
|GPU Status|2d_canvas: enabled<br>checker_imaging: disabled_off<br>flash_3d: enabled<br>flash_stage3d: enabled<br>flash_stage3d_baseline: enabled<br>gpu_compositing: enabled<br>multiple_raster_threads: enabled_on<br>native_gpu_memory_buffers: disabled_software<br>rasterization: enabled<br>video_decode: enabled<br>video_encode: enabled<br>webgl: enabled<br>webgl2: enabled|
|Memory (System)|15.93GB (7.80GB free)|
|Process Argv|C:\Users\omartin\AppData\Local\Programs\Microsoft VS Code\Code.exe|
|Screen Reader|no|
|VM|0%|

</details>

omartin2010:
BTW, I see this on every one of my storage accounts, and in 3 subscriptions I have access to (my view was filtered to 2 subs, I've added one more to see and I'm seeing the same). I only see the behavior with Files, Queues and Tables. Not with blob. And I only have to click to expand the File Shares, for example to see that message appear. No need to drill down, in fact, I'm no longer able to drill down. I didn't restart VSCode yet but I suspect that would help. If you want to see a repro, ping me internally.

awyl:
I have the same issue.

StephenWeatherford:
@awyl Does it repro for blobs or only for Files/Queues/Tables?

awyl:
Files and Queues

StephenWeatherford:
[...]

PR #854, opened 2020-10-19
Add retry logic for deploy

wwlorey:
Fixes https://github.com/microsoft/vscode-azurestorage/issues/219


src/commands/storageAccountActionHandlers.ts
```diff
@@ -94,7 +94,19 @@ async function deployStaticWebsite(context: IActionContext, target?: vscode.Uri
 
     await runPreDeployTask(sourcePath, context);
 
-    return destContainerTreeItem.deployStaticWebsite(context, sourcePath);
+    let shouldRetry: boolean = false;
```
EricJizbaMSFT, line 97: Yes, you should definitely have a max number of attempts. Also, exponential retry would be best. Here's a few examples where I use the [p-retry](https://www.npmjs.com/package/p-retry) npm module, which defaults to exponential retry

https://github.com/microsoft/vscode-azurefunctions/blob/d1318cad6809dbe65029dfe2012c76a1b17d0df5/src/commands/deploy/notifyDeployComplete.ts#L40

https://github.com/microsoft/vscode-azuretools/blob/5999c2ad4423e86f22d2c648027242d8816a50e4/appservice/src/utils/kuduUtils.ts#L13

src/commands/storageAccountActionHandlers.ts
```diff
@@ -94,7 +96,18 @@ async function deployStaticWebsite(context: IActionContext, target?: vscode.Uri
...
-    return destContainerTreeItem.deployStaticWebsite(context, sourcePath);
+    const retries: number = 4;
+    await retry(
+        async (currentAttempt) => {
+            context.telemetry.properties.deployAttempt = currentAttempt.toString();
+            if (currentAttempt > 1) {
+                const message: string = localize('retryingDeploy', 'Retrying deploy (Attempt {0}/{1})...', currentAttempt, retries + 1);
+                ext.outputChannel.appendLog(message);
+            }
+            await nonNullValue(destContainerTreeItem).deployStaticWebsite(context, nonNullValue(sourcePath));
```
EricJizbaMSFT, line 107: Before you used `p-retry`, you were only retrying for a specific error message, which is still possible with p-retry (I think with the `onFailedAttempt` option). Any particular reason you changed? If you were just following my example, one key difference is that I was retrying one specific API call at a time (as opposed to your logic, which wraps the whole "deployStaticWebsite" method). There are several types of errors we don't want to retry, including if the user cancels.

I suggest you reduce the amount of code that is inside "retry". Assuming you do that, I could go either way on whether you should only retry for specific errors

wwlorey: My bad - I wasn't sure if retrying for specific errors was possible. I added logic to only retry in the "server failed to authenticate" case.

I also moved the retry logic to `deployStaticWebsiteCore` and only retry listing blobs to delete, deleting blobs, and uploading blobs

src/tree/blob/BlobContainerTreeItem.ts
```diff
@@ -205,47 +207,65 @@ export class BlobContainerTreeItem extends AzureParentTreeItem<IStorageRoot> imp
...
-                let message = `The storage container "${this.friendlyContainerName}" contains ${blobsToDelete.length} files. Deploying will delete all of these existing files.  Continue?`;
-                let deleteAndDeploy: vscode.MessageItem = { title: 'Delete and Deploy' };
-                const result = await vscode.window.showWarningMessage(message, { modal: true }, deleteAndDeploy, DialogResponses.cancel);
-                if (result !== deleteAndDeploy) {
-                    context.telemetry.properties.cancelStep = 'AreYouSureYouWantToDeleteExistingBlobs';
-                    throw new UserCancelledError();
+        const retries: number = 4;
+        await retry(
+            async (currentAttempt) => {
+                context.telemetry.properties.deployAttempt = currentAttempt.toString();
```
EricJizbaMSFT, line 213: After we release, make sure to check telemetry to see if people are succeeding to deploy with `deployAttempt` at 2+ attempts. If you only see people succeeding at 1 attempt, we know the retry logic isn't helping and we should remove it because it just delays showing the error to the user

EricJizbaMSFT (approved):

src/tree/blob/BlobContainerTreeItem.ts
wwlorey: It looks like this retry is rarely used based on telemetry:

| EventName | EventCount | Properties_deployattempt |
| -- | -- | -- |
| vscode-azurestorage/azurestorage.deploystaticwebsite | 1 | 5 |
| vscode-azurestorage/azurestorage.deploystaticwebsite | 1 | 2 |
| vscode-azurestorage/azurestorage.deploystaticwebsite | 6611 | 1 |

@EricJizbaMSFT do you think it's worth keeping around?

EricJizbaMSFT: Uhh well that's so rare that it's not really hurting anything either. I would probably do more investigation before removing it. Maybe start by seeing how often the "server failed to authenticate" error happened in previous releases and try to understand why it's so rare in this release. Also, you might want to see if there's other errors where a retry might help

PR #857, opened 2020-10-20
Support downloading multiple blobs, files, and directories

wwlorey:
Fixes https://github.com/microsoft/vscode-azurestorage/issues/788

### DIFF
--- a/src/commands/downloadFile.ts
+++ b/src/commands/downloadFile.ts
@@ -6,9 +6,11 @@
 import { FromToOption, ILocalLocation, IRemoteSasLocation } from "@azure-tools/azcopy-node";
 import { BlockBlobClient } from "@azure/storage-blob";
 import { ShareFileClient } from "@azure/storage-file-share";
-import { extname, join } from "path";
-import { ProgressLocation, SaveDialogOptions, Uri, window } from "vscode";
-import { IActionContext, UserCancelledError } from "vscode-azureextensionui";
+import * as fse from 'fs-extra';
+import { join, posix } from "path";
+import { ProgressLocation, window } from "vscode";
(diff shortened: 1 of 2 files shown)


src/commands/downloadFile.ts
```diff
@@ -6,96 +6,151 @@
...
-    } else {
-        throw new UserCancelledError();
+
+    let hasParent: boolean;
+    let overwriteChoice: { choice: OverwriteChoice | undefined } = { choice: undefined };
+    const foldersToDownload: IAzCopyDownload[] = [];
+    const filesToDownload: IAzCopyDownload[] = [];
+
+    // Only download folders and files if their containing folder isn't already being downloaded.
+    for (const folderDownload of allFolderDownloads) {
```
wwlorey, line 123: I know this duplicate detection logic is pretty similar to the [upload](https://github.com/microsoft/vscode-azurestorage/blob/wwl/multi-down/src/commands/uploadToAzureStorage.ts#L43) version but I wasn't sure if I should invest the time combining the two. There are some differences here like how the `remoteFilePath` is used in `isSubPath` while the `localFilePath` is used in `checkForDownload`. The upload version uses the same path (`fsPath`) for everything.

nturinski: I think it's fine as is, but keep in mind when you're refactoring, you don't have to do an all or nothing approach.  For example, you could just factor out the parts that _are_ the same like these lines https://github.com/microsoft/vscode-azurestorage/pull/857/files?diff=split&w=1#diff-0b84588c3a1fc7afe2c6b7de2e41941feb6b43773dca3433067d01b109d8a8caR124-R129
and turn them into some util `hasParent` function for azCopy.

nturinski (approved):

Issue #677, opened 2020-05-20
Canceling the Download prompt of Storage Explorer


nturinski:
Does this occur consistently? 
Repro steps:

1. Don't have Storage Explorer installed
1. Right-click a resource and open in Storage Explorer
1. When it prompts you to download, click cancel

Action: azureStorage.openTable
Error type: ERR_INVALID_ARG_VALUE
Error Message: The argument 'file' cannot be empty. Received ''

Version: 0.9.0
OS: win32
OS Release: 10.0.18362
Product: Visual Studio Code
Product Version: 1.45.0
Language: en

<details>
<summary>Call Stack</summary>

```
normalizeSpawnArguments child_process.js:414:11
Object.spawn child_process.js:547:16
extension.bundle.js:157:116812extension.bundle.js:157:116812
Function.<anonymous> extension.bundle.js:157:116667
extension.bundle.js:157:116496extension.bundle.js:157:116496
i extension.bundle.js:157:116182
Function.launch extension.bundle.js:157:116627
Function.<anonymous> extension.bundle.js:398:812420
fulfilled extension.bundle.js:398:810204
processTicksAndRejections task_queues.js:85:5
```

</details>

PR #859, opened 2020-10-21
Don't throw an error when cancelling SE notification on Windows

wwlorey:
Fixes https://github.com/microsoft/vscode-azurestorage/issues/677

`callWithTelemetryAndErrorHandling` was swallowing the error thrown after x-ing out the notification


src/storageExplorerLauncher/windowsStorageExplorerLauncher.ts
```diff
@@ -33,39 +33,36 @@ export class WindowsStorageExplorerLauncher implements IStorageExplorerLauncher
...
+        if (regVal) {
+            // Parse from e.g.: "C:\Program Files (x86)\Microsoft Azure Storage Explorer\StorageExplorer.exe" -- "%1"
+            exePath = regVal.split("\"")[1];
+        }
+        if (exePath && await WindowsStorageExplorerLauncher.fileExists(exePath)) {
+            // tslint:disable-next-line:no-unsafe-finally // Grandfathered in
+            return exePath;
+        } else {
+            context.telemetry.properties.storageExplorerNotFound = 'true';
+            let selected: "Download" = <"Download">await vscode.window.showWarningMessage("Cannot find a compatible Storage Explorer. Would you like to download the latest Storage Explorer?", "Download");
```
nturinski, line 57: You should convert this to `ext.ui.showWarningMessage`

src/storageExplorerLauncher/windowsStorageExplorerLauncher.ts
```diff
@@ -95,8 +92,8 @@ export class WindowsStorageExplorerLauncher implements IStorageExplorerLauncher
         await Launcher.launch("cmd", "/c", "start", downloadPageUrl);
     }
 
-    private static async launchStorageExplorer(args: string[] = []): Promise<void> {
-        let storageExplorerExecutable = await WindowsStorageExplorerLauncher.getStorageExplorerExecutable();
+    private static async launchStorageExplorer(context: IActionContext, args: string[] = []): Promise<void> {
+        let storageExplorerExecutable = await WindowsStorageExplorerLauncher.getStorageExplorerExecutable(context);
```
nturinski, line 96: I think here you could do something like

```
if (!storageExplorerExecutable) {
}
```

That way you don't have to touch so many files.

nturinski (commented):
Could you not just rethrow the error instead of making all of these changes?

src/storageExplorerLauncher/windowsStorageExplorerLauncher.ts
```diff
@@ -55,17 +57,14 @@ export class WindowsStorageExplorerLauncher implements IStorageExplorerLauncher
...
+                const download: MessageItem = { title: localize('download', 'Download') };
+                const message: string = localize('cantFindSE', 'Cannot find a compatible Storage Explorer. Would you like to download the latest Storage Explorer?');
+                await ext.ui.showWarningMessage(message, download);
+                context.telemetry.properties.downloadStorageExplorer = 'true';
+                await WindowsStorageExplorerLauncher.downloadStorageExplorer();
+                return undefined;
             }
-            // tslint:disable-next-line: strict-boolean-expressions
-        }) || '';
+        });
```
nturinski, line 67: I think leaving this as it was and having the return type be only string might be better since this is kind of changing behavior a little bit.  Also, it's not consistent with the Mac function anymore.

nturinski (commented):
As far as I can tell, the Mac function has the exact implementation.  Any idea as to why there's no error thrown in that case?

wwlorey:
> As far as I can tell, the Mac function has the exact implementation. Any idea as to why there's no error thrown in that case?

If you cancel out of the notification on Mac, `getStorageExplorerExecutable` returns an empty string but [this line](https://github.com/microsoft/vscode-azurestorage/blob/wwl/se/src/storageExplorerLauncher/macOSStorageExplorerLauncher.ts#L78) doesn't throw an error in that case.

nturinski (approved):

nturinski:
Honestly we should probably throw a UserCancelError for the Mac scenario as well.  Even though it doesn't throw an error, we should avoid invoking the launcher even with empty parameters.  Who knows what kind of effects it might have.

Issue #846, opened 2020-10-12
No confirmation dialog pops up when deleting a blob directory


v-mengwe:
**OS:** All
**Build Version:** 20201011.1

**Repro Steps:**

1. Upload a folder to a blob container --> Right click the folder --> Select "Delete...".
2. Check the result.

**Expect:**
A confirmation dialog "Are you sure you want to delete the blob 'directory name' and all of its files and 
subdirectories? “ pops up.

**Actual:**
No confirmation dialog pops up.

PR #861, opened 2020-10-21
Add delete confirmation message for blob directories

wwlorey:
Fixes https://github.com/microsoft/vscode-azurestorage/issues/846

### DIFF
--- a/src/tree/blob/BlobDirectoryTreeItem.ts
+++ b/src/tree/blob/BlobDirectoryTreeItem.ts
@@ -6,11 +6,12 @@
 import * as azureStorageBlob from "@azure/storage-blob";
 import * as path from 'path';
 import * as vscode from 'vscode';
-import { AzExtTreeItem, AzureParentTreeItem, IActionContext, ICreateChildImplContext, parseError } from "vscode-azureextensionui";
+import { AzExtTreeItem, AzureParentTreeItem, DialogResponses, IActionContext, ICreateChildImplContext, parseError } from "vscode-azureextensionui";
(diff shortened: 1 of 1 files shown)


src/tree/blob/BlobDirectoryTreeItem.ts
```diff
@@ -83,27 +84,34 @@ export class BlobDirectoryTreeItem extends AzureParentTreeItem<IStorageRoot> imp
...
-                throw new Error(`Errors occurred when deleting "${this.dirName}".`);
-            }
-        });
+    public async deleteTreeItemImpl(context: ISuppressMessageContext): Promise<void> {
+        let response: vscode.MessageItem | undefined;
+        if (!context.suppressMessage) {
+            const message: string = localize('deleteBlobDir', "Are you sure you want to delete the blob directory '{0}' and all its contents?", this.label);
+            response = await ext.ui.showWarningMessage(message, { modal: true }, DialogResponses.deleteResponse, DialogResponses.cancel);
+        }
+        if (response === DialogResponses.deleteResponse || context.suppressMessage) {
```
nturinski, line 93: This check is unnecessary since `showWarningMessage` throws a userCancelledError if the user doesn't select deleteResponse.

wwlorey: Do we need it in this case since we're also checking if `context.suppressMessage` is true when deciding to delete?

wwlorey: Lol nevermind 😄

nturinski (approved):

Issue #659, opened 2020-03-31
Pop up an error when clicking "Load More..." from local emulator


v-dantch:
**OS:** Windows
**Build Version:** [20200330.1](https://dev.azure.com/ms-azuretools/AzCode/_build/results?buildId=16923&view=artifacts&type=publishedArtifacts)

**Repro Steps:**
1. Create a new blob container from Local Emulator.
2. Open a folder in VS Code --> Upload the folder in Step1 to the new created blob container from Local Emulator.
3. Back to Azure Storage View --> Drill down from Local Emulator --> Blob Containers --> The new created blob container and refresh it.
4. Click the "Load More...".

**Expect:**
Succeed to load more.

**Actual:**
Pop up an error:
Element with id /attachedStorageAccounts/devstoreaccount1/Blob Containers/test0/favicons-c2a605fbc0e687b2e1b4b90a7c445cdd is already registered.
![image](https://user-images.githubusercontent.com/34729092/77981470-a3b4f980-733c-11ea-9107-08dee5e06b39.png)

**More Info:**
There is no button on the error dialog.

wwlorey:
@v-dantch are you using the Azurite extension here or the CLI?

v-dantch:
Hi @wwlorey , yes, I have installed Azurite Extension and both Azurite Queue Service and Azurite Blob Service are running.
![image](https://user-images.githubusercontent.com/34729092/78202319-962b7b00-74c6-11ea-8ddb-d94470abc980.png)

Add more info:
The blob display empty after clicking "Load More...".
![image](https://user-images.githubusercontent.com/34729092/78202399-ca9f3700-74c6-11ea-83ba-324a995491df.png)

wwlorey:
Related: https://github.com/Azure/Azurite/issues/605

alexweininger:
Closing as stale

PR #870, opened 2020-10-26
Prevent duplicate tree items when loading more (emulator)

wwlorey:
When calling `listBlobsByHierarchy` with the local emulator, all the nested children of the first blob directory will be loaded before any other resources at that same level. This is different from the non-emulated case which lists all blobs and directories at one level without loading nested children.

Because the emulator is loading all nested children but their parent isn't expanded, only the parent is returned after clicking "Load More..." (assuming greater than `maxPageSize` children in the directory):

Screen Shot 2020-10-26 at 2 41 06 PM

This fix prevents duplicate tree items from showing up when loading more children from emulated blob containers. ~This means that clicking "Load More..." sometimes won't do anything as it loads the nested children.~

~Fixes~ Related: https://github.com/microsoft/vscode-azurestorage/issues/659

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3012,12 +3012,6 @@
             "resolved": "https://registry.npmjs.org/diagnostic-channel-publishers/-/diagnostic-channel-publishers-0.3.5.tgz",
             "integrity": "sha512-AOIjw4T7Nxl0G2BoBPhkQ6i7T4bUd9+xvdYizwvG7vVAM1dvr+SDrcUudlmzwH0kbEwdR2V1EcnKT0wAeYLQNQ=="
         },
-        "didyoumean": {
-            "version": "1.2.1",
-            "resolved": "https://registry.npmjs.org/didyoumean/-/didyoumean-1.2.1.tgz",
-            "integrity": "sha1-6S7f2tplN9SE1zwBcv0eugxJdv8=",
-            "dev": true
-        },
         "diff": {
             "version": "3.5.0",
             "resolved": "https://registry.npmjs.org/diff/-/diff-3.5.0.tgz",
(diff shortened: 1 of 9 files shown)


nturinski:
I'm not sure that having Load More do nothing is really acceptable behavior.  It makes it look like something is broken/unresponsive.  Maybe we can use load it again if we see that the array is empty after you do the filter?

wwlorey:
> I'm not sure that having Load More do nothing is really acceptable behavior. It makes it look like something is broken/unresponsive. Maybe we can use load it again if we see that the array is empty after you do the filter?

That's a good point - I added logic to keep loading until a new tree item loads

nturinski:
So this is adding a Load More button even when there aren't any children to load.  I have two storage accounts here with the exact same contents:

image


Furthermore, this skips the first folder and is loading one folder at a time now:
image

I think it might be better to investigate how to use the continuation token properly than this workaround.  It seems like the potential to cause a lot of bugs is pretty high.

nturinski:
Okay, I think I see the problem here.

When you call https://github.com/microsoft/vscode-azurestorage/blob/main/src/utils/blobUtils.ts#L40, this response is an iterator.  There's another property on `next` besides `value`, which is called `done`.

When you `listBlobsByHierarchy` an Azure blob, `done` is set to 'true' however, if you check on the emulator, it's marked as 'false'.
image

What this means is that there's actually more to the response and you have to keep invoking `next()` to get the remainder of the response. I tried it once and you'll notice that the prefixes are different:

image

So you'll need to iterate through the response until `done === true`.  I guess Azure was already taking care of this for us, but the emulator doesn't.

wwlorey:
Per standup discussion: it looks like this loading behavior is a problem with Azurite. I just pushed a temporary fix that should prevent the weird "load more" behavior unless people have large blob directories

nturinski (approved):

PR #878, opened 2020-11-02
Prep for 0.11.0 🦃🍂

wwlorey:


CHANGELOG.md
```diff
@@ -1,6 +1,15 @@
 # Change Log
 All notable changes to the "vscode-azurestorage" extension will be documented in this file.
 
+## 0.11.0 - 2020-11-02
+### Added
+- Support for downloading multiple blobs, files, and directories at a time
+- Added setting "azureStorage.deleteBeforeDeploy" to delete existing blobs before deploying to static website
```
EricJizbaMSFT, line 7: This makes it sound like previously we were _not_ deleting stuff. We should probably re-word it a little bit or just add something to the end like `(defaults to "true")`

EricJizbaMSFT (approved):

PR #881, opened 2020-11-10
Add missing AzCopy error message

wwlorey:
I'm not sure how I originally missed this but I did...

Here's an example error message:
Screen Shot 2020-11-09 at 4 35 13 PM

### DIFF
--- a/src/commands/azCopy/azCopyTransfer.ts
+++ b/src/commands/azCopy/azCopyTransfer.ts
@@ -43,7 +43,7 @@ async function handleJob(context: IActionContext, jobInfo: IJobInfo, transferLab
     context.telemetry.properties.jobStatus = finalTransferStatus?.JobStatus;
     if (!finalTransferStatus || finalTransferStatus.JobStatus !== 'Completed') {
         // tslint:disable-next-line: strict-boolean-expressions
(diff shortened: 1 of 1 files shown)


src/commands/azCopy/azCopyTransfer.ts
```diff
@@ -59,7 +59,7 @@ async function handleJob(context: IActionContext, jobInfo: IJobInfo, transferLab
                     ext.outputChannel.appendLog(`\t${skippedTransfer.Dst}`);
                 }
             }
-        } else {
+        } else if (!jobInfo.errorMessage) {
```
nturinski, line 62: I'm not sure if you should change this or not.  I'm not entirely familiar with the properties, but seems safe to just leave it as an else statement since we can assume it failed if the jobStatus isn't completed, right?

nturinski (approved):

src/commands/azCopy/azCopyTransfer.ts
wwlorey: I originally changed this because I was worried it would look cluttered but it's not too bad. Just removed this check.

Screen Shot 2020-11-10 at 1 05 19 PM

PR #885, opened 2020-11-18
Lazy load azure storage package

EricJizbaMSFT:
Related to https://github.com/microsoft/vscode-azuretools/pull/806

I don't actually think this will help startup performance as much as other extensions, but I want us to use this pattern in all of our extensions regardless. It could also help isolate azure stack changes inside of `createStorageClient`

### DIFF
--- a/package.json
+++ b/package.json
@@ -854,6 +854,7 @@
         "@types/p-retry": "^3.0.1",
         "@types/vscode": "1.48.0",
         "@types/winreg": "1.2.30",
+        "@azure/arm-resources": "^3.0.0",
         "copy-webpack-plugin": "^5.1.2",
         "gulp": "^4.0.0",
         "mocha": "^7.1.1",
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -854,6 +854,7 @@
         "@types/p-retry": "^3.0.1",
         "@types/vscode": "1.48.0",
         "@types/winreg": "1.2.30",
+        "@azure/arm-resources": "^3.0.0",
```
nturinski, line 857: Is this move related?  I don't recall you moving any arm packages into dev-dependencies and I'd assume that we're using more than the typings from these packages.

nturinski (approved):

package.json
EricJizbaMSFT: It's related in the sense that when I add `azureClients.ts` to a repo, I do it for all `@azure/` dependencies in package.json. I looked for all usages of `@azure/arm-resources`, only found one in `test/`, and thus moved it to `devDependencies`

PR #886, opened 2020-11-20
Add Support to Azure Stack in azure storage package

v-tsun:


src/tree/SubscriptionTreeItem.ts
```diff
@@ -34,26 +39,26 @@ export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
     }
 
     public async createChildImpl(context: ICreateChildImplContext): Promise<AzureTreeItem> {
-        const defaultLocation = 'westus';
+        const defaultLocation = this.isStack ? 'local' : 'westus';
```
rakku-ms, line 42: default location is not always 'local' for Azure Stack. You can get this value by parsing the ARM url. @bganapa

Issue #829, opened 2020-10-01
Folder and files aren't supposed to coexist but this can be circumvented via `Upload Files...`


nturinski:
Repro steps:

Only for blobs:
1. Create a file named "1234"
2. Upload Files...
3. When prompted for destination directory, type in "1234"
4. Files will upload, and it appears to be in the root
5. Refresh it, and you will have an invalid tree item
![image](https://user-images.githubusercontent.com/5290572/94861176-ef2f8980-03eb-11eb-9060-9701f4620dc7.png)
6. Open in file explorer, and it appears as a folder AND a file
![image](https://user-images.githubusercontent.com/5290572/94861208-feaed280-03eb-11eb-9065-eba8623fdf52.png)
7. ???
8. $$$$$PROFIT$$$$$

wwlorey:
Lets see what Storage Explorer does

wwlorey:
Storage explorer handles this gracefully 

<img width="673" alt="Screen Shot 2020-11-23 at 2 38 42 PM" src="https://user-images.githubusercontent.com/22795803/100012898-dc508a00-2d88-11eb-824a-2acb0c13c590.png">

nturinski:
Now that's hot 🔥

PR #893, opened 2020-12-01
Allow blobs and blob directories to coexist 😊

wwlorey:
Attempting to load a blob and directory with the same name would throw an error since their IDs were the same. This ensures that blob directory IDs will always be different from blob IDs.

Fixes https://github.com/microsoft/vscode-azurestorage/issues/829
EDIT: Depends on https://github.com/microsoft/vscode-azuretools/pull/828

### DIFF
--- a/src/tree/blob/BlobDirectoryTreeItem.ts
+++ b/src/tree/blob/BlobDirectoryTreeItem.ts
@@ -19,7 +19,6 @@ import { BlobTreeItem, ISuppressMessageContext } from "./BlobTreeItem";
 
 export class BlobDirectoryTreeItem extends AzureParentTreeItem<IStorageRoot> implements ICopyUrl {
     public static contextValue: string = 'azureBlobDirectory';
-    public static idSuffix: string = '-blobDirectory';
     public contextValue: string = BlobDirectoryTreeItem.contextValue;
     public iconPath: { light: string | vscode.Uri; dark: string | vscode.Uri } = {
         light: path.join(getResourcesPath(), 'light', 'folder.svg'),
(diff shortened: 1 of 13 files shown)


src/tree/blob/BlobDirectoryTreeItem.ts
```diff
@@ -51,6 +52,10 @@ export class BlobDirectoryTreeItem extends AzureParentTreeItem<IStorageRoot> imp
         return this.dirName;
     }
 
+    public get id(): string {
+        return `${this.dirName}${BlobDirectoryTreeItem.idSuffix}`;
```
ejizba, line 56: This part makes me the most nervous because we use id/fullId all over the place and it would be weird if it didn't match the azure resource id. I wonder if you could find a solution in [AzExtTreeDataProvider](https://github.com/microsoft/vscode-azuretools/blob/main/ui/src/tree/AzExtTreeDataProvider.ts) to kind of automatically prevent us from running into this problem without having to change the value of id/fullId.

Like, maybe if [this code](https://github.com/microsoft/vscode-azuretools/blob/main/ui/src/tree/AzExtTreeDataProvider.ts#L82) detects tree items with duplicate ids _but different classes_, it could add a temporary `uniqueFullId` property that includes the class name to use [here](https://github.com/microsoft/vscode-azuretools/blob/main/ui/src/tree/AzExtTreeDataProvider.ts#L46). As long as `uniqueFullId` is only used one or two places and not exposed outside the UI package, I feel like that could be less risky. Another option is to just _always_ append the class name to a `uniqueFullId` property - this means you wouldn't have to detect the case during `AzExtTreeDataProvider.getChildren`, but the ids we give to VS Code would be longer (not sure if that matters)

wwlorey:
I got rid of the ID changes here. The logic to make sure the right resource type is opened didn't change

ejizba (approved):

Issue #140, opened 2018-05-10
[Linux]Clicking 'Open Blob Container in Storage Explorer' pops up a warning when Storage Explorer is opened


v-xuanzh:
Repro Steps:
1. Make sure Storage Explorer is opened.
2. Right click one blob container and select 'Open Blob Container in Storage Explorer'.

Expect:
The selected blob container can be opened in Storage Explorer successfully.

Actual:
Pop up an error "Can't find Storage Explorer".
![error](https://user-images.githubusercontent.com/34729022/39859377-3174d45a-5429-11e8-8cd7-6513e6406d15.png)

More info:
No response after clicking 'Download' in the dialog.

PrashanthCorp:
Related to an existing issue on Azure Storage Explorer - the deep link feature does not work on Linux.

PrashanthCorp:
Mac: 
I could not repro the issue on Mac. If I click Download, I'm directed to the Storage Explorer landing page. Clicking browse allows me to pick the location of the file. 

Linux: 
Browse opens up a dialog, but the StorageExplorer executable is not visible due to the filter on the fileType, here: 
https://github.com/Microsoft/vscode-azurestorage/blob/857d35df8fab1d66369dd9d12ffa0203bbcd4310/src/storageExplorerLauncher/macOSStorageExplorerLauncher.ts#L91
The workaround (manually setting a path in the settings) tries to open the wrong path. Fixing that, the deep link issue in Linux further prevents Storage Explorer from opening. 
We can failry say that this feature is not yet supported on Linux.

PR #900, opened 2020-12-07
Don't show 'Open in Storage Explorer' menu item on Linux or Codespaces

wwlorey:
Fixes https://github.com/microsoft/vscode-azurestorage/issues/140
Fixes https://github.com/microsoft/vscode-azurestorage/issues/887

### DIFF
--- a/package.json
+++ b/package.json
@@ -403,7 +403,7 @@
                 {
                     "$comment": "========= azureStorageAccount =========",
                     "command": "azureStorage.openStorageAccount",
-                    "when": "view == azureStorage && viewItem == azureStorageAccount && !azureStorage.isLinux",
+                    "when": "view == azureStorage && viewItem == azureStorageAccount && !isWeb && !isLinux",
                     "group": "navigation@1"
                 },
                 {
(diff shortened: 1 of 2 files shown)


nturinski (approved):
Looks good to me-- setting contexts isn't something that we typically do so I'm not sure what design pattern that we want to go with.

However, this is what's in the vscode docs: https://github.com/microsoft/vscode-docs/commit/4ae8f109dec6d7ed58439abdcb44b31a6b9ea8e6#diff-6b81afb0e3d221d6a0240a5399bd500fd604df9a0c7f6cb50b2a5dd068081339R213-R221

The only thing I might suggest is to move this into the activate command below just to make sure everything is set up (probably doesn't matter with platform, but we also don't need to run this unless we activate anyway)

ejizba:
In regards to Nathan's comments, "setContext" is okay to use (we _do_ use it in a few other places), but in this particular case there's a simple "isLinux" context already. Also seems to be an "isWeb" property, which might be related to Codespaces and might be helpful? idk

https://code.visualstudio.com/docs/getstarted/keybindings#_available-contexts

wwlorey:
`isLinux` is false on Codespaces so I added a check for `isWeb` too

ejizba (approved):

Issue #632, opened 2020-03-09
Include size/mtime/ctime in AzureStorageFS.stat


ejizba:
Right now we're setting these to `0`, but [vscode.d.ts](https://github.com/microsoft/vscode/blob/master/src/vs/vscode.d.ts#L5960) has this to say: 

> *Note:* If the file changed, it is important to provide an updated `size`. Otherwise there may be optimizations in place that will not show the updated file contents in an editor for example.

PR #903, opened 2020-12-14
Populate ctime, mtime, and size in AzureStorageFS.stat

wwlorey:
Fixes https://github.com/microsoft/vscode-azurestorage/issues/632

### DIFF
--- a/src/AzureStorageFS.ts
+++ b/src/AzureStorageFS.ts
@@ -100,6 +100,10 @@ export class AzureStorageFS implements vscode.FileSystemProvider, vscode.TextDoc
     }
 
     async stat(uri: vscode.Uri): Promise<vscode.FileStat> {
+        let ctime: number = 0;
+        let mtime: number = 0;
+        let size: number = 0;
+
         return await callWithTelemetryAndErrorHandling('stat', async (context) => {
             context.telemetry.suppressIfSuccessful = true;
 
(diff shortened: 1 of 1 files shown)


src/AzureStorageFS.ts
```diff
@@ -111,12 +111,28 @@ export class AzureStorageFS implements vscode.FileSystemProvider, vscode.TextDoc
 
             let treeItem: AzureStorageTreeItem = await this.lookup(uri, context);
             let fileType: vscode.FileType = treeItem instanceof DirectoryTreeItem || treeItem instanceof FileShareTreeItem || treeItem instanceof BlobDirectoryTreeItem || treeItem instanceof BlobContainerTreeItem ? vscode.FileType.Directory : vscode.FileType.File;
+            let ctime: number | undefined;
```
nturinski, line 114: I took a shot at refactoring it.  This feels easier to read for me while removing some duplicated code.  This also should guarantee that a props object will exist though because of the if/else if conditional, typescript won't honor that.  Though I'm pretty sure only `BlobTreeItem` and `FileTreeItem` are of the `FileType.File`.

```
        let ctime: number = 0;
        let mtime: number = 0;
        let size: number = 0;

        return await callWithTelemetryAndErrorHandling('stat', async (context) => {

            if (uri.path.endsWith('/')) {
                // Ignore trailing forward slashes
                // https://github.com/microsoft/vscode-azurestorage/issues/576
                uri = uri.with({ path: uri.path.slice(0, -1) });
            }

            let treeItem: AzureStorageTreeItem = await this.lookup(uri, context);
            let fileType: vscode.FileType = treeItem instanceof DirectoryTreeItem || treeItem instanceof FileShareTreeItem || treeItem instanceof BlobDirectoryTreeItem || treeItem instanceof BlobContainerTreeItem ? vscode.FileType.Directory : vscode.FileType.File;

            if (fileType === vscode.FileType.File) {
                let props: (BlobGetPropertiesResponse & FileGetPropertiesResponse) | undefined;
                if (treeItem instanceof BlobTreeItem) {
                    const blockBlobClient: BlockBlobClient = createBlockBlobClient(treeItem.root, treeItem.container.name, treeItem.blobPath);
                    props = await blockBlobClient.getProperties();
                } else if (treeItem instanceof FileTreeItem) {
                    const fileClient: ShareFileClient = createFileClient(treeItem.root, treeItem.shareName, treeItem.directoryPath, treeItem.fileName);
                    props = await fileClient.getProperties();
                }

                ctime = props?.createdOn?.valueOf() || props?.fileCreatedOn?.valueOf() || 0;
                mtime = props?.lastModified?.valueOf() || 0;
                size = props?.contentLength || 0;
            }

            return { type: fileType, ctime, mtime, size };
        }) || { type: vscode.FileType.Unknown, ctime, mtime, size };
    }
```

wwlorey: I think this change looks good 👌 The only thing I got rid of is the `fileType === vscode.FileType.File` check because the `treeItem instanceof` checks should suffice on their own

nturinski (approved):

PR #905, opened 2020-12-19
Support Azure stack with getting environment for azure stack in vscode extension ui

v-tsun:


src/tree/SubscriptionTreeItem.ts
```diff
@@ -23,37 +23,42 @@ export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
...
+            );
+        } catch (error) {
+            throw error;
+        }
     }
 
     public async createChildImpl(context: ICreateChildImplContext): Promise<AzureTreeItem> {
-        const defaultLocation = 'westus';
+        let isStack = this.root.environment.name === "AzurePPE" ? true : false;
+        const defaultLocation = isStack ? 'local' : 'westus';
```
rakku-ms, line 42: default location is not always 'local' for AzureStack. You need to get this info from ARM. @bganapa FYI

src/tree/SubscriptionTreeItem.ts
```diff
@@ -23,37 +23,49 @@ export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
...
+                sa => sa.name
+            );
+        } catch (error) {
+            throw error;
+        }
     }
 
     public async createChildImpl(context: ICreateChildImplContext): Promise<AzureTreeItem> {
-        const defaultLocation = 'westus';
+        let isStack = this.root.environment.name === "AzurePPE" ? true : false;
```
wwlorey, line 41: I think it would be useful to make `isStack` an extension variable (https://github.com/microsoft/vscode-azurestorage/blob/main/src/extensionVariables.ts). I could see it being needed elsewhere in the extension. And you wouldn't have to pass it into `StaticWebsiteEnableStep` and could remove the constructor you added

src/tree/SubscriptionTreeItem.ts
```diff
@@ -23,37 +23,49 @@ export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
...
         const wizardContext: IStorageAccountWizardContext = Object.assign(context, this.root);
+        let stackLocation = (await LocationListStep.getLocations(wizardContext)).find(l => l.displayName !== undefined || l.name !== undefined);
+        let defaultStackLocation: string;
+        if (stackLocation === undefined) {
+            throw new Error("there is no available location for resource provider in azure stack");
+        } else {
+            defaultStackLocation = <string>(stackLocation.name !== undefined ? stackLocation.name : stackLocation.displayName);
+        }
+        const defaultLocation: string = isStack ? defaultStackLocation : 'westus';
```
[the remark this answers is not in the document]
v-tsun: Hey will, the default location for azure stack would be different from each other. Therefore, we cannot have a hard code location suitable for all environments. If you have any other idea to make this piece of code better, just let me know.

src/tree/SubscriptionTreeItem.ts
```diff
@@ -23,37 +23,49 @@ export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
...
         const promptSteps: AzureWizardPromptStep<IStorageAccountWizardContext>[] = [new StorageAccountNameStep()];
         const executeSteps: AzureWizardExecuteStep<IStorageAccountWizardContext>[] = [
-            new StorageAccountCreateStep({ kind: StorageAccountKind.StorageV2, performance: StorageAccountPerformance.Standard, replication: StorageAccountReplication.LRS }),
+            new StorageAccountCreateStep({ kind: isStack ? StorageAccountKind.Storage : StorageAccountKind.StorageV2, performance: StorageAccountPerformance.Standard, replication: StorageAccountReplication.LRS }),
```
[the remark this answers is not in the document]
v-tsun: hey Will, hide static website in context menu. Also hide them in commands, if you think we should leave them in commands, just let me know, I will undo them.

bganapa:
@v-tsun lets change this to PR , accounts PR is merged now https://github.com/microsoft/vscode-azure-account/pull/241

src/tree/SubscriptionTreeItem.ts
```diff
@@ -34,11 +36,21 @@ export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
...
 
     public async createChildImpl(context: ICreateChildImplContext): Promise<AzureTreeItem> {
-        const defaultLocation = 'westus';
+        if (this.root.environment.name === "AzurePPE") {
+            ext.isStack = true;
+        }
         const wizardContext: IStorageAccountWizardContext = Object.assign(context, this.root);
+        let stackLocation: SubscriptionModels.Location | undefined = (await LocationListStep.getLocations(wizardContext)).find(l => l.displayName !== undefined || l.name !== undefined);
+        let defaultStackLocation: string;
+        if (stackLocation === undefined) {
```
wwlorey, line 45: This check should include `isStack === true` since it's an Azure Stack specific error

src/utils/azureClients.ts
```diff
@@ -10,5 +10,9 @@ import { createAzureClient, ISubscriptionContext } from 'vscode-azureextensionui
 // NOTE: The client is the only import that matters, the rest of the types disappear when compiled to JavaScript
 
 export async function createStorageClient<T extends ISubscriptionContext>(context: T): Promise<StorageManagementClient> {
-    return createAzureClient(context, (await import('@azure/arm-storage')).StorageManagementClient);
+    if (context.environment.name === "AzurePPE") {
+        return <StorageManagementClient><unknown>createAzureClient(context, (await import('@azure/arm-storage-profile-2019-03-01-hybrid')).StorageManagementClient);
```
wwlorey, line 14: Will this cast to unknown be fixed by either the Azure account PR or the Azure tools PR?

wwlorey (commented):
> hey Will, hide static website in context menu. Also hide them in commands, if you think we should leave them in commands, just let me know, I will undo them.

It does worries me to touch lots of the commands in package.json. That seems like it could introduce more problems that it's worth. 

Let's revert the package.json changes and just rely on V1 account errors that are already in place (like the one below.) That should be fine especially because people using V1 accounts will likely still be in the minority.

Screen Shot 2021-01-27 at 1 10 14 PM

src/utils/azureClients.ts
v-tsun: Nope, we still need this one since we are still using different api version in azure stack. Therefore, we should import different packages in order to use the one with correct api version.

package-lock.json
```diff
@@ -401,6 +468,29 @@
                 "p-retry": "*"
             }
         },
+        "@types/request": {
```
wwlorey, line 471: Can you run `npm install` to get rid of this?

src/tree/createWizard/StaticWebsiteEnableStep.ts
```diff
@@ -12,12 +12,24 @@ import { StaticWebsiteErrorDocument404Step } from "./StaticWebsiteErrorDocument4
 import { StaticWebsiteIndexDocumentStep } from "./StaticWebsiteIndexDocumentStep";
 
 export class StaticWebsiteEnableStep extends AzureWizardPromptStep<IStaticWebsiteConfigWizardContext> {
+    isStack: boolean;
+    /**
+     *
+     */
```
wwlorey, line 18: You can remove this comment

src/tree/createWizard/StaticWebsiteEnableStep.ts
```diff
@@ -12,12 +12,24 @@ import { StaticWebsiteErrorDocument404Step } from "./StaticWebsiteErrorDocument4
 import { StaticWebsiteIndexDocumentStep } from "./StaticWebsiteIndexDocumentStep";
 
 export class StaticWebsiteEnableStep extends AzureWizardPromptStep<IStaticWebsiteConfigWizardContext> {
+    isStack: boolean;
```
wwlorey, line 15: nit: 

```suggestion
    private isStack: boolean;
```

src/tree/SubscriptionTreeItem.ts
```diff
@@ -88,4 +92,19 @@ export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
     public isAncestorOfImpl(contextValue: string): boolean {
         return contextValue !== AttachedStorageAccountTreeItem.baseContextValue && contextValue !== AttachedStorageAccountTreeItem.emulatedContextValue;
     }
+
+    private async getDefaultLocation(wizardContext: IStorageAccountWizardContext): Promise<string> {
+        const stackLocation: SubscriptionModels.Location | undefined = (await LocationListStep.getLocations(wizardContext)).find(l => l.displayName !== undefined || l.name !== undefined);
+        let defaultStackLocation: string;
```
wwlorey, line 98: nit: Since this isn't always the stack location
```suggestion
        let defaultLocation: string;
```

src/tree/SubscriptionTreeItem.ts
```diff
@@ -88,4 +92,19 @@ export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
...
     }
+
+    private async getDefaultLocation(wizardContext: IStorageAccountWizardContext): Promise<string> {
+        const stackLocation: SubscriptionModels.Location | undefined = (await LocationListStep.getLocations(wizardContext)).find(l => l.displayName !== undefined || l.name !== undefined);
+        let defaultLocation: string;
+        if (this.isStack) {
+            if (stackLocation === undefined) {
+                throw new Error(localize("noAvilableLocation", "There is no available location for resource provider in Azure Stack"));
+            } else {
+                defaultLocation = <string>(stackLocation.name !== undefined ? stackLocation.name : stackLocation.displayName);
```
nturinski, line 103: Location can actually be `undefined` and our wizard will handle it.  I don't think you need to cast the value like this and I'd prefer if you made the `getDefaultLocation` return type be `Promise<string | undefined>`.

Then you could change this line to 
```suggestion
                defaultLocation = stackLocation.name || stackLocation.displayName;
```

nturinski:
Oh, is there a list of locations specifically for the AzurePPE stack?

src/tree/SubscriptionTreeItem.ts
```diff
@@ -21,6 +23,7 @@ import { StorageAccountTreeItem } from './StorageAccountTreeItem';
 export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
     public childTypeLabel: string = "Storage Account";
     public supportsAdvancedCreation: boolean = true;
+    private isStack: boolean;
```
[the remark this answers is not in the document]
v-tsun: I still think we need to keep this value, since we need to check it out, and in storage account create step we need to give different kind according to this value

v-tsun:
> Oh, is there a list of locations specifically for the AzurePPE stack?

I believe we cannot get a list of locations since location for Azure stack would be different from each other under different envrionments

src/tree/SubscriptionTreeItem.ts
v-tsun: hey I am trying what you said but I found that in line 62 in SubscriptionTreeItem.ts. LocationListStep.setLocation doesn't accept undefined value

bganapa:
> Oh, is there a list of locations specifically for the AzurePPE stack?

The location names in AzureStack are customer defined, it is different for every azurestack deployment.

src/tree/SubscriptionTreeItem.ts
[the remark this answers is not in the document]
v-tsun: @nturinski Hey Nathan, what do think if we add this isStack to ISubscriptionContext instead of IStorageAccountWizardContext as we disussed before, since we also will add azure stack support to resource group and so on, so I think add it in ISubscriptionContext would let it can be used more generally for other extensions.

v-tsun:
The IStorageWizardContext update PR in ui is https://github.com/microsoft/vscode-azuretools/pull/8

v-tsun:
Hey @ejizba, also on this PR

src/tree/SubscriptionTreeItem.ts
```diff
@@ -34,11 +36,11 @@ export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
     }
 
     public async createChildImpl(context: ICreateChildImplContext): Promise<AzureTreeItem> {
-        const defaultLocation = 'westus';
         const wizardContext: IStorageAccountWizardContext = Object.assign(context, this.root);
+        const defaultLocation: string | undefined = await this.getDefaultLocation(wizardContext);
```
ejizba, line 40: The only reason we have a default location is to reduce the number of prompts. It's a minor enhancement, not something critical. I think I'd rather remove the semi-complex code for custom clouds (`this.getDefaultLocation`) and just do something like this:
```
const defaultLocation: string | undefined = wizardContext.isCustomCloud ? undefined : 'westus';
```

package.json
```diff
@@ -873,6 +873,7 @@
     "dependencies": {
         "@azure-tools/azcopy-node": "1.0.0",
         "@azure/arm-storage": "^15.0.0",
+        "@azure/arm-storage-profile-2019-03-01-hybrid": "^1.0.0",
         "@azure/storage-blob": "^12.1.1",
```
ejizba, line 877: v0.39.5 of the ui package is published and ready for you to use

src/tree/SubscriptionTreeItem.ts
bganapa: @ejizba single Azure stack cloud supports only one location. I think it makes sense to get the location ourselves and set it here

src/utils/azureClients.ts
bganapa: I think this comment is resolved. pls close it @ejizba / @wwlorey

bganapa (approved):

src/tree/SubscriptionTreeItem.ts
v-tsun: Hey @ejizba, I don't think we can set the defaultLocation to undefined. First of all, https://github.com/microsoft/vscode-azurestorage/blob/1d73643104cd6e5e8ee29ee3c9db202f7c0020b1/src/tree/SubscriptionTreeItem.ts#L60-L62 If we set defaultLocation as undefined here, we can't get into the setLocation step. The location of wizard would be undefined when execute. But I believe in both Resourcegroup and Storageaccount creation steps we need to have name of locations, like https://github.com/microsoft/vscode-azuretools/blob/60dd5faeb66104cb3c33f34dfdbd5df1f74f2c4a/ui/src/wizard/ResourceGroupCreateStep.ts#L23 and https://github.com/microsoft/vscode-azurestorage/blob/1d73643104cd6e5e8ee29ee3c9db202f7c0020b1/src/tree/createWizard/storageAccountCreateStep.ts#L24 If the location is undefined, it would throw error can't get name from undefined. 
And as @bganapa said, in Azure Stack, we have only one location per environment, but they might be various from each other. Therefore, we can't have a single default location for us. That is why we need to have this.getDefaultLocation to get location for each environment. Thx

ejizba: In that case, I think you need to modify `LocationListStep`. It sounds like it's just broken for custom clouds and we should fix it there, probably inside [`getLocations`](https://github.com/microsoft/vscode-azuretools/blob/main/ui/src/wizard/LocationListStep.ts#L50). As I said, the `defaultLocation` here is just meant as a minor enhancement to reduce the number of prompts. You're talking about actual functionality and we want to fix that for all extensions, not just here.

v-tsun: @ejizba I made a new PR in azuretools extension to fix the issue. https://github.com/microsoft/vscode-azuretools/pull/859. Maybe you can take a look.

src/tree/SubscriptionTreeItem.ts
```diff
@@ -51,11 +53,13 @@ export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
...
-                indexDocument: StaticWebsiteIndexDocumentStep.defaultIndexDocument,
-                errorDocument404Path: StaticWebsiteErrorDocument404Step.defaultErrorDocument404Path
+                enableStaticWebsite: wizardContext.isCustomCloud ? false : true,
+                indexDocument: wizardContext.isCustomCloud ? "" : StaticWebsiteIndexDocumentStep.defaultIndexDocument,
+                errorDocument404Path: wizardContext.isCustomCloud ? "" : StaticWebsiteErrorDocument404Step.defaultErrorDocument404Path
             });
-            await LocationListStep.setLocation(wizardContext, defaultLocation);
+            if (defaultLocation) {
+                await LocationListStep.setLocation(wizardContext, defaultLocation);
+            }
```
ejizba, line 62:
```suggestion
            if (defaultLocation) {
                await LocationListStep.setLocation(wizardContext, defaultLocation);
            } else {
                LocationListStep.addStep(wizardContext, promptSteps);
            }
```

src/tree/SubscriptionTreeItem.ts
ejizba: > I don't think we can set the defaultLocation to undefined.

Based on your comment, I thought `getLocations` was broken for custom clouds, but you only changed `setLocation`. This is much simpler than I thought - all you have to do is make sure the `LocationListStep` is in the wizard prompt steps. See suggested change in other comment.

src/tree/SubscriptionTreeItem.ts
v-tsun: I see, I will close the PR for setLocation in Azuretools and just add location pickup for Azure Custom Cloud.

package-lock.json
```diff
@@ -1,13513 +1,8 @@
 {
     "name": "vscode-azurestorage",
     "version": "0.12.1-alpha",
-    "lockfileVersion": 2,
+    "lockfileVersion": 1,
```
ejizba, line 4: You gotta update to the latest npm (at least v7 I think) and run `npm install` again so that there's not a ton of changes to this file

ejizba (approved):

Issue #848, opened 2020-10-12
[Suggestion] Add a validation when entering an existed name for the new directory


v-mengwe:
**OS:** Win10
**Build Version:** 20201011.1

**Repro Steps:**

1. Create a file share -> Right click the file share -> Create an empty file with name `test`.
2. Right click the file share -> Select "Create Directory..." -> Enter a name with `test` for the new directory.
3. Check the result.

**Actual:**
There is an error "The specified resource type does not match the type of the existing resource" after pressing enter.
![image](https://user-images.githubusercontent.com/62053665/95737285-04dd6280-0cba-11eb-9da2-10fde8e8afc3.png)

**Suggestion:**
Add a validation when entering a name for the new directory.

**More Info:**
Upload a folder with name "test" to a blob container -> Create an empty blob with name "test" -> Display an error.
![image](https://user-images.githubusercontent.com/62053665/95736462-db700700-0cb8-11eb-91c0-d7ad13c75e0c.png)

PR #907, opened 2020-12-21
Check new file/directory names against existing resources

wwlorey:
This also combines name validation logic for files and directories because it's so similar

Fixes https://github.com/microsoft/vscode-azurestorage/issues/848

### DIFF
--- a/src/utils/fileUtils.ts
+++ b/src/utils/fileUtils.ts
@@ -5,13 +5,16 @@
 
 import * as azureStorageShare from '@azure/storage-file-share';
 import * as mime from 'mime';
+import { posix } from 'path';
 import { ProgressLocation, window } from "vscode";
-import { AzureParentTreeItem, ICreateChildImplContext, UserCancelledError } from "vscode-azureextensionui";
+import { AzureParentTreeItem, ICreateChildImplContext } from "vscode-azureextensionui";
 import { ext } from "../extensionVariables";
 import { IFileShareCreateChildContext } from "../tree/fileShare/FileShareTreeItem";
 import { FileTreeItem } from "../tree/fileShare/FileTreeItem";
 import { IStorageRoot } from "../tree/IStorageRoot";
-import { validateFileName } from "./validateNames";
+import { doesDirectoryExist } from './directoryUtils';
+import { localize } from './localize';
+import { validateFileOrDirectoryName } from './validateNames';
 
 export function createShareClient(root: IStorageRoot, shareName: string): azureStorageShare.ShareClient {
     const shareServiceClient: azureStorageShare.ShareServiceClient = root.createShareServiceClient();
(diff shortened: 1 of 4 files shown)


src/utils/validateNames.ts
```diff
@@ -22,37 +22,19 @@ export function validateBlobDirectoryName(name: string): string | undefined {
...
-        return "File name cannot be empty";
-    }
-
-    if (name.length < validLength.min || name.length > validLength.max) {
-        return `File name must contain between ${validLength.min} and ${validLength.max} characters`;
-    }
-
-    if (invalidFileChars.some(ch => name.indexOf(ch) >= 0)) {
-        return `File name cannot contain the following characters: ${invalidFileCharsString}'`;
+        return localize('nameCantContain', 'Name cannot contain the following characters: {0}', invalidFileCharsString);
```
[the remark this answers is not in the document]
wwlorey: Are you talking about this check? 
```ts
 if (invalidFileChars.some(ch => name.indexOf(ch) >= 0)) {
```

It was used for files and directories. `validateFileName` and `validateFileDirectoryName` were the same except for  the return strings (one used "File" verbiage and the other used "Directory" so I just combined them)

src/utils/fileUtils.ts
```diff
@@ -29,29 +32,25 @@ export function createFileClient(root: IStorageRoot, shareName: string, director
...
+    return await window.withProgress({ location: ProgressLocation.Window }, async (progress) => {
+        context.showCreatingTreeItem(fileName);
+        progress.report({ message: `Azure Storage: Creating file '${fileName}'` });
+        await createFile(directoryPath, fileName, shareName, parent.root);
+        return new FileTreeItem(parent, fileName, directoryPath, shareName);
+    });
 }
 
-export async function getFileName(parent: AzureParentTreeItem<IStorageRoot>, directoryPath: string, shareName: string, value?: string): Promise<string> {
+export async function getFileOrDirectoryName(parent: AzureParentTreeItem<IStorageRoot>, directoryPath: string, shareName: string, value?: string): Promise<string> {
```
nturinski, line 44: I thought you made it so files and directories could co-exist.  Wouldn't this prevent me from having a file and folder both named 'abc'?

nturinski: Also kind of weird for it to be in the `fileUtils.ts` now.

nturinski (commented):
I think I'm okay with combining the validation logic, but I would advise against combining the getName logic.

src/utils/fileUtils.ts
wwlorey: > I thought you made it so files and directories could co-exist. Wouldn't this prevent me from having a file and folder both named 'abc'?

That fix was only for blobs and blob directories. File share files and directories with the same name aren't allowed to exist in Azure. The naming gets confusing because you upload either file share files or blobs using the "Upload Files..." command.

> Also kind of weird for it to be in the fileUtils.ts now.

I agree that putting it in `fileUtils.ts` isn't the best fit but I was hesitant to create a new file just for this. Do you have a recommendation for a better home?

wwlorey (commented):
> I think I'm okay with combining the validation logic, but I would advise against combining the getName logic.

The only difference between the getName functions if I separated them out would be the strings (i.e. "file already exists" vs "directory already exists"). Is that the main reason you don't want to combine the getName logic?

nturinski:
It was actually more because of this
> > I thought you made it so files and directories could co-exist. Wouldn't this prevent me from having a file and folder both named 'abc'?
>
> That fix was only for blobs and blob directories. File share files and directories with the same name aren't allowed to exist in Azure. The naming gets confusing because you upload either file share files or blobs using the "Upload Files..." command.

I guess we could call it fileShareUtils or something since it actually encompasses more than just files anyway.  That being said, I feel like a lot of those could be taken out and put into their own commands in the fileShare folder or something.  Doesn't have to be now, but that utils folder is ridiculously overloaded with random code that probably could find a home.

nturinski (approved):

src/utils/validateNames.ts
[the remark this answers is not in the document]
neelip: In that case, you might want to rename `invalidFileCharsString` to include directories as well.

Issue #896, opened 2020-12-04
[Codespace] Downloading append blob or page blob triggers to launch or install Storage Explorer - can't in Codespace


jinglouMSFT:
There are places in our code where we ask users to use or install Storage Explorer. We need to disable or remove the UI when the extension is run on Codespace.  

Repro Steps:
1.  Preparation: In Storage Explorer upload an append blob. You can create these by uploading in Storage Explorer and choose to upload as an append blog.
2. Once the blob is uploaded, go to the Codespace Storage extension treeview, refresh the UI so you can see the blob that was just uploaded
3. Right click on the blob to bring up the context menu for download
4. In the command palette select the location where you want to download the blob, then select OK
5. You will get an info window pop up at the lower right corner:
![image](https://user-images.githubusercontent.com/10081483/101111535-3d017300-3590-11eb-9857-df0c808fb9c6.png)
6. Click on the "Open Resources in Storage Explorer button"
7. You will be prompted again:
![image](https://user-images.githubusercontent.com/10081483/101111635-70440200-3590-11eb-9575-2bbfb07fb204.png)
8. Click on the "Download" button, nothing will happen

Since VS Code manages to install the extension from within the VS Code UI, I am not sure if they provided any API's for us to install Storage Explorer from clicking the "Download" button in step 8.  If no, we probably should stop at Step 5, but still need a way to tell the user to use Storage Explorer somewhere else.  

We could also consider supporting page blob and append blob at some point.

wwlorey:
Tracking page and append blob support in https://github.com/microsoft/vscode-azurestorage/issues/897 and https://github.com/microsoft/vscode-azurestorage/issues/867

PR #908, opened 2020-12-23
Don't show 'Open resource in Storage Explorer' error action on Linux (Codespaces)

wwlorey:
If Storage Explorer is required in Linux, an error is still thrown but without an action button to open in SE.

| Mac/Windows | Linux/Codespaces |
| - | - |
| Screen Shot 2021-01-07 at 1 12 00 PM| Screen Shot 2021-01-07 at 1 09 38 PM |

Fixes https://github.com/microsoft/vscode-azurestorage/issues/896

### DIFF
--- a/src/storageExplorerLauncher/windowsStorageExplorerLauncher.ts
+++ b/src/storageExplorerLauncher/windowsStorageExplorerLauncher.ts
@@ -6,13 +6,13 @@ import * as fs from "fs";
 import { MessageItem } from "vscode";
 import { callWithTelemetryAndErrorHandling, UserCancelledError } from "vscode-azureextensionui";
 import * as winreg from "winreg";
+import { storageExplorerDownloadUrl } from "../constants";
 import { ext } from "../extensionVariables";
 import { Launcher } from "../utils/launcher";
 import { localize } from "../utils/localize";
 import { IStorageExplorerLauncher } from "./IStorageExplorerLauncher";
 import { ResourceType } from "./ResourceType";
 
-const downloadPageUrl: string = "https://go.microsoft.com/fwlink/?LinkId=723579";
 const regKey: { hive: string, key: string } = { hive: "HKCR", key: "\\storageexplorer\\shell\\open\\command" };
 
 export class WindowsStorageExplorerLauncher implements IStorageExplorerLauncher {
@@ -92,7 +92,7 @@ export class WindowsStorageExplorerLauncher implements IStorageExplorerLauncher 
 
     private static async downloadStorageExplorer(): Promise<void> {
         //I'm not sure why running start directly doesn't work. Opening separate cmd to run the command works well
-        await Launcher.launch("cmd", "/c", "start", downloadPageUrl);
+        await Launcher.launch("cmd", "/c", "start", storageExplorerDownloadUrl);
     }
 
     private static async launchStorageExplorer(args: string[] = []): Promise<void> {
--- a/src/storageExplorerLauncher/windowsStorageExplorerLauncher.ts
+++ b/src/storageExplorerLauncher/windowsStorageExplorerLauncher.ts
@@ -10,6 +10,7 @@ import { storageExplorerDownloadUrl } from "../constants";
 import { ext } from "../extensionVariables";
 import { Launcher } from "../utils/launcher";
 import { localize } from "../utils/localize";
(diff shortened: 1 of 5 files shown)


src/utils/askOpenInStorageExplorer.ts
```diff
@@ -11,7 +12,8 @@ import { localize } from "./localize";
 
 export async function askOpenInStorageExplorer(context: IActionContext, errorMessage: string, resourceId: string, subscriptionId: string, resourceType: ResourceType, resourceName: string): Promise<void> {
     const message: string = localize("openInSE", "Open resource in Storage Explorer");
-    window.showErrorMessage(errorMessage, message).then(async result => {
+    const items: string[] = platform() === 'linux' ? [] : [message];
```
nturinski, line 15: Does the platform show up as Linux on Codespaces?

neelip:
Should we add a link directing the user to the Storage Explorer download page? https://azure.microsoft.com/en-us/features/storage-explorer/

wwlorey:
> Is it better to add a link directing the user to the Storage Explorer download page instead of completely removing a link? https://azure.microsoft.com/en-us/features/storage-explorer/

I dig it 👍

src/utils/askOpenInStorageExplorer.ts
wwlorey: Yeah the platform shows up as Linux. Interestingly enough, the context value "isLinux" if false on Codespaces while "isWeb" is true.

src/storageExplorerLauncher/windowsStorageExplorerLauncher.ts
```diff
@@ -92,7 +92,7 @@ export class WindowsStorageExplorerLauncher implements IStorageExplorerLauncher
 
     private static async downloadStorageExplorer(): Promise<void> {
         //I'm not sure why running start directly doesn't work. Opening separate cmd to run the command works well
-        await Launcher.launch("cmd", "/c", "start", downloadPageUrl);
+        await Launcher.launch("cmd", "/c", "start", storageExplorerDownloadUrl);
```
ejizba, line 95: oof we should get rid of this `Launcher.launch("cmd"` stuff and just use `openUrl` at some point

src/utils/askOpenInStorageExplorer.ts
```diff
@@ -3,18 +3,26 @@
...
+    const openMessage: string = localize('openInSE', 'Open resource in Storage Explorer');
+    const downloadMessage: string = localize('downloadSE', 'Download Storage Explorer');
+    const message: string = platform() === 'linux' ? downloadMessage : openMessage;
     window.showErrorMessage(errorMessage, message).then(async result => {
-        if (result === message) {
+        if (result === openMessage) {
             context.telemetry.properties.openInStorageExplorer = 'true';
             await storageExplorerLauncher.openResource(resourceId, subscriptionId, resourceType, resourceName);
+        } else if (result === downloadMessage) {
+            context.telemetry.properties.openStorageExplorerDownloadUrl = 'true';
```
ejizba, line 24: The user will never be able to install Storage Explorer in Codespaces, so we shouldn't point them to it. Best we could probably do is show some generic "This is not supported" error

wwlorey: They wouldn't be able to download Storage Explorer in Codespaces but they could always download it on their local machine. Do you think it would be confusing to suggest they download it via a notification in Codespaces?

ejizba: Well I thought Jing wanted it removed:

> There are places in our code where we ask users to use or install Storage Explorer. We need to disable or remove the UI when the extension is run on Codespace.

I guess I could go either way now. Maybe Jing should decide since it was his issue. It does bother me that it says "Download" every time even though I might already have it installed on my local machine. Another option to consider is a more subtle approach where you link to Storage explorer in the message itself - something like `Please use [Storage Explorer](http://aka.ms/dfsdfds) ...`

wwlorey: It had been so long since I looked at the original issue that I forgot Jing did explicitly say that 😅

With that said I like the "`Please use [Storage Explorer]()...`" approach. It seems better than a "not supported" error. And I agree it could be confusing to prompt to download it if someone already has it installed on their machine.

ejizba (approved):

PR #918, opened 2021-01-15
Retry writeFile for 'Server failed to authenticate' error

wwlorey:
Fixes https://github.com/microsoft/vscode-azurestorage/issues/869

The diff looks really scary but I recommend hiding whitespace changes.

### DIFF
--- a/src/AzureStorageFS.ts
+++ b/src/AzureStorageFS.ts
@@ -5,10 +5,11 @@
 
 import { BlobClient, BlobDownloadResponseModel, BlobGetPropertiesResponse, BlockBlobClient } from "@azure/storage-blob";
 import { FileDownloadResponseModel, FileGetPropertiesResponse, ShareFileClient } from "@azure/storage-file-share";
+import * as retry from 'p-retry';
 import * as path from "path";
 import * as querystring from "querystring";
 import * as vscode from "vscode";
-import { AzExtTreeItem, callWithTelemetryAndErrorHandling, IActionContext, parseError, UserCancelledError } from "vscode-azureextensionui";
+import { AzExtTreeItem, callWithTelemetryAndErrorHandling, IActionContext, IParsedError, parseError, UserCancelledError } from "vscode-azureextensionui";
 import { download } from "./commands/downloadFile";
 import { maxRemoteFileEditSizeBytes, maxRemoteFileEditSizeMB } from "./constants";
 import { ext } from "./extensionVariables";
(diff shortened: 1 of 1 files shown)


src/AzureStorageFS.ts
```diff
@@ -245,72 +246,95 @@ export class AzureStorageFS implements vscode.FileSystemProvider, vscode.TextDoc
...
+                            }
+                        });
+                    }
+                },
+                {
+                    retries,
+                    minTimeout: 2 * 1000,
+                    onFailedAttempt: error => {
+                        const parsedError: IParsedError = parseError(error);
+                        if (/server failed to authenticate/i.test(parsedError.message)) {
```
ejizba, line 330: What makes us think this will fix the issue? The person said "cannot open anymore in VSCode" implying to me that it doesn't change when they manually retry

wwlorey: The thought behind it is that the `content` passed into `writeFile` won't change and will remain valid even if the upload fails and corrupts the file in Azure. So retrying would keep attempting to upload the local content which hasn't been corrupted. If `writeFile` failed to upload and corrupted that file in Azure, I don't think the file would show as corrupted until you loaded it from Azure.

ejizba: Have you been able to repro? It seems weird to me that it would "fail to authenticate" and yet somehow at the same time corrupt the file in Azure. I would think the error would prevent it from doing anything in Azure.

Also any more info on this? https://github.com/microsoft/vscode-azurestorage/pull/854#discussion_r525563663 Is retrying that error still not helpful?

ejizba: Also not sure if you've googled this error, but seems like there are a lot of potential avenues to explore to see if you can repro/fix rather than just a retry, for example https://stackoverflow.com/questions/24492790/azurestorage-blob-server-failed-to-authenticate-the-request-make-sure-the-value

wwlorey: > Have you been able to repro? It seems weird to me that it would "fail to authenticate" and yet somehow at the same time corrupt the file in Azure. I would think the error would prevent it from doing anything in Azure.

I haven't been able to repro it. I agree that it seems weird that an authentication failure could be the main culprit for corrupting files. The auth failure seemed to be what happened right before the file was corrupted but I suppose it could have been any number of things.

> Also any more info on this? [#854 (comment)](https://github.com/microsoft/vscode-azurestorage/pull/854#discussion_r525563663) Is retrying that error still not helpful?

There are still very low rates of retrying deploys and it doesn't seem to be helpful (`deployattempt` got up to 5 in a few cases):

|EventName                                           |deployattempt|Machines|EventCount|AveEventsPerMachine|
|----------------------------------------------------|-------------|--------|----------|-------------------|
|vscode-azurestorage/azurestorage.deploystaticwebsite|5            |2       |3         |1                  |
|vscode-azurestorage/azurestorage.deploystaticwebsite|2            |2       |2         |1                  |
|vscode-azurestorage/azurestorage.deploystaticwebsite|1            |1527    |11687     |7                  |
|vscode-azurestorage/azurestorage.deploystaticwebsite|null         |890     |2283      |2                  |

After looking at the link you posted and other threads, it looks like there's a ton of things that could be causing the error. I'll do some more digging and see if there's a better catch-all fix than retrying.

wwlorey:
As I mentioned, there are a lot of potential causes for the "Server failed to authenticate" error and retrying doesn't help

Issue #873, opened 2020-10-30
Upload folder to blobs is exceeding 100%


nturinski:
When uploading a folder to a blob, I noticed that I uploaded 101%.

![image](https://user-images.githubusercontent.com/5290572/97743226-96aee300-1aa2-11eb-87c4-451e85c0ea83.png)

Will pointed out that Storage is supposed to ignore .git and .vscode folders when uploading as shown here:
https://github.com/microsoft/vscode-azurestorage/blob/4411ce77b379704a853e3eb59ce54494830ba86c/src/utils/uploadUtils.ts#L97

But you can see in my uploaded folders, both .vscode and .git were uploaded.
![image](https://user-images.githubusercontent.com/5290572/97743194-8ac32100-1aa2-11eb-8ba5-63875aa97271.png)

PR #921, opened 2021-01-27
Only ignore folders at the root level of the directory path being uploaded

wwlorey:
The code that gets the number of resources in a directory was using glob patterns. These patterns would ignore `.git` and `.vscode` directories at any level in the tree structure.

There was a discrepancy between this number and the number of resources uploaded by AzCopy because AzCopy does not recursively match the exclude path and does not provide a way to do so. From the AzCopy CLI:

> ```      
> --exclude-path string          Exclude these paths when copying. This option does not support wildcard characters (*).  Checks relative path prefix(For example: myFolder;myFolder/subDirName/file.pdf)...
> ```

So I decided to only ignore `.git` and `.vscode` directories at the root level since we can't ignore them anywhere else in the tree.

Here's what it looks like to upload a single git repo followed by a folder with multiple repos in it. The old version of uploading `multiRepoTest` was waay off because none of the git files in each repo were included in the total.

| Old | New |
| - | - |
| Screen Shot 2021-01-26 at 4 29 38 PM | Screen Shot 2021-01-26 at 4 31 29 PM |

Fixes https://github.com/microsoft/vscode-azurestorage/issues/873

### DIFF
--- a/src/utils/uploadUtils.ts
+++ b/src/utils/uploadUtils.ts
@@ -100,7 +100,9 @@ export async function getDestinationDirectory(destinationDirectory?: string): Pr
 
 async function getNumResourcesInDirectory(directoryPath: string, countFolders?: boolean): Promise<number> {
     const options: readdirp.ReaddirpOptions = {
-        directoryFilter: ['!.git', '!.vscode'],
+        directoryFilter: (entryInfo: readdirp.EntryInfo) => {
+            return entryInfo.path !== '.git' && entryInfo.path !== '.vscode';
+        },
         type: countFolders ? 'files_directories' : 'files'
     };
     const resources: readdirp.EntryInfo[] = await readdirp.promise(directoryPath, options);


src/utils/uploadUtils.ts
```diff
@@ -100,7 +100,9 @@ export async function getDestinationDirectory(destinationDirectory?: string): Pr
 
 async function getNumResourcesInDirectory(directoryPath: string, countFolders?: boolean): Promise<number> {
     const options: readdirp.ReaddirpOptions = {
-        directoryFilter: ['!.git', '!.vscode'],
+        directoryFilter: (entryInfo: readdirp.EntryInfo) => {
```
ejizba, line 103: Any chance azcopy would just give us this number? Rather than trying to match it up with different glob implementations and getting it wrong a few times...

wwlorey: Yes there is and it's a very easy change 🥲

PR #922, opened 2021-01-27
Always use AzCopy to get total resources to upload

wwlorey:
Fixes https://github.com/microsoft/vscode-azurestorage/issues/873

### DIFF
--- a/src/commands/uploadFolder.ts
+++ b/src/commands/uploadFolder.ts
@@ -47,7 +47,6 @@ export async function uploadFolder(
 
     try {
         if (notificationProgress && cancellationToken) {
-            // AzCopy recognizes folders as a resource when uploading to file shares. So only set `countFoldersAsResources=true` in that case
             await uploadLocalFolder(actionContext, treeItem, sourcePath, destPath, notificationProgress, cancellationToken, destPath);
         } else {
             const title: string = getUploadingMessageWithSource(sourcePath, treeItem.label);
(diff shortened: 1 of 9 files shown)


src/commands/uploadFolder.ts
```diff
@@ -48,11 +48,11 @@ export async function uploadFolder(
     try {
         if (notificationProgress && cancellationToken) {
             // AzCopy recognizes folders as a resource when uploading to file shares. So only set `countFoldersAsResources=true` in that case
-            await uploadLocalFolder(actionContext, treeItem, sourcePath, destPath, notificationProgress, cancellationToken, destPath, treeItem instanceof FileShareTreeItem);
+            await uploadLocalFolder(actionContext, treeItem, sourcePath, destPath, notificationProgress, cancellationToken, destPath);
```
nturinski, line 51: Can't you get of the `countFoldersAsResources` parameter now?

nturinski (commented):
I'm a little confused as to how this is working.  Isn't this the code that was reporting the progress?  https://github.com/microsoft/vscode-azurestorage/blob/main/src/commands/azCopy/azCopyTransfer.ts#L105-L109  It looks like that this was always using the `TranserStatus.status` preferentially and only relying on the passed in total work if it couldn't find it in the status.

wwlorey:
> Isn't this the code that was reporting the progress? https://github.com/microsoft/vscode-azurestorage/blob/main/src/commands/azCopy/azCopyTransfer.ts#L105-L109

This has been confusing me for a bit lol. So that check was originally to make sure that no progress was shown until we were sure that `totalWork` existed. But you're right that it clobbers whatever `totalWork` may have been passed when creating the `TransferProgress`. So I can totally remove `totalWork` as a class member

wwlorey:
Let me know if you want me to break this up into multiple PRs - that latest commit introduced a lot of changes

ejizba (approved):

Issue #917, opened 2021-01-14
Detect if libsecret-1-dev is installed prior to running AzCopy commands


robidas:
Does this occur consistently?  Yes
Repro steps:

1. Follow the steps documented at https://docs.microsoft.com/en-us/learn/modules/automatic-update-of-a-webapp-using-azure-functions-and-signalr/7-exercise-host-a-static-website-using-a-storage-account
2.Do all steps up to and including "Deploy the web application to Azure Storage". This is the point where the failure occurs.

Action: azureStorage.deployStaticWebsite
Error type: Error
Error Message: AzCopy Transfer: "Failed".

Version: 0.11.0
OS: linux
OS Release: 4.4.0-19041-Microsoft
Product: Visual Studio Code
Product Version: 1.52.1
Language: en

<details>
<summary>Call Stack</summary>

```
extension.bundle.js:37:15387
extension.bundle.js:37:13040
n extension.bundle.js:37:12726
handleJob extension.bundle.js:37:14031
Object.<anonymous> extension.bundle.js:37:15435
fulfilled extension.bundle.js:37:12794
```

</details>

wwlorey:
@robidas are you using WSL or Remote SSH? Or is linux your host OS?

robidas:
WSL

Ubuntu 20.04.1 LTS (GNU/Linux 4.4.0-19041-Microsoft x86_64)
running on a Win10 pro guest vm, which in-turn is running on a win 10 pro hyper-v host

________________________________

wwlorey:
@robidas You can fix this by installing `libsecret-1-dev` in WSL:

```bash
sudo apt update
sudo apt install libsecret-1-dev
```

It looks like the packages installed by default in Ubuntu 20.04 on Windows are pretty minimal. This package appears to be included in standard Ubuntu.

PR #924, opened 2021-01-28
Show link to wiki when AzCopy fails (with no explanation) on Linux

wwlorey:
Related: https://github.com/microsoft/vscode-azurestorage/issues/917

After looking into it more, I was hesitant to auto-detect this package because you have to use that distro's package manager to query the installed packages. This varies based on the distro and keeping all of those commands up to date sounds like a can 'o worms 🪱

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,6 +62,16 @@ Use these configuration settings to customize the Azure Storage extension.
 * `azureStorage.file.showSavePrompt`: Set to `false` to prevent showing a warning dialog on File file save.
 * `azureStorage.blob.showSavePrompt`: Set to `false` to prevent showing a warning dialog on Blob file save.
 
+## Known Issues
+
+* `libsecret-1-dev` is required for AzCopy transfers on Linux and is not installed by default in many distros running in WSL.
+To install it on Debian based distros, run
+
+    ```
+    sudo apt update
+    sudo apt install libsecret-1-dev
+    ```
+
 <!-- region exclude-from-marketplace -->
 
 ## Contributing
--- a/README.md
+++ b/README.md
@@ -62,16 +62,6 @@ Use these configuration settings to customize the Azure Storage extension.
 * `azureStorage.file.showSavePrompt`: Set to `false` to prevent showing a warning dialog on File file save.
 * `azureStorage.blob.showSavePrompt`: Set to `false` to prevent showing a warning dialog on Blob file save.
 
-## Known Issues
-
-* `libsecret-1-dev` is required for AzCopy transfers on Linux and is not installed by default in many distros running in WSL.
-To install it on Debian based distros, run
-
-    ```
-    sudo apt update
-    sudo apt install libsecret-1-dev
-    ```
-
 <!-- region exclude-from-marketplace -->
 
 ## Contributing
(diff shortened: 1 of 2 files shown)


nturinski:
Were you passing which into the shell?  Have you looked into using this package? https://www.npmjs.com/package/which

wwlorey:
> Were you passing which into the shell? Have you looked into using this package? https://www.npmjs.com/package/which

Unfortunately `which` doesn't detect if this package is installed (I guess it's not in the path.) Here's an example of what querying the package database looks like on Kali with `libsecret-1-dev` installed:

Screen Shot 2021-01-28 at 1 27 32 PM

README.md
```diff
@@ -62,6 +62,16 @@ Use these configuration settings to customize the Azure Storage extension.
 * `azureStorage.file.showSavePrompt`: Set to `false` to prevent showing a warning dialog on File file save.
 * `azureStorage.blob.showSavePrompt`: Set to `false` to prevent showing a warning dialog on Blob file save.
 
+## Known Issues
+
+* `libsecret-1-dev` is required for AzCopy transfers on Linux and is not installed by default in many distros running in WSL.
```
ejizba, line 67: Is this how Storage Explorer handles this as well? Just by documenting it?

ejizba: > you have to use that distro's package manager to query the installed packages

Fwiw, I agree this sounds like a can of worms not worth opening. Just curious how SE handles it

wwlorey: They include libsecret as a dependency when creating their snap package. I also looked at the AzCopy repo and didn't find any guidance on downloading required dependencies on WSL

ejizba: And I assume that doesn't help us at all? Can we check telemetry and see how frequent this problem is? It seems like a big deal considering how much we rely on AzCopy now.

In terms of "ideas that don't actually fix it", what if we catch the error on Linux and point them to a "Troubleshooting" page or something (could just be in our GitHub Wiki).

nturinski: Yeah, I prefer to have at least something like that.  I feel like the visibility on the README is pretty low as in users aren't likely to look at it.

wwlorey: I agree that the README visibility isn't great. I'll add a add a troubleshooting notification on linux if we see the AzCopy failed error

wwlorey:
Here's what the AzCopy fail error looks like on linux:

Screen Shot 2021-01-29 at 3 30 17 PM

src/commands/azCopy/azCopyTransfer.ts
```diff
@@ -75,6 +75,10 @@ async function handleJob(context: IActionContext, jobInfo: IJobInfo, transferLab
             // tslint:disable-next-line: no-floating-promises
             ext.ui.showWarningMessage(message);
         } else {
+            if (/^AzCopy Transfer: "Failed"\.\s$/i.test(message) && process.platform === 'linux') {
+                message += localize('viewTheWiki', 'View [the wiki](https://github.com/microsoft/vscode-azurestorage/wiki/Known-Issues) for help with known issues.');
```
ejizba, line 79: The message here should be more about _what_ they should should view rather then _where_ they should view it. Who knows, maybe we move the content out of the wiki!

I'd suggest
```
View help with [known issues]().
```

Or
```
View [troubleshooting tips]().
```

ejizba, line 79: Use an aka.ms link

ejizba (approved):

Issue #62, opened 2018-02-07
Become eslint clean


StephenWeatherford:
- [x] Add tslint to travis
- [ ] Enable all rules

StephenWeatherford:
-[ ] Add tslint to travis

nturinski:
Tslint is deprecated

wwlorey:
Fixed by https://github.com/microsoft/vscode-azurestorage/pull/925

PR #925, opened 2021-02-02
Switch to ESLint

wwlorey:
I based `.eslintrc.js` on Docker's configuration. I made a few changes to what they had and I can list those out if you want to see them but I don't think anything was too controversial. 

Also I didn't add a rule enforcing `strict-boolean-expressions` because I found that one particularly annoying in TSLint. I can find the ESLint equivalent if needed.

There are some residual TSLint rule exceptions that need to be cleaned up and I'll take care of those in a separate PR.

Related: https://github.com/microsoft/vscode-azurestorage/issues/62


.vscode/extensions.json
```diff
@@ -1,6 +1,6 @@
 {
```
ejizba, line 1: What about "settings.json" and "tasks.json"? I feel like there's stuff to change there, too.

.eslintignore
```diff
@@ -0,0 +1,3 @@
+test/
```
ejizba, line 1: I like disabling a lot of rules for the "test" folder, but I'm not sure I'd go so far as to say I want to disable them all. You'll notice that's why we've have a separate "tslint.json" in the "test" folder that's pretty much only used to turn off rules.

gulpfile.ts
```diff
@@ -35,8 +37,28 @@ async function setAzCopyExePermissions(): Promise<void> {
     }
 }
 
+function lint(): void {
```
ejizba, line 40: Mmm looks like the Docker team went a lot more hard-core on gulp than we have. Unless we explicitly decide to do the same, I'd prefer if you stick with the simple npm script route. Documented in a lot of places, like here: https://code.visualstudio.com/api/advanced-topics/tslint-eslint-migration#eslint-run

package.json
```diff
@@ -854,15 +855,21 @@
...
         "gulp": "^4.0.0",
+        "gulp-eslint": "^6.0.0",
+        "gulp-if": "^3.0.0",
         "mocha": "^7.1.1",
         "mocha-junit-reporter": "^1.18.0",
         "mocha-multi-reporters": "^1.1.7",
         "ts-node": "^7.0.1",
-        "tslint": "^5.7.0",
-        "tslint-microsoft-contrib": "5.0.1",
+        "tslint": "^6.1.3",
```
ejizba, line 871: Can't we get rid of these tslint dependencies now?

package.json
```diff
@@ -873,6 +880,7 @@
     "dependencies": {
         "@azure-tools/azcopy-node": "1.0.0",
         "@azure/arm-storage": "^15.0.0",
+        "@azure/core-paging": "^1.1.3",
```
ejizba, line 883: Not sure this is worth an explicit dependency. Looks like it's only referenced once, and I think you could just refactor to be like this so we don't explicitly import `PageSettings`:
```typescript
let response: AsyncIterableIterator<azureStorageBlob.ContainerListBlobHierarchySegmentResponse> = containerClient.listBlobsByHierarchy(path.posix.sep, { prefix }).byPage({
    continuationToken,
    // https://github.com/Azure/Azurite/issues/605
    maxPageSize: parent.root.isEmulated ? maxPageSize * 10 : maxPageSize
});
```

wwlorey: I removed it as a dependency and it wasn't an issue after switching to the recommended ESLint rules 👍

src/tree/StorageAccountTreeItem.ts
```diff
@@ -279,15 +280,15 @@ export class StorageAccountTreeItem extends AzureParentTreeItem<IStorageRoot> {
         await this.ensureHostingCapable(context, websiteHostingStatus);
 
         if (!websiteHostingStatus.enabled) {
-            window.showInformationMessage(`Account '${this.label}' does not currently have static website hosting enabled.`);
+            void window.showInformationMessage(`Account '${this.label}' does not currently have static website hosting enabled.`);
```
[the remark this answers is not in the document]
wwlorey: Yessir. From ESLint:

Screen Shot 2021-02-04 at 6 00 58 PM

I think it's a lot more slick than an icky "eslint-disable" comment 🤢

package.json
```diff
@@ -835,8 +835,8 @@
         "cleanReadme": "gulp cleanReadme",
         "compile": "tsc -watch",
         "package": "vsce package --githubBranch main",
-        "lint": "tslint --project tsconfig.json -t verbose",
-        "lint-fix": "tslint --project tsconfig.json -t verbose --fix",
+        "lint": "eslint -c .eslintrc.js --ext .ts src && eslint -c .eslintrc.js --ext .ts test",
```
wwlorey, line 838: I originally updated this command to be `eslint -c .eslintrc.js --ext .ts {src,test}`.

But when the checks in Azure Pipelines ran, Windows and Linux threw this error: `No files matching the pattern "{src,test}" were found.`

Do you know of a more elegant, cross platform way to do this? I didn't know glob pattern interpretation varied like that...

src/commands/storageAccountActionHandlers.ts
```diff
@@ -105,7 +105,7 @@ async function runPreDeployTask(deployFsPath: string, context: IActionContext):
         const preDeployTask: vscode.Task | undefined = tasks.find((task: vscode.Task) => isTaskEqual(taskName, deployFsPath, task));
         if (preDeployTask) {
             await vscode.tasks.executeTask(preDeployTask);
-            await new Promise((resolve: () => void, reject: (error: Error) => void): void => {
+            await new Promise((resolve: (value?: unknown) => void, reject: (error: Error) => void): void => {
```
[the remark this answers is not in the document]
wwlorey: I was tryin to get ESLint off my back

Screen Shot 2021-02-05 at 2 20 51 PM

ejizba (approved):

Issue #929, opened 2021-02-07
[Suggestion] It would be better to add hyperlink to warning for using Azurite extension


v-mengwe:
**OS:** Win10
**Build Version:** 20210206.4

**Repro Steps:**

1. Make sure there is no Azurite extension installed.
2. Expand Local Emulator -> Right click the "Blob Containers" node -> Select "Create Blob Container...".
3. Check the result.

**Actual:**
There is an error "...reason: connect ECONNREFUSED 127.0.0.1:10000".
![image](https://user-images.githubusercontent.com/62053665/107140831-e2a1db80-695f-11eb-88d1-e6dc9a87f5b7.png)

**Suggestion:**
It would be better to add hyperlink to warning for using Azurite extension, such as:
![image](https://user-images.githubusercontent.com/62053665/107141539-4201ea80-6964-11eb-92d6-9dbfb23db246.png)

**More Info:**
This issue also reproduces for the "Queue" node of Local Emulator.
![image](https://user-images.githubusercontent.com/62053665/107140892-4c21ea00-6960-11eb-954e-cf470f4963c3.png)

wwlorey:
If you expand "Blob Containers" and click "Start Blob Emulator" (without Azurite installed) a notification to install Azurite is shown.

![image](https://user-images.githubusercontent.com/22795803/107264405-cb163180-69f7-11eb-87c0-de5708755c70.png)

![image](https://user-images.githubusercontent.com/22795803/107264428-ce112200-69f7-11eb-81e7-f73fca6d64b0.png)

But I think it's useful to throw a more helpful error if you try to create a blob container without starting the emulator.

Issue #928, opened 2021-02-07
[Suggestion] It would be better to change the button from "Install Azurite" to "View in Marketplace" when prompting to install Azurite extension


v-mengwe:
**OS:** Win10
**Build Version:** 20210206.4

**Repro Steps:**

1. Make sure there is no Azurite extension installed.
2. Expand Local Emulator -> Blob Containers -> Click "Start Blob Emulator".
3. Click "Install Azurite" button in the dialog.

**Actual:**
It navigates to Azurite extension page in marketplace, not install the extension directly.
![image](https://user-images.githubusercontent.com/62053665/107140431-ca7c8d00-695c-11eb-96f4-03323c34ff26.png)

**Suggestion:**
It would be better to change the button from "Install Azurite" to "View in Marketplace" when prompting to install Azurite extension.

nturinski:
Will be fixed by https://github.com/microsoft/vscode-azurestorage/issues/929

PR #934, opened 2021-02-11
Update 'must install Azurite' prompt

wwlorey:
This also adds an install prompt if you try to create an emulated queue or blob container without Azurite installed.

Fixes https://github.com/microsoft/vscode-azurestorage/issues/928
Fixes https://github.com/microsoft/vscode-azurestorage/issues/929

### DIFF
--- a/src/commands/blob/blobContainerGroupActionHandlers.ts
+++ b/src/commands/blob/blobContainerGroupActionHandlers.ts
@@ -5,12 +5,12 @@
 
 import { IActionContext, registerCommand } from 'vscode-azureextensionui';
 import { BlobContainerGroupTreeItem } from '../../tree/blob/BlobContainerGroupTreeItem';
+import { isAzuriteInstalled, warnAzuriteNotInstalled } from '../../utils/azuriteUtils';
 import { createChildNode } from '../commonTreeCommands';
-import { isAzuriteCliInstalled, isAzuriteExtensionInstalled, warnAzuriteNotInstalled } from '../startEmulator';
 
 export function registerBlobContainerGroupActionHandlers(): void {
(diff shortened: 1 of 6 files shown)


src/commands/blob/blobContainerGroupActionHandlers.ts
```diff
@@ -6,7 +6,13 @@
 import { IActionContext, registerCommand } from 'vscode-azureextensionui';
 import { BlobContainerGroupTreeItem } from '../../tree/blob/BlobContainerGroupTreeItem';
 import { createChildNode } from '../commonTreeCommands';
+import { isAzuriteCliInstalled, isAzuriteExtensionInstalled, warnAzuriteNotInstalled } from '../startEmulator';
 
 export function registerBlobContainerGroupActionHandlers(): void {
-    registerCommand("azureStorage.createBlobContainer", async (context: IActionContext, treeItem?: BlobContainerGroupTreeItem) => await createChildNode(context, BlobContainerGroupTreeItem.contextValue, treeItem));
+    registerCommand("azureStorage.createBlobContainer", async (context: IActionContext, treeItem?: BlobContainerGroupTreeItem) => {
+        if (treeItem?.root.isEmulated && !isAzuriteExtensionInstalled() && !(await isAzuriteCliInstalled())) {
```
nturinski, line 13: nit: Might be worth to just have a utility function that accepts a treeItem and checks for if it's emulated and if Azurite is installed

nturinski (approved):

src/commands/blob/blobContainerGroupActionHandlers.ts
wwlorey: I had trouble coming up a clean name for a function that would check `isEmulated` and `isAzuriteInstalled` so I just created an `isAzuriteInstalled` function. I also moved some things around into a new file `azuriteUtils.ts`

PR #959, opened 2021-03-29
Deploy static website project

v-wuzhai:


test/nightly/deploy.test.ts
```diff
@@ -0,0 +1,55 @@
...
+});
+
+// The workspace folder that vscode is opened against for tests
+function getWorkspacePath(testWorkspaceName: string): string {
+    let workspacePath: string = '';
+    const workspaceFolders: readonly vscode.WorkspaceFolder[] | undefined = vscode.workspace.workspaceFolders;
+    if (!workspaceFolders || workspaceFolders.length === 0) {
+        throw new Error("No workspace is open");
+    } else {
+        for (const obj of workspaceFolders) {
```
wwlorey, line 47: nit: Renaming `obj` to something like `folder` would be more descriptive

test/test.code-workspace
```diff
@@ -1,7 +1,7 @@
 {
     "folders": [
         {
-            "path": "../testWorkspace"
+            "path": "./nightly/testFolder/vue-realworld-example-app"
```
wwlorey, line 4: The current sample project has a lot of files. I think using https://github.com/Azure-Samples/html-docs-hello-world would be good. It's smaller and less complex

ejizba: Honestly you could get away with a single `index.html` file with something like this:
```
<!DOCTYPE html>
<html lang="en">
<head>
</head>
<body>
    <h1>Hello World!</h1>
</body>
</html>
```

test/nightly/deploy.test.ts
```diff
@@ -53,3 +52,18 @@ function getWorkspacePath(testWorkspaceName: string): string {
         return workspacePath;
     }
 }
+
+// Polling to send the request within the maximum time
+async function validateWebSite(webUrl: string | undefined, client: ServiceClient, maximumValidationMs: number, pollingMs: number) {
+    const startTime: number = Date.now() + maximumValidationMs;
```
wwlorey, line 58: I believe this should be called `endTime`

test/nightly/deploy.test.ts
```diff
@@ -0,0 +1,69 @@
...
+        if (!longRunningTestsEnabled) {
+            this.skip();
+        }
+    });
+
+    test("deployStaticWebsite", async () => {
+        const resourceName: string = getRandomHexString().toLowerCase();
+        resourceGroupsToDelete.push(resourceName);
+        const context: IActionContext = { telemetry: { properties: {}, measurements: {} }, errorHandling: { issueProperties: {} }, ui: testUserInput, valuesToMask: [] };
+        const testFolderPath: string = getWorkspacePath('html-docs-hello-world');
```
wwlorey, line 28: You don't have to use the same name as what I suggested since we're only uploading one file now. `html-hello-world` would be fine

wwlorey (approved):

PR #972, opened 2021-05-05
[AzureStack] Set azcopy default service api version to 2019-02-02 to support Azure Stack

rakku-ms:
… Stack


src/utils/azureClients.ts
```diff
@@ -11,6 +11,8 @@ import { createAzureClient, ISubscriptionContext } from 'vscode-azureextensionui
 
 export async function createStorageClient<T extends ISubscriptionContext>(context: T): Promise<StorageManagementClient> {
     if (context.isCustomCloud) {
+        // set API version for Azure Stack
```
wwlorey, line 14: Is there a GitHub issue, blog post, or something you can comment here so we can track where this version came from?

rakku-ms: I picked the service api version ("2019-02-02") used in azcopy version 10.5.0 which works against Azure Stack without any setting. @bganapa should I use "2019-06-01" as given here? https://github.com/Azure/azure-rest-api-specs/blob/a394eccc2ce26369cacbc113cf0c3a9c9501e689/profile/2020-09-01-hybrid.json#L440

rakku-ms: also, this azcopy issue against Azure Stack is known - https://docs.microsoft.com/en-us/azure-stack/user/azure-stack-storage-transfer?view=azs-2008&tabs=az1#azcopy-101-configuration-and-limits

bganapa: Yes , we could use 2019-06-01 if azcopy is already supporting the version. If not 2019-02-02 is fine

rakku-ms: **2019-06-01** is not working against Azure Stack. so I will keep 2019-02-02 in this PR
<pre>
2021/05/06 15:17:08 ==> REQUEST/RESPONSE (Try=1/365.0483ms, OpTime=1.4831594s) -- RESPONSE STATUS CODE ERROR
   PUT https://testsa004.blob.northwest.azs-longhaul-03.selfhost.corp.microsoft.com/bc0041/3e898e9a-91ca-ab40-41f5-504099b5f16d.log?se=2021-05-09t15%3A17%3A07z&sig=-REDACTED-&sp=rwl&srt=co&ss=bf&sv=2019-07-07&timeout=901
   Content-Length: [6586]
   User-Agent: [azcopy-node, 2.0.0, win32,  AzCopy/10.10.0 Azure-Storage/0.13 (go1.15; Windows_NT)]
   X-Ms-Blob-Cache-Control: []
   X-Ms-Blob-Content-Disposition: []
   X-Ms-Blob-Content-Encoding: []
   X-Ms-Blob-Content-Language: []
   X-Ms-Blob-Content-Md5: []
   X-Ms-Blob-Content-Type: [text/plain]
   X-Ms-Blob-Type: [BlockBlob]
   X-Ms-Client-Request-Id: [1c18e1a8-2f5a-4d26-5c76-e379d47b50c0]
   X-Ms-Version: <b>[2019-06-01]</b>
   --------------------------------------------------------------------------------
   RESPONSE Status: 400 The value for one of the HTTP headers is not in the correct format.
   Content-Length: [328]
   Content-Type: [application/xml]
   Date: [Thu, 06 May 2021 15:17:08 GMT]
   Server: [Microsoft-HTTPAPI/2.0]
   X-Ms-Request-Id: [00fa0377-0c81-0856-484a-63c1abcabb41]
Response Details: ﻿<Code>InvalidHeaderValue</Code><Message>The value for one of the HTTP headers is not in the correct format. </Message><HeaderName>x-ms-version</HeaderName><HeaderValue>2019-06-01</HeaderValue>
</pre>

wwlorey: Is there documentation somewhere about why "2019-02-02" works over other api versions? It would be nice to track that in this change

bganapa: @wwlorey Thanks for looking in to this PR

In the following page which is under review
https://review.docs.microsoft.com/en-us/azure-stack/user/azure-stack-profiles-azure-resource-manager-versions?view=azs-1802&branch=pr-en-us-7213

It is menioned that for Microsoft.Storage we support the api version '2019-06-06' . Typically we mention only the highest api version supported and do not mention all the api versions lower. It is assumed that the lower versions are supported.

In this case, the azcopy jumped to supporting 2019-12-12 from 2019-02-02  without supporting 2019-06-06. Hence we are in need of 2019-02-02 version here

Blame on the following line shows that https://github.com/Azure/azure-storage-azcopy/blob/067b2fc3f79aca43a378ee1a6761268c7d1e4f38/common/environment.go#L320

bganapa (approved):

wwlorey (approved):

PR #974, opened 2021-05-13
Bump azcopy-node

wwlorey:


wwlorey:
Discussed offline with Jing: since storage isn't going to ship until after Build, we might as well bump this package now since we're several versions behind.

MRayermannMSFT:
tenor

nturinski (approved):

PR #976, opened 2021-05-25
Update check for showing 'known issues' link (AzCopy)

wwlorey:
`jobInfo.errorMessage` didn't used to be set for this type of failure. This check should cover more unexpected failures than what we were looking for previously. But I wouldn't be surprised if this error message behavior changes again

Screen Shot 2021-05-25 at 3 59 51 PM

### DIFF
--- a/src/commands/azCopy/azCopyTransfer.ts
+++ b/src/commands/azCopy/azCopyTransfer.ts
@@ -62,8 +62,7 @@ function handleJob(context: IActionContext, jobInfo: IJobInfo, transferLabel: st
             // Add an additional error log since we don't have any more info about the failure
             ext.outputChannel.appendLog(localize('couldNotTransfer', 'Could not transfer "{0}"', transferLabel));
 
-            const isDefaultFailMessage: boolean = !finalTransferStatus?.ErrorMsg && finalTransferStatus?.JobStatus === 'Failed';
-            if (isDefaultFailMessage && process.platform === 'linux') {
+            if (finalTransferStatus?.JobStatus === 'Failed' && process.platform === 'linux') {
                 message += localize('viewHelp', ' View help with [known issues](https://aka.ms/AAb0i6o).');
             }
         }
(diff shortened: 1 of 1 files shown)


src/commands/azCopy/azCopyTransfer.ts
```diff
@@ -62,9 +62,9 @@ function handleJob(context: IActionContext, jobInfo: IJobInfo, transferLabel: st
             // Add an additional error log since we don't have any more info about the failure
             ext.outputChannel.appendLog(localize('couldNotTransfer', 'Could not transfer "{0}"', transferLabel));
 
-            const isDefaultFailMessage: boolean = !jobInfo.errorMessage && finalTransferStatus?.JobStatus === 'Failed';
+            const isDefaultFailMessage: boolean = !finalTransferStatus?.ErrorMsg && finalTransferStatus?.JobStatus === 'Failed';
```
ejizba, line 65: > But I wouldn't be surprised if this error message behavior changes again

You're not exactly filling me with confidence in your fix. Is there something we can do to make this more robust?

For example:
1. Why do we need to check `jobInfo.errorMessage` or `finalTransferStatus?.ErrorMsg` anyways?
2. Maybe add unit tests

wwlorey: > Why do we need to check jobInfo.errorMessage or finalTransferStatus?.ErrorMsg anyways?

When libsecret is not installed, AzCopy used to fail with no error message at all. So both `jobInfo.errorMessage` and `finalTransferStatus.ErrorMsg` were empty. This behavior changed and now a `jobInfo.errorMessage` is provided in that case, while `finalTransferStatus.ErrorMsg` is still empty. I figured that checking `finalTransferStatus.ErrorMsg` would be a more robust way of determining if we hit libsecret problem after this change, because that error message has always been empty in that failure case.

My concern is that they could change the error message behavior again though.. And a unit test would be good. Do you think it's worth running tests on a special linux instance just to check this?

ejizba: Yeah but your troubleshooting link is pretty generic. Is there a downside if you _don't_ check `!finalTransferStatus?.ErrorMsg` and just display it more often?

wwlorey: I like that option, especially since the linked wiki page is real short

ejizba (approved):

PR #978, opened 2021-06-04
Swallow 'website disabled' error in deploy test

wwlorey:
After we deploy to static site and check the result, the site is rarely configured & active yet. `client.sendRequest` started throwing a "website disabled" error but if you just wait a bit the site works as expected (anywhere between a few seconds to 1min in my testing)

### DIFF
--- a/test/nightly/deploy.test.ts
+++ b/test/nightly/deploy.test.ts
@@ -59,10 +59,21 @@ async function validateWebSite(webUrl: string | undefined, client: ServiceClient
     let response: HttpOperationResponse;
     // eslint-disable-next-line no-constant-condition
     while (true) {
-        response = await client.sendRequest({ method: 'GET', url: webUrl });
-        if (Date.now() > endTime || response.status == 200) {
-            break;
+        try {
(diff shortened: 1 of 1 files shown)


test/nightly/deploy.test.ts
```diff
@@ -59,10 +59,21 @@ async function validateWebSite(webUrl: string | undefined, client: ServiceClient
...
-        if (Date.now() > endTime || response.status == 200) {
-            break;
+        try {
+            response = await client.sendRequest({ method: 'GET', url: webUrl });
+            if (Date.now() > endTime || response.status == 200) {
+                break;
+            }
+        } catch (error) {
+            // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+            if (error?.statusCode === 404 && /websitedisabled/i.test(error?.body)) {
```
ejizba, line 69: Honestly I'd probably just ignore all errors here. They're probably transient and as long as we get a 200 eventually that's all that really matters for the test

ejizba (approved):

PR #986, opened 2021-06-16
Implement workspace.fs for fse commands

nturinski:
As discussed offline, I want to try implementing the workspace.fs stuff in Storage first as it has some basic workspace.fs commands that we can test.


src/utils/workspaceFsUtils.ts
```diff
@@ -0,0 +1,61 @@
...
+        } catch { /*ignore*/ }
+        if (stats && stats.type === FileType.File) return;
+
+        const dir: string = dirname(path);
+        const folder = Uri.file(dir);
+        try {
+            if ((await workspace.fs.stat(folder)).type !== FileType.Directory) {
+                // parent is not a directory
+                // This is just to cause an internal ENOTDIR error to be thrown
+                await workspace.fs.readDirectory(folder);
```
neelip, line 25: This line with `...readDirectory(...)`doesn't make sense if the folder is not a directory.

src/utils/workspaceFsUtils.ts
```diff
@@ -0,0 +1,61 @@
...
+import { parseError } from 'vscode-azureextensionui';
+
+export namespace workspaceFsUtils {
+    export async function ensureFile(path: string): Promise<void> {
+        let stats: FileStat | undefined;
+        const file = Uri.file(path);
+        try {
+            stats = await workspace.fs.stat(file);
+        } catch { /*ignore*/ }
+        if (stats && stats.type === FileType.File) return;
```
neelip, line 17: I think this can be replaced by `pathExists()`.

nturinski: I was thinking about that too, but I need the stats that come with doing workspace.fs.stat().

src/utils/workspaceFsUtils.ts
nturinski: Sure can! 

If we are at this part of the code, that means the file doesn't exist because it would have returned at L17.  Then we check to see if the dirname of the file we're ensuring is a folder or not.  If it is, it skips the if statement and then creates the file in the folder.

If it is not, we are going to call `readDirectory` specifically to cause an error to be thrown.  If the error that is thrown is `FileNotFound` then we know that we need to create a folder with the given dirname, and then we can create the file in the new directory that was just created.

If the error is something else, then we'll just throw it.

src/utils/workspaceFsUtils.ts
```diff
@@ -0,0 +1,61 @@
...
+
+        const dir: string = dirname(path);
+        const folder = Uri.file(dir);
+        try {
+            if ((await workspace.fs.stat(folder)).type !== FileType.Directory) {
```
ejizba, line 22: I feel like we should add `ensureDir` and use it here in `ensureFile` to make sure the parent exists

src/utils/workspaceFsUtils.ts
```diff
@@ -0,0 +1,61 @@
...
+        await workspace.fs.writeFile(file, new Uint8Array());
+    }
+
+    export async function readFile(path: string): Promise<Uint8Array> {
+        const file = Uri.file(path);
+        return await workspace.fs.readFile(file);
+    }
+
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any, @typescript-eslint/explicit-module-boundary-types
+    export async function writeFile(path: string, contents: any): Promise<void> {
```
ejizba, line 47: I don't think `contents` should be `any`. I say we enforce `string` until we come across someone using something else. I'd also be fine if we allow `Uint8Array` or `Buffer`, but I don't think we ever use that honestly

src/utils/workspaceFsUtils.ts
```diff
@@ -0,0 +1,61 @@
...
+        const file = Uri.file(path);
+        await workspace.fs.writeFile(file, new Uint8Array(contents));
```
ejizba, line 49: I think `Buffer.from()` is preferred over `new Uint8Array()`

src/utils/workspaceFsUtils.ts
ejizba: I don't think the functionality of `ensureFile` is complex enough that we need to base it off someone else. If you can come up with a cleaner way to do this, I would prefer that. VS Code behavior (including how errors are handled) is probably slightly different than fs anyways

src/utils/workspaceFsUtils.ts
```diff
@@ -0,0 +1,61 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { dirname } from 'path';
+import { FileStat, FileType, Uri, workspace } from 'vscode';
+import { parseError } from 'vscode-azureextensionui';
+
+export namespace workspaceFsUtils {
```
ejizba, line 10: Assuming this will eventually be in the UI package, we should match the naming in that package. I would suggest something along the lines of `AzExtWorkspaceFs` or `AzExtFsExtra`

gulpfile.ts
```diff
@@ -4,25 +4,25 @@
...
-import * as fse from 'fs-extra';
 import * as gulp from 'gulp';
 import * as path from 'path';
 import { gulp_installAzureAccount, gulp_installResourceGroups, gulp_webpack } from 'vscode-azureextensiondev';
+import { workspaceFsUtils } from './src/utils/workspaceFsUtils';
 
 async function prepareForWebpack(): Promise<void> {
     const mainJsPath: string = path.join(__dirname, 'main.js');
-    let contents: string = (await fse.readFile(mainJsPath)).toString();
+    let contents: string = (await workspaceFsUtils.readFile(mainJsPath)).toString();
```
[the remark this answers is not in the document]
nturinski: Yeah, I was wondering about that.  So that means we can't totally remove the fs-extra dependency then?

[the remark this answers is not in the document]
ejizba: I feel like our options are
1. Use fs-extra, but at least make it a dev dependency
2. Just use Node.js's built-in fs 

I think I lean towards 2. Looks like we only need readFile and writeFile, so I bet we can figure that out without fs-extra

gulpfile.ts
```diff
@@ -4,25 +4,25 @@
...
         .replace('out/src/extension', 'dist/extension.bundle')
         .replace(', true /* ignoreBundle */', '');
-    await fse.writeFile(mainJsPath, contents);
+    fs.writeFileSync(mainJsPath, contents);
 }
 
 async function cleanReadme(): Promise<void> {
     const readmePath: string = path.join(__dirname, 'README.md');
-    let data: string = (await fse.readFile(readmePath)).toString();
+    let data: string = fs.readFileSync(readmePath).toString();
```
nturinski, line 23: I'm using sync because just `readFile` relied on callbacks rather than Promises.  I feel like in the gulpfile, it's fine to just use the synchronous code anyway.

src/utils/AzExtFsExtra.ts
```diff
@@ -0,0 +1,69 @@
...
+            } else {
+                throw err
+            }
+        }
+    }
+
+    export async function ensureFile(path: string): Promise<void> {
+        try {
+            // file exists so exit
+            if (await isFile(path)) return;
```
ejizba, line 41: It seems like you have some extra logic here to handle if `ensureFile` is trying to create a file that's actually already a directory. I think we want the same (but opposite) for `ensureDir`. Also what is the experience in that case? I think I would expect a specific VS Code error for that

src/utils/AzExtFsExtra.ts
```diff
@@ -0,0 +1,69 @@
...
+        } catch (err) {
+            const dir: string = dirname(path);
```
ejizba, line 43: I feel like you should be parsing the error here like you're doing for `ensureDir`

src/utils/AzExtFsExtra.ts
```diff
@@ -0,0 +1,69 @@
...
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { dirname } from 'path';
+import { FileStat, FileType, Uri, workspace } from 'vscode';
+import { parseError } from 'vscode-azureextensionui';
+
+export namespace AzExtFsExtra {
+    export async function isDirectory(path: string): Promise<boolean> {
+        const dir = Uri.file(path);
```
ejizba, line 12: How about we take in `Uri | string`, since you're basically doing `uri.fsPath` and then `Uri.file(path)` in a lot of cases. Might as well cut out the middle man when possible.

src/utils/AzExtFsExtra.ts
```diff
@@ -0,0 +1,69 @@
...
+        const stats = await workspace.fs.stat(dir);
+        return stats.type === FileType.Directory;
+    }
+
+    export async function isFile(path: string): Promise<boolean> {
+        const dir = Uri.file(path);
```
ejizba, line 18: I would name your variables something more generic like `uri` so that you don't run into copy/paste errors converting `dir` functions to `file` functions

test/AzExtFsExtra.test.ts
```diff
@@ -0,0 +1,91 @@
...
+import assert = require('assert');
+import * as os from 'os';
+import * as path from 'path';
+import { workspace, WorkspaceFolder } from 'vscode';
+import { AzExtFsExtra, getRandomHexString } from "../extension.bundle";
+
+
+suite('AzExtFsExtra', function (this: Mocha.Suite): void {
+    let workspacePath: string;
+    this.timeout(3 * 60 * 1000);
```
ejizba, line 15: I would hope these tests are super fast. Maybe this timeout was a copy/paste thing we don't actually need?

test/AzExtFsExtra.test.ts
```diff
@@ -0,0 +1,91 @@
...
+        const workspaceFolders: readonly WorkspaceFolder[] | undefined = workspace.workspaceFolders;
+        if (!workspaceFolders) {
+            throw new Error("No workspace is open");
+        }
+
+        workspacePath = workspaceFolders[0].uri.fsPath;
+    });
+
+    test('pathExists', async () => {
+        assert.strictEqual(await AzExtFsExtra.pathExists(workspacePath), true);
```
ejizba, line 26: I'd test one file and one dir to make sure both cases work

test/AzExtFsExtra.test.ts
```diff
@@ -0,0 +1,91 @@
...
+        assert.strictEqual(await AzExtFsExtra.pathExists('./path/does/not/exist'), false);
+    });
+
+    test('ensureDir that does not exist', async () => {
+        const fsPath = path.join(os.tmpdir(), `azExtFsExtra${getRandomHexString()}`);
```
ejizba, line 31: tmpdir files don't get cleaned up super fast, if at all. I'd prefer if you pick a root dir that you can remove during suiteTeardown

test/AzExtFsExtra.test.ts
```diff
@@ -0,0 +1,91 @@
...
+        assert.strictEqual(await AzExtFsExtra.pathExists(fsPath), false);
+        await AzExtFsExtra.ensureDir(fsPath);
+        assert.strictEqual(await AzExtFsExtra.pathExists(fsPath), true);
```
ejizba, line 34: This should probably verify that the path created is actually a directory

test/AzExtFsExtra.test.ts
```diff
@@ -0,0 +1,91 @@
...
+    });
+
+    test('isDirectory && !isFile', async () => {
+        const fsPath = path.join(os.tmpdir(), `azExtFsExtra${getRandomHexString()}`);
+        await AzExtFsExtra.ensureDir(fsPath);
```
ejizba, line 39: It's a bit dodgy to use `AzExtFsExtra` so much when testing `AzExtFsExtra`. Can you verify that every method you use has already been covered in a different test by itself? If not, we might want to look into using `fs` or something else for verification. For example, this test uses `ensureDir`, but the `ensureDir` test doesn't actually verify the thing it creates is a directory (as my other comment says). But now we're in a circular loop if the `ensureDir` test uses `isDirectory`

src/utils/AzExtFsExtra.ts
nturinski: Throws these errors:

image

test/AzExtFsExtra.test.ts
```diff
@@ -0,0 +1,139 @@
...
+    });
+
+    test('pathExists for file', async () => {
+        assert.strictEqual(await AzExtFsExtra.pathExists(workspaceFilePath), fs.existsSync(workspaceFilePath));
+        assert.strictEqual(await AzExtFsExtra.pathExists(nonExistingFilePath), fs.existsSync(nonExistingFilePath));
+    });
+
+    test('isDirectory properly detects folders', async () => {
+        assert.strictEqual(await AzExtFsExtra.isDirectory(workspacePath), isDirectoryFs(workspacePath));
+        assert.strictEqual(await AzExtFsExtra.isDirectory(workspaceFilePath), isDirectoryFs(workspaceFilePath));
```
ejizba, line 48: `isDirectory` is what this test is for, so we can use that from `AzExtFsExtra`. Since we know what the "expected" value should be, we can hard-code that to `true` and `false`
```suggestion
        assert.strictEqual(await AzExtFsExtra.isDirectory(workspacePath), true);
        assert.strictEqual(await AzExtFsExtra.isDirectory(workspaceFilePath), false);
```

test/AzExtFsExtra.test.ts
```diff
@@ -0,0 +1,139 @@
...
+        assert.strictEqual(await AzExtFsExtra.isFile(workspacePath), isFileFs(workspacePath));
+    });
+
+    test('ensureDir that does not exist', async () => {
+        const fsPath = path.join(os.homedir(), testFolderPath, getRandomHexString());
+        assert.strictEqual(await AzExtFsExtra.pathExists(fsPath), fs.existsSync(fsPath));
+        await AzExtFsExtra.ensureDir(fsPath);
+
+        assert.strictEqual(await AzExtFsExtra.isDirectory(fsPath), isDirectoryFs(fsPath));
+        assert.strictEqual(await AzExtFsExtra.pathExists(fsPath), fs.existsSync(fsPath));
```
ejizba, line 63: This test is for `ensureDir`, so that's the only thing we call from `AzExtFsExtra`. Then we use `fs` to get the "actual" results, and use `true` or `false` as the "expected" result.

```suggestion
        assert.strictEqual(fs.existsSync(fsPath), false);
        await AzExtFsExtra.ensureDir(fsPath);

        assert.strictEqual(isDirectoryFs(fsPath), true);
        assert.strictEqual(fs.existsSync(fsPath), true);
```

ejizba (commented):
The changes around using `fs` in the tests were not what I had in mind. First of all, I like hard-coding `true` or `false` as the expected result, so that should change back. Secondly, the only method from `AzExtFsExtra` we should call in each test is the specific method that test is for. If you find you want to use another method, _that_ is where `fs` comes in. I've added a few comments with suggested changes as examples.

test/AzExtFsExtra.test.ts
```diff
@@ -0,0 +1,149 @@
...
+    let testFolderPath: string;
+    let workspaceFilePath: string;
+
+    const indexHtml: string = 'index.html';
+
+    const nonExistingPath: string = ' ./path/does/not/exist';
+    const nonExistingFilePath = path.join(nonExistingPath, indexHtml);
+
+    suiteSetup(function (this: Mocha.Context): void {
+        testFolderPath = `azExtFsExtra${getRandomHexString()}`
```
ejizba, line 25: Just make this 
```
path.join(workspacePath, `azExtFsExtra${getRandomHexString()}`)
```

So you don't have to join with `workspacePath` every time. Also to make the name more accurate because currently it's really just `testFolderName`

ejizba (approved):

test/AzExtFsExtra.test.ts
```diff
@@ -0,0 +1,149 @@
...
+        if (!workspaceFolders) {
+            throw new Error("No workspace is open");
+        }
+
+        workspacePath = workspaceFolders[0].uri.fsPath;
+        workspaceFilePath = path.join(workspacePath, indexHtml);
+    });
+
+    suiteTeardown(async function (this: Mocha.Context): Promise<void> {
+        await workspace.fs.delete(Uri.file(path.join(workspacePath, testFolderPath)), { recursive: true })
```
ejizba, line 36: Missed one
```suggestion
        await workspace.fs.delete(Uri.file(testFolderPath), { recursive: true })
```

PR #993, opened 2021-08-11
Run Azurite in loose mode to workaround 'Failed with code "500"' error

wwlorey:
This issue would happen anytime you tried creating a new emulated blob. It's being tracked on Azurite [here](https://github.com/Azure/Azurite/issues/676). Until it's fixed there, we can workaround by running Azurite in loose mode.

### DIFF
--- a/src/commands/startEmulator.ts
+++ b/src/commands/startEmulator.ts
@@ -5,21 +5,27 @@
 
 import * as vscode from 'vscode';
 import { IActionContext } from 'vscode-azureextensionui';
-import { emulatorTimeoutMS } from '../constants';
+import { azuriteExtensionPrefix, azuriteLooseSetting, emulatorTimeoutMS } from '../constants';
 import { ext } from "../extensionVariables";
 import { isAzuriteCliInstalled, isAzuriteExtensionInstalled, warnAzuriteNotInstalled } from '../utils/azuriteUtils';
 import { cpUtils } from '../utils/cpUtils';
+import { updateWorkspaceSetting } from '../utils/settingsUtils';
 
 export async function startEmulator(context: IActionContext, emulatorType: EmulatorType): Promise<void> {
     if (isAzuriteExtensionInstalled()) {
         // Use the Azurite extension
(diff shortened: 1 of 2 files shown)


src/commands/startEmulator.ts
```diff
@@ -5,21 +5,27 @@
...
 import { cpUtils } from '../utils/cpUtils';
+import { updateWorkspaceSetting } from '../utils/settingsUtils';
 
 export async function startEmulator(context: IActionContext, emulatorType: EmulatorType): Promise<void> {
     if (isAzuriteExtensionInstalled()) {
         // Use the Azurite extension
-        await vscode.commands.executeCommand(`azurite.start_${emulatorType}`);
+
+        // Enable loose mode (workaround for https://github.com/Azure/Azurite/issues/676)
+        await updateWorkspaceSetting(azuriteLooseSetting, true, '', azuriteExtensionPrefix);
```
ejizba, line 19: Is there a corresponding issue on our repo? How did we discover this bug? The user can set this themselves, so I'm just trying to figure out if it's worth us hard-coding a workaround like this

wwlorey: I found this issue when I upgraded the @azure/storage-blob SDK to 12.7.0 and I didn't file an issue on this repo. Azurite extension users can set the loose setting themselves but we'd have to set it for people who use the CLI when we start the emulator for them.

Would you rather we catch the '500' error and ask people if they want to run in loose mode? Right now the error isn't very helpful:

Screen Shot 2021-08-12 at 2 54 42 PM

ejizba: So is the issue with the emulator or v12.7.0 of the blob sdk?

wwlorey: It's an Azurite issue. A new property `x-ms-encryption-algorithm` was added around v12.6.0 of the blob SDK and Azurite has yet to add support for that property. So Azurite throws the 500 error if it's included in any requests

wwlorey:
Closing in favor of https://github.com/microsoft/vscode-azurestorage/pull/996 per offline discussion

Issue #1006, opened 2021-12-02
Ditch the `azure-storage` package in favor of `@azure/storage-queue` and `@azure/data-tables`


wwlorey:
I was about to file an issue on [azure-storage](https://www.npmjs.com/package/azure-storage) related to their md5.js vulnerability but I'd forgotten that it's been deprecated:

> Please note, newer packages @azure/storage-blob, @azure/storage-queue and @azure/storage-file are available as of November 2019 and @azure/data-tables is available as of June 2021 for the individual services. While the legacy azure-storage package will continue to receive critical bug fixes, we strongly encourage you to upgrade.

When I did SDK upgrades in this extension a while ago, the queue and table packages weren't available yet. So we should switch to them now.

PR #1010, opened 2021-12-11
Switch from `azure-storage` to `@azure/storage-queue` and `@azure/data-tables` 

wwlorey:
This fixes GHSA-xx4c-jj58-r7x6, CVE-2021-3765, and the "md5.js" vulnerability leaving just the low severity "string_decoder" vulnerability for this extension 🥳

There are some annoying differences between the new SDKs as well as one problem I encountered with the tables SDK[^1]. But since queues and tables have such narrow support in this extension I don't think bugging the SDK team is worth it.

Fixes https://github.com/microsoft/vscode-azurestorage/issues/1006

[^1]:  The tables SDK doesn't throw any errors when problems arise while creating a new table, it just fails.


src/tree/queue/QueueGroupTreeItem.ts
```diff
@@ -108,24 +102,24 @@ export class QueueGroupTreeItem extends AzExtParentTreeItem {
...
-                } else {
-                    resolve(result);
-                }
-            });
-        });
+    private async createQueue(name: string): Promise<azureStorageQueue.QueueItem> {
+        const queueServiceClient = this.root.createQueueServiceClient();
+        await queueServiceClient.createQueue(name);
+
+        const queuesResponse: azureStorageQueue.ListQueuesSegmentResponse= await this.listQueues();
```
bwateratmsft, line 109: Nit:
```suggestion
        const queuesResponse: azureStorageQueue.ListQueuesSegmentResponse = await this.listQueues();
```

bwateratmsft (approved):

PR #1053, opened 2022-05-13
Merge unified into main

alexweininger:


nturinski (approved):
I noticed some places don't have the license files (my fault).  We should do a scan to make sure that everything has it.

Issue #1042, opened 2022-05-06
Show a warning "<workspacename> unable to resolve workspace folder" in the new opened File Explorer when the resource groups by Resource Type/Location/ARM Tag


v-xinda:
**OS:** All
**Azure Resource Extension Build Version:** 20220505.12 (unified)
**Azure Storage Extension Build Version:** 20220505.2 (unified)
**Regression:** Not a regression

**Repro Steps:**

1. Make sure the resource groups by **Resource Type/Location/ARM Tag**.
2. Expand a storage account -> Expand a Blob Containers -> Right click a blob container -> Click "Open in File Explorer..." -> Click "Open in new window".
3. Check whether shows all folders and blobs that are under the blob container in the new opened File Explorer.

**Expect:**
Show all folders and blobs that are under the blob container in the new opened File Explorer.

**Actual:**
Show a warning "<workspacename> unable to resolve workspace folder" in the new opened File Explorer.
![image](https://user-images.githubusercontent.com/57662131/167100830-cdbd595e-8ae5-4b8f-a58e-86f1134b0038.png)

**More Info:**

1. This issue also reproduces when opening a file share in File Explorer.
2. This issue also reproduces when selecting "Open in current window" or "Add to workspace".
3. This issue doesn't reproduce when the resource groups by Resource Group.

nturinski:
Should be fixed by fixing this https://github.com/microsoft/vscode-azurestorage/issues/1041

v-ruizh:
This issue can be reproduced on the latest main branch build 20220518.1. And it also reproduces when the resource groups by "Resource Group".

nturinski:
This bug is somewhat nasty.  I know for sure that it wasn't working because we weren't finding the tree item correctly unless it was in resourceGroup grouping.  After my fix, I think it broke again because we aren't resolving all of the children automatically anymore.  

To fix this bug, you have to open the appropriate group, let it resolve, and then refresh the tree item (so that its children appear).  Once you've done that, refresh the workspace, and everything works as expected.


It has decent usage and I would want to fix it, but I'm not really sure how we would.

nturinski:
What happens is, findTreeItem is called.  It automatically calls `getCachedChildren,` but since the StorageTreeItem Proxy isn't resolved yet, it returns an empty array.  We probably need to make findTreeItem also resolve the AppResource (unless we are looking for that node and are done)

bwateratmsft:
[...]

PR #1060, opened 2022-05-19
Resolve storage accounts before trying to find Blob/File children

nturinski:
Fixes https://github.com/microsoft/vscode-azurestorage/issues/1042

When `findTreeItem` is called. It automatically calls getCachedChildren, but since the StorageTreeItem Proxy isn't resolved yet, it returns an empty array. 

It's being called by AzureFSProvider to get the blob container to display the blob and its contents in the file explorer.  I have a util function that only returns if there is a storage account AppResource attached to the blob/file share and resolves it,  If it is already resolved, it's a no-op so I don't feel too bad about always checking/resolving.

### DIFF
--- a/src/AzureStorageFS.ts
+++ b/src/AzureStorageFS.ts
@@ -18,6 +18,7 @@ import { BlobTreeItem } from "./tree/blob/BlobTreeItem";
 import { DirectoryTreeItem, IDirectoryDeleteContext } from "./tree/fileShare/DirectoryTreeItem";
 import { FileShareTreeItem, IFileShareCreateChildContext } from "./tree/fileShare/FileShareTreeItem";
 import { FileTreeItem } from "./tree/fileShare/FileTreeItem";
+import { getAppResourceIdFromId } from "./utils/azureUtils";
 import { createBlobClient, createBlockBlobClient, createOrUpdateBlockBlob, doesBlobExist, IBlobContainerCreateChildContext } from './utils/blobUtils';
 import { createFileClient, doesFileExist, updateFileFromText } from "./utils/fileUtils";
 import { localize } from "./utils/localize";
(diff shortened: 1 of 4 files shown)


alexweininger (commented):
Empty file?
image

src/AzureStorageFS.ts
```diff
@@ -440,6 +441,18 @@ export class AzureStorageFS implements vscode.FileSystemProvider, vscode.TextDoc
     private async lookupRoot(uri: vscode.Uri, context: IActionContext, resourceId: string): Promise<FileShareTreeItem | BlobContainerTreeItem> {
         const rootName: string = path.basename(resourceId);
         const loadingMessage: string = this.isFileShareUri(uri) ? localize('loadingFileShare', 'Loading file share "{0}"...', rootName) : localize('loadingContainer', 'Loading blob container "{0}"...', rootName);
+
+        try {
+            // the storage account needs to be resolved for the file system to read the files
+            const appResourceId = getAppResourceIdFromId(resourceId);
+            if (appResourceId) {
+                const appResource = await ext.rgApi.appResourceTree.findTreeItem(appResourceId, { ...context, loadAll: true }) as unknown as { resolve: () => Promise<void> }
```
alexweininger, line 449: We should think about exposing an API from RGs that can do this more elegantly (after build).

alexweininger (approved):

src/AzureStorageFS.ts
nturinski: Yeah, that is what I was originally working on but it would require changes in the API and RG so I was like...naaah.

alexweininger (approved):

src/AzureStorageFS.ts
```diff
@@ -440,6 +441,18 @@ export class AzureStorageFS implements vscode.FileSystemProvider, vscode.TextDoc
...
+                await appResource.resolve();
```
bwateratmsft, line 450: This should be in the RG extension; clients shouldn't be calling `resolve`. I think we need `findTreeItem` to resolve any resource nodes it hits along its search.

nturinski: Agreed, I definitely used the unknown hack just to get something working reliably before release.

I was planning to add an API command specifically for just `findTreeItem` to get around the fact that az-utils (where `findTreeItem` lives) doesn't have access to resolve, but it requires changes in 3 different places so I thought I'd wait until after //build.  Do you think it's worth doing this correctly before release?

bwateratmsft: I think it's fine to wait, but we should open an issue so it doesn't get missed.

nturinski:
Tracled by https://github.com/microsoft/vscode-azureresourcegroups/issues/276

PR #1076, opened 2022-06-22
Automatically log uploaded file URLs to output + quick copy functionality

### LINKED ISSUE #1027: Feat: Return blob URL after upload completes
As a user, when I upload a file to my blob storage in VSC, I want to know the URL associated to the file location so I can quickly retrieve it and add it to my HTML file I'm working with or share the link with someone.

Some suggestions are:
- including the URL in the `output` log
- adding a `copy URL` action to the notification popup

image

I am aware of `copy URL` in the left nav menu but I would prefer an option within the context of the upload action
image

MicroFish91:
Closes #1027 

In this example, we "upload files" to the blob container which I have named "lob" 😆, we see the output log shows the file URLs now.  We get an information message pop-up that allows you to copy the most recent file URL to the clipboard.

image

If you click the "Copy to Clipboard" button, the following will happen as the next step:

image

Implementation - I made two main changes/additions:

1.  I split up the copyUrl tree command in BlobContainerTreeItem & FileShareTreeItem into two components => copyUrl & getUrl

I did this because I needed a way to extract the URL without automatically copying to clipboard

2.  I created a new util upload function to handle the core functionality of the issue - outputAndCopyUploadedFileUrls() => main logic found there

### DIFF
--- a/src/commands/uploadFiles.ts
+++ b/src/commands/uploadFiles.ts
@@ -81,7 +81,7 @@ export async function uploadFiles(
         });
 
         if (!calledFromUploadToAzureStorage) {
-            void outputAndCopyUploadedFileUrls(treeItem.getUrl(), fileEndings);
+            await outputAndCopyUploadedFileUrls(treeItem.getUrl(), fileEndings);
         }
 
         return resolution;
--- a/src/commands/uploadFiles.ts
+++ b/src/commands/uploadFiles.ts
@@ -79,7 +79,7 @@ export async function uploadFiles(
         });
 
         if (!calledFromUploadToAzureStorage) {
-            showUploadSuccessMessage(treeItem.label);
+            showUploadSuccessMessage(treeItem);
         }
 
         return resolution;
--- a/src/commands/uploadFiles.ts
+++ b/src/commands/uploadFiles.ts
@@ -12,7 +12,7 @@ import { FileShareTreeItem } from "../tree/fileShare/FileShareTreeItem";
 import { AzExtFsExtra } from "../utils/AzExtFsExtra";
 import { isAzCopyError, multipleAzCopyErrorsMessage, throwIfCanceled } from "../utils/errorUtils";
 import { nonNullValue } from "../utils/nonNull";
-import { checkCanUpload, convertLocalPathToRemotePath, getDestinationDirectory, getUploadingMessage, outputUploadedFileUrls, OverwriteChoice, showUploadSuccessMessage, upload } from "../utils/uploadUtils";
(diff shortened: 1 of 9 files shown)


src/utils/uploadUtils.ts
```diff
@@ -54,6 +54,22 @@ export function getUploadingMessageWithSource(sourcePath: string, treeItemLabel:
...
+    ext.outputChannel.show();
+
+    for (const fileUrl of fileUrls) {
+        const url: string = localize('uploadedFileUrl', 'Uploaded file URL: {0}', `${parentUrl}/${fileUrl}`);
+        ext.outputChannel.appendLog(url);
+    }
+
+    const shouldCopy: boolean = !!(await vscode.window.showInformationMessage(`Finished uploading ${fileUrls.length} file(s).`, "Copy to Clipboard"));
+    if (shouldCopy) {
+        const lastFileUrl: string = `${parentUrl}/${fileUrls[fileUrls.length - 1]}`;
```
alexweininger, line 67: It'd be cool if there was a way to link to all the files that were uploaded, opposed to just the last uploaded file. But I don't think that's possible.

MicroFish91: Yeah I was thinking about that too - I suppose if people are looking through their clipboard history it could be useful to have all the links added instead of just the last one, but I'm not sure if that's any faster than just copying directly from the output log... what do you think?

src/commands/uploadFiles.ts
```diff
@@ -79,7 +81,7 @@ export async function uploadFiles(
         });
 
         if (!calledFromUploadToAzureStorage) {
-            showUploadSuccessMessage(treeItem.label);
+            void outputAndCopyUploadedFileUrls(treeItem.getUrl(), fileEndings);
```
alexweininger, line 84: Is there a reason you didn't `await` this function call?

MicroFish91: I just figured at this point that it wasn't crucial to wait for the function call to complete before moving onwards. Do you think an await be better to use here instead?

alexweininger: I think it's best to always await a promise unless there's an explicit reason. Un-awaited promises can cause unexpected behavior. For example, if an error is thrown inside an un-awaited promise, it won't get handled properly.

Plus, in this case the async call is after the files have been uploaded, so awaiting here won't slow down or delay the upload action.

MicroFish91: Thanks for that explanation - that makes a lot of sense.  I'll keep that in mind for the future :)

alexweininger (approved):

jeremyfiel:
Thanks a lot @MicroFish91 👍

PR #1084, opened 2022-07-06
Add Upload Files Support to the Activity Log

MicroFish91:
Added upload files support to the Activity Log:

image

This involved refactoring the `uploadFiles.ts` logic so that it is now using an `AzureWizard`.  The refactored logic has been consolidated in the new `uploadFiles/` directory:

image

Let me know what you guys think!


src/commands/uploadFiles/IExistingFileContext.ts
```diff
@@ -0,0 +1,6 @@
+import { IActionContext } from "@microsoft/vscode-azext-utils";
```
bwateratmsft, line 1: This file is missing the copyright header

src/commands/uploadFiles/UploadFilesStep.ts
```diff
@@ -0,0 +1,143 @@
...
+let lastUriUpload: Uri | undefined;
+
+export class UploadFilesStep extends AzureWizardExecuteStep<IUploadFilesWizardContext> {
+    public priority: number = 100;
+    public treeItem?: BlobContainerTreeItem | FileShareTreeItem;
+    public uris?: Uri[];
+    public notificationProgress?: NotificationProgress;
+    public cancellationToken?: CancellationToken;
+
+    public constructor(treeItem?: BlobContainerTreeItem | FileShareTreeItem, uris?: Uri[], notificationProgress?: NotificationProgress, cancellationToken?: CancellationToken) {
```
bwateratmsft, line 29: What you have is valid, but you can do this to save some lines of code. Removes the need for the declaration above and the setter in the body of the constructor.

```typescript
public constructor(
    public treeItem?: BlobContainerTreeItem | FileShareTreeItem,
    public uris?: Uri[],
    public notificationProgress?: NotificationProgress,
    public cancellationToken?: CancellationToken
) {
    super();
}
```

If the properties do not need to be changed after initialization, you can do `public readonly` instead of `public`; if they don't need to be read outside of the class I'd do `private` / `private readonly`.

src/commands/uploadFiles/UploadFilesStep.ts
```diff
@@ -0,0 +1,143 @@
...
+        super();
+        this.treeItem = treeItem;
+        this.uris = uris;
+        this.notificationProgress = notificationProgress;
+        this.cancellationToken = cancellationToken;
+    }
+
+    public async execute(context: IUploadFilesWizardContext, _progress: NotificationProgress): Promise<void> {
+        const calledFromUploadToAzureStorage: boolean = this.uris !== undefined;
```
bwateratmsft, line 38: Is `[]` (empty array) semantically equivalent to `undefined` in this context? If so, I'd suggest:
```suggestion
        const calledFromUploadToAzureStorage: boolean = !!this.uris?.length;
```

src/commands/uploadFiles/UploadFilesStep.ts
```diff
@@ -0,0 +1,140 @@
...
+                    canSelectMany: true,
+                    defaultUri: lastUriUpload,
+                    filters: {
+                        "All files": ['*']
+                    },
+                    openLabel: upload
+                }
+            );
+        }
+        context.activityTitle = localize('activityLogUploadingFiles', `Uploading ${this.uris?.length ?? ''} file(s) to ${this.treeItem?.label ? `"${this.treeItem.label}"` : 'storage'}`);
```
alexweininger, line 50: A few minor things on the wording of the activity title:

1. We already know how many files are being uploaded, so I'd rather put `file` or `files` instead of always putting `file(s)`.
1. The pattern so far for activity titles has been to avoid using gerunds aka words ending in -ing (I had to look that up btw). So it should be "Upload" instead of "Uploading".
1. For a single file, it'd be useful to include the name of the file being uploaded. Ex: `Upload "filename.txt" to...`.

alexweininger, line 50: If you move this line of code to after line 54, then `this.treeItem` will always be defined, so we can change ```this.treeItem?.label ? `"${this.treeItem.label}"` : 'storage'``` to just `this.treeItem.label`.

src/commands/uploadFiles/uploadFiles.ts
```diff
@@ -0,0 +1,39 @@
...
+import { BlobContainerTreeItem } from "../../tree/blob/BlobContainerTreeItem";
+import { FileShareTreeItem } from "../../tree/fileShare/FileShareTreeItem";
+import { createActivityContext } from "../../utils/activityUtils";
+import { localize } from "../../utils/localize";
+import { IAzCopyResolution } from "../azCopy/IAzCopyResolution";
+import { GetDestinationDirectoryStep } from './GetDestinationDirectoryStep';
+import { IUploadFilesWizardContext } from "./IUploadFilesWizardContext";
+import { UploadFilesStep } from "./UploadFilesStep";
+
+export async function uploadFiles(
```
alexweininger, line 18: If I do "Upload Files..." then cancel out of the file picker, it shows the activity as failed. The activity shouldn't show up until after this prompt.

image

alexweininger (changes requested):

src/commands/uploadFiles/GetDestinationDirectoryStep.ts
```diff
@@ -0,0 +1,40 @@
...
+                }
+            );
+        }
+        context.treeItem = context.treeItem || await ext.rgApi.pickAppResource<BlobContainerTreeItem | FileShareTreeItem>(context, {
+            filter: storageFilter,
+            expectedChildContextValue: [BlobContainerTreeItem.contextValue, FileShareTreeItem.contextValue]
+        });
+    }
+
+    public shouldPrompt(_context: IUploadFilesWizardContext): boolean {
```
alexweininger, line 37:
```suggestion
    public shouldPrompt(): boolean {
```

src/commands/uploadFiles/UploadFilesStep.ts
```diff
@@ -0,0 +1,112 @@
...
+
+            if (!context.calledFromUploadToAzureStorage) {
+                outputAndCopyUploadedFileUrls(context.treeItem.getUrl(), fileEndings);
+            }
+
+            context.resolution = resolution;
+        }
+    }
+
+    public shouldExecute(_context: IUploadFilesWizardContext): boolean {
```
alexweininger, line 64:
```suggestion
    public shouldExecute(): boolean {
```

alexweininger (dismissed):
This PR looks good to merge. However, there are a handful of improvements I'd like to make to Activity Logs

1. In order to fully comply with the `azureResourceGroups.suppressActivityNotifications` setting, we shouldn't be showing this notification when it's set to `true`. 

image

This change should be made later when Activities have implemented cancel support.

2. An easy one: there should be a right click -> copy file url on the successful upload activity. I think we can and should do this soon.
3. A more advanced future improvement to the above point: every uploaded file should become a child of the successful activity. That way users can copy the url of any uploaded file.
4. Refactor upload folder in the same manner so that it hooks into the activity log
5. Explore hooking up other actions like download file to the activity log

PR #1087, opened 2022-07-13
Add Upload Folder Support to the Activity Log

MicroFish91:
Refactored the uploadFolder.ts logic into the uploadFolder/ directory where it is now being run through an AzureWizard.

image

**Sidenote**: I was thinking of going back to uploadFiles/ in a follow-up PR to clean some stuff up, for example, to change the GetDestinationDirectoryStep to something like GetFileDestinationDirectoryStep.... to match the same naming convention as the newer GetFolderDestinationDirectoryStep


bwateratmsft (approved):

alexweininger (approved):
> uploadFiles/ in a follow-up PR to clean some stuff up

Good idea for sure. If we can share more stuff between these steps that would be good.

PR #1095, opened 2022-07-16
Small fixes/improvements to earlier uploadFiles activityLog refactor

MicroFish91:


src/utils/uploadUtils.ts
```diff
@@ -61,7 +61,7 @@ export function outputAndCopyUploadedFileUrls(parentUrl: string, fileUrls: strin
     }
 
     void vscode.window.showInformationMessage(
-        `Finished uploading ${fileUrls.length} file(s).`,
+        `Finished uploading ${fileUrls.length === 1 ? 'file' : `${fileUrls.length} files`}.`,
```
bwateratmsft, line 64: This string should be localized, and ideally use `context.ui.showInformationMessage` from the `IActionContext` item, if at all possible.

bwateratmsft, line 64: Personally I would prefer it to say "Finished uploading 1 file" instead of "Finished uploading file", but that's of course subjective.

MicroFish91: Thank you for catching the localize!  In regards to `context.ui.showInformationMessage`, I'm not picking that method up on IActionContext / the azureTools package.  Do you think we should have that added to `AzExtUserInput` or just go with the `vscode.window.showInformationMessage`?

src/utils/uploadUtils.ts
```diff
@@ -61,8 +61,8 @@ export function outputAndCopyUploadedFileUrls(parentUrl: string, fileUrls: strin
     }
 
     void vscode.window.showInformationMessage(
-        `Finished uploading ${fileUrls.length === 1 ? 'file' : `${fileUrls.length} files`}.`,
-        "Copy to Clipboard"
+        localize('outputAndCopyFinished', `Finished uploading ${fileUrls.length} ${fileUrls.length === 1 ? 'file' : 'files'}.`),
```
bwateratmsft, line 64: Localize does formatting differently:
```suggestion
        localize('outputAndCopyFinished', 'Finished uploading {0} {1}.', fileUrls.length, fileUrls.length === 1 ? 'file' : 'files'),
```

src/utils/uploadUtils.ts
bwateratmsft: If you pass the `IActionContext` down from the methods calling this, you'll have access to `context.ui`.

src/utils/uploadUtils.ts
MicroFish91: Oof, thank you for pointing this out to me.  I think I did it right once or twice before and then proceeded to not use it correctly every other time after... 😧

I did notice that there were other places that I had not touched that were not always using it where they (probably) should be... I might look into fixing those in the near future.

src/utils/uploadUtils.ts
nturinski: Yeah, we never added `showInformationMessage` to the `AzExtUserInput`. We never felt an urge strong enough to implement it because typically, it's just a window that shows a message-- not something that we have any actions/follow up with.

src/utils/uploadUtils.ts
bwateratmsft: Well, technically the interpolated string will work, but, it would not really permit for real localization, if we ever actually do that.

bwateratmsft (approved):

Issue #1093, opened 2022-07-15
 Pop up an error after deleting a folder for a blob container


v-chenzha:
**OS:** Win 10
**Build:** 20220714.11
**Regression:** Not a regression

**Repro Steps:**

1. Expand a storage account -> Create a blob container 
2. Right click the blob container -> Click "Create Empty Blob..." -> Type name "a/b.txt".
3. Right click the blob container again -> Click "Create Empty Blob..." -> Type name "a/c/d.txt".
4. Right click "a" folder ->Click "Delete..." 
5. Check whether no error pops up.

**Expect:**
No error pops up.

**Actual:**
Pop up an error 'Failed with code "404"'.
![image](https://user-images.githubusercontent.com/107091828/179184313-908d130d-6145-4b90-8c3b-5c384f7189c2.png)

alexweininger:
This might be related to or a duplicate of #1073

Issue #1073, opened 2022-06-21
An error occurs when deleting a file for the blob container or file share


v-xinda:
**OS:** Win10
**Build Version:**  20220619.1
**Regression:** Previous release

**Repro Steps:**
1. Expand a storage account -> Create a blob container -> Create an empty blob.
2. Delete the above empty blob.
3. Check whether no error occurs.

**Expect:**
No error occurs.

**Actual:**
An error occurs.
![image](https://user-images.githubusercontent.com/57662131/174706740-e505c326-68d3-4415-9099-46f3ea823951.png)

**More Info:**
1. This issue also reproduces for the file share.
2. This issue also reproduces for the attached storage account and emulator account.

alexweininger:
Does this issue still reproduce?

v-xinda:
@alexweininger   Verified this issue with the latest build 20220630.3, it still reproduces now.

PR #1102, opened 2022-08-09
Throw FileSystemError on readFile and stat

nturinski:
Fixes https://github.com/microsoft/vscode-azurestorage/issues/1073
Fixes https://github.com/microsoft/vscode-azurestorage/issues/1093

When we changed to T2, I'm pretty sure that the errorTypes changed. Whenever a resource is not found, it throws a "404" rather than the other types that were previously there.

Also, stat would sometimes be called and would call `await fileClient.getProperties();` before we would actually delete the treeItem. 

I'm guessing that this started happening because when we moved to the RG view, rather checking the cache of Storage which is clearewd instantly, we call `await ext.rgApi.appResourceTree.findTreeItem(uriPath, { ...context, loadAll: true });`. I'm guessing RG wasn't clearing the child out fast enough before `stat` was called again.  This made it so treeItem was defined, we call `getProperties()`, ping Azure, only to throw a 404 because the file was actually deleted.

### DIFF
--- a/src/AzureStorageFS.ts
+++ b/src/AzureStorageFS.ts
@@ -107,6 +107,8 @@ export class AzureStorageFS implements vscode.FileSystemProvider, vscode.TextDoc
 
         return await callWithTelemetryAndErrorHandling('stat', async (context) => {
             context.telemetry.suppressIfSuccessful = true;
+            context.errorHandling.rethrow = true;
+            context.errorHandling.suppressDisplay = true;
 
             if (uri.path.endsWith('/')) {
                 // Ignore trailing forward slashes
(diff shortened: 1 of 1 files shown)


src/AzureStorageFS.ts
```diff
@@ -228,7 +230,7 @@ export class AzureStorageFS implements vscode.FileSystemProvider, vscode.TextDoc
                 result = await this.streamToString(downloaded.readableStreamBody);
             } catch (error) {
                 const pe = parseError(error);
-                if (pe.errorType === 'BlobNotFound' || pe.errorType === 'ResourceNotFound') {
+                if (pe.errorType === '404') {
```
alexweininger, line 233: Do we need to add a comment anywhere explaining what this is doing?

alexweininger (approved):

alexweininger:
Also, https://github.com/microsoft/vscode-azurestorage/issues/1093 seems like the same issue. Does this PR fix that as well?

src/AzureStorageFS.ts
nturinski: I mean, I feel like a 404 error is probably clear enough that we're just ignoring the error if the file wasn't found.

Also I realized that stat was being called constantly because we open the empty blob in the editor (also the reason we have the create/delete events).

alexweininger: Ok, this PR and linked issue has enough context if we need to understand this later on. LGTM

nturinski:
Yep, just tried it and it seems to fix https://github.com/microsoft/vscode-azurestorage/issues/1093 as well.  Good thinking! 💡

PR #1108, opened 2022-08-11
Remove leading slash from user input when creating a blob

### LINKED ISSUE #970: The input textbox of the name pops up again when creating a blob with the name "/a/b/c/test.txt"
**OS:** Windows
**Build Version:** 20210426.1

**Repro Steps:**
1. Create a new storage account -> Create a new blob container for it.
2. Right Click the blob container -> Create Empty Blob -> Type blob name `/a/b/c/test.txt` -> Enter.
3. Check Whether the input textbox of the name pops up again.

**Expect:**
The input textbox of the name doesn't pop up again.

**Actual:**
The input textbox of the name pops up again after typing the name `/a/b/c/test.txt`.

**More Info:**
1. Pop up an error "c.createChild is not a function" after creating the blob.
2. Only display the `test.txt` file under the "blob container" node after creating the blob, the folder displays after refreshing.
3. The `test.txt` file can't be opened. 
abc
4. This issue doesn't reproduce when typing a name that start with a number or letter.

alexweininger:
Fixes #970 

These changes do two things:

#### 1. Support creating blobs that have unnamed directories in their paths.

Yes this is actually allowed by Azure and is even displayed on the Portal.

The result of creating `/a/b/c/test.txt`:
image

image

#### 2. Strip any leading slash from user input so that it doesn't create an unnamed directories because that behavior is so weird.

I wanted to still add support for creating unnamed directories in case we allow this down the road for some reason.

### DIFF
--- a/src/utils/blobUtils.ts
+++ b/src/utils/blobUtils.ts
@@ -61,7 +61,7 @@ export async function loadMoreBlobChildren(parent: BlobContainerTreeItem | BlobD
 
 // Currently only supports creating block blobs
 export async function createChildAsNewBlockBlob(parent: BlobContainerTreeItem | BlobDirectoryTreeItem, context: ICreateChildImplContext & IBlobContainerCreateChildContext): Promise<BlobTreeItem> {
-    const blobPath: string = context.childName || await getBlobPath(context, parent);
+    const blobPath: string = context.childName ?? await getBlobPath(context, parent);
 
     return await vscode.window.withProgress({ location: vscode.ProgressLocation.Window }, async (progress) => {
         context.showCreatingTreeItem(blobPath);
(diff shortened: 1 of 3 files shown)


src/utils/blobUtils.ts
```diff
@@ -61,7 +61,7 @@ export async function loadMoreBlobChildren(parent: BlobContainerTreeItem | BlobD
 
 // Currently only supports creating block blobs
 export async function createChildAsNewBlockBlob(parent: BlobContainerTreeItem | BlobDirectoryTreeItem, context: ICreateChildImplContext & IBlobContainerCreateChildContext): Promise<BlobTreeItem> {
-    const blobPath: string = context.childName || await getBlobPath(context, parent);
+    const blobPath: string = context.childName ?? await getBlobPath(context, parent);
```
alexweininger, line 64: Using `||` here is problematic because `''` is evaluated as `false`. Using the [nullish coalescing `??` operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator) instead makes sure the left side is actually `undefined`.

bwateratmsft (approved):

bwateratmsft:
> Yes this is actually allowed by Azure

🤮

alexweininger:
> > Yes this is actually allowed by Azure
> 
> 🤮

🤮 indeed. I was so surprised when it allowed me to create and it actually displayed in the portal.

alexweininger:
We could have caught this with an eslint rule... but this rule is really annoying. 

```
rules: {
        "@typescript-eslint/strict-boolean-expressions": [
            2,
            {
                allowString: false
            }
        ]
    }
```

PR #1109, opened 2022-08-11
Disable 'Open in Storage Explorer' default behavior when in a remote workspace

### LINKED ISSUE #1023: [Codespaces][Suggestion] Don't show "Open in Storage Explorer" menu item on local Codespaces
**Environment:** **Local Codespaces**
**Build Version:** 20220128.1
**Regression:** Not a regression

**Repro Steps:**
1. Create a new storage account.
2. Right click the storage account and check its context menu.

**Actual:**
The menu item "Open in Storage Explorer" shows in the context menu.
image

**Suggestion:**
Don't show "Open in Storage Explorer" menu item on local Codespaces.

**More Info:**
There is no "Open in Storage Explorer" menu item on Linux or remote Codespaces.

MicroFish91:
Fixes #1023 

Added a check when running `Open in Storage Explorer` that checks `remoteName`.  If populated, we warn that the feature is not supported and return..

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -337,7 +337,7 @@
                 {
                     "$comment": "========= azureStorageAccount =========",
                     "command": "azureStorage.openStorageAccount",
-                    "when": "view == azureResourceGroups && viewItem =~ /azureStorageAccount/ && !isWeb && !isLinux",
+                    "when": "view == azureResourceGroups && viewItem =~ /azureStorageAccount/ && !isWeb && !isLinux && !remoteName",
                     "group": "navigation@1"
                 },
                 {
@@ -711,6 +711,10 @@
                 {
                     "command": "azureStorage.uploadToAzureStorage",
                     "when": "never"
+                },
+                {
+                    "command": "azureStorage.openStorageAccount",
+                    "when": "!remoteName"
                 }
             ]
         },
--- a/src/commands/storageAccountActionHandlers.ts
+++ b/src/commands/storageAccountActionHandlers.ts
@@ -27,12 +27,6 @@ export function registerStorageAccountActionHandlers(): void {
 }
 
 async function openStorageAccountInStorageExplorer(context: IActionContext, treeItem?: ResolvedAppResourceTreeItem<ResolvedStorageAccount>): Promise<void> {
-    if (vscode.env.remoteName) {
-        const noRemoteSupport: string = localize('noRemoteSupport', 'This feature is not supported on remote workspaces.');
-        void context.ui.showWarningMessage(noRemoteSupport);
-        return;
-    }
-
     if (!treeItem) {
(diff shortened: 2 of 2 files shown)


src/commands/storageAccountActionHandlers.ts
```diff
@@ -27,6 +27,12 @@ export function registerStorageAccountActionHandlers(): void {
 }
 
 async function openStorageAccountInStorageExplorer(context: IActionContext, treeItem?: ResolvedAppResourceTreeItem<ResolvedStorageAccount>): Promise<void> {
+    if (vscode.env.remoteName) {
```
bwateratmsft, line 30: A `when` clause context in `package.json` would be a more appropriate place to trap this. i.e. in `contributes.menus.commandPalette`:

```json
{
    "command": "azureStorage.openStorageAccount",
    "when": "!remoteName"
},
```

bwateratmsft: And a similar edit should be made in `contributes.menus.view/item/context` for this command.

package.json
```diff
@@ -337,7 +337,7 @@
                 {
                     "$comment": "========= azureStorageAccount =========",
                     "command": "azureStorage.openStorageAccount",
-                    "when": "view == azureResourceGroups && viewItem =~ /azureStorageAccount/ && !isWeb && !isLinux",
+                    "when": "view == azureResourceGroups && viewItem =~ /azureStorageAccount/ && !isWeb && !isLinux && !remoteName",
```
bwateratmsft, line 340: Double check that this works; I don't know for sure. It might want `&& remoteName == ""` or `!remoteName || remoteName == ""`, etc.

bwateratmsft (approved):

package.json
alexweininger: I don't think `remoteName` is an [available context](https://code.visualstudio.com/api/references/when-clause-contexts#available-contexts). If there isn't a suitable one available, it might be best to use `setContext` to create our own.

bwateratmsft: It's not documented, but I found it was there, set by the remote extensions. It sorta makes sense for it to not be documented there since the remote extensions are not "officially" part of VSCode

alexweininger: > It's not documented, but I found it was there, set by the remote extensions. It sorta makes sense for it to not be documented there since the remote extensions are not "officially" part of VSCode

Oh, nice!

MicroFish91: Yup I double checked and it worked when I ran it on Codespaces!

alexweininger (approved):

package.json
bwateratmsft: I used the command "Developer: Inspect Context Keys" for that, by the way. It logs all of the context keys to the Developer Tools window ("Developer: Toggle developer tools")

PR #1113, opened 2022-08-12
Add delete many support for blob and fileshare (files/directories)

### LINKED ISSUE #951: Deleting all selected blobs
It would be time-saving if one could delete all selected blobs in one go! Right now it is possible to select several blobs, but right clicking and choosing "Delete" only deletes one.

This is a feature request.

### LINKED ISSUE #1114: Feat: Delete many support for File Shares
Similar piggyback off of #951

MicroFish91:
Closes #951 
Closes #1114 

Added support and consolidated the number of `deleteBlob` commands.  Will look into doing something similar for FileShare in the near future... (likely could consolidate most of the delete code into one command but would need to investigate further)

There's two edge cases that I'm covering in this iteration:

<b>Case 1</b>:  If you select multiple items but then right click and delete a different (non-selected) node... 

I decided to model the behavior off of how deleting worked in the normal Windows FS GUI (so basically, only delete the right clicked node when it's not part of the normal selection)

image

<b>Case 2</b>:  If your selection includes files that are nested in directories of other selections...

image

To avoid `not found` errors, I check to see if each node is a directory.  If it is, I go through the remaining nodes and check to see if they are subPaths.  If they are, I remove them from the remaining list.  The implementation uses a nested for loop which I wasn't thrilled about, but I didn't have an obvious alternative that I liked and felt that the average user's selection size probably wasn't anything that would scale out of control.  If you have a better suggestion though, please let me know.  Thanks!

### DIFF
--- a/src/commands/blob/blobActionHandlers.ts
+++ b/src/commands/blob/blobActionHandlers.ts
@@ -23,18 +23,18 @@ export async function deleteBlob(context: IActionContext, treeItem: BlobTreeItem
     }
 
     const dirPaths: string[] = [];
-    let shouldContinue;
+    let shouldSkip;
     for (const node of selection) {
-        shouldContinue = false;
+        shouldSkip = false;
 
-        // Check to see if it's a child we've already deleted
+        // Check to see if it's a resource we've already deleted
         for (const dirPath of dirPaths) {
             if (isSubpath(dirPath, node.fullId)) {
-                shouldContinue = true;
+                shouldSkip = true;
                 break;
             }
         }
-        if (shouldContinue) continue;
+        if (shouldSkip) continue;
         if (isTreeItemDirectory(node)) {
             dirPaths.push(node.fullId);
         }
--- a/src/commands/blob/blobActionHandlers.ts
+++ b/src/commands/blob/blobActionHandlers.ts
@@ -3,9 +3,6 @@
  *  Licensed under the MIT License. See License.md in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import { registerCommand } from "@microsoft/vscode-azext-utils";
-import { deleteFilesAndDirectories } from "../deleteFilesAndDirectories";
-
 export function registerBlobActionHandlers(): void {
-    registerCommand("azureStorage.deleteBlob", deleteFilesAndDirectories);
+    // Placeholder
 }
--- a/src/commands/blob/blobActionHandlers.ts
+++ b/src/commands/blob/blobActionHandlers.ts
@@ -14,9 +14,7 @@ export function registerBlobActionHandlers(): void {
     registerCommand("azureStorage.deleteBlob", deleteBlob);
 }
 
-export async function deleteBlob(context: IActionContext, treeItem: BlobTreeItem | BlobDirectoryTreeItem): Promise<void> {
-    const { selection } = ext.rgApi.appResourceTreeView;
-
+export async function deleteBlob(context: IActionContext, treeItem: BlobTreeItem | BlobDirectoryTreeItem, selection: (BlobTreeItem | BlobDirectoryTreeItem)[] = []): Promise<void> {
     // Covers both single selection and an edge case where the node you delete from isn't part of the selection
     if (!selection.some(s => s === treeItem)) {
         await ext.rgApi.appResourceTreeView.reveal(treeItem);
--- a/src/commands/blob/blobActionHandlers.ts
+++ b/src/commands/blob/blobActionHandlers.ts
@@ -22,11 +22,12 @@ export async function deleteBlob(context: IActionContext, treeItem: BlobTreeItem
         return;
     }
 
-    // When deleting a directory, go through the remaining selections and check to see if it's a child we've already deleted
     const dirPaths: string[] = [];
     let shouldContinue;
     for (const node of selection) {
         shouldContinue = false;
+
+        // Check to see if it's a child we've already deleted
         for (const dirPath of dirPaths) {
             if (isSubpath(dirPath, node.fullId)) {
                 shouldContinue = true;
--- a/src/commands/blob/blobActionHandlers.ts
+++ b/src/commands/blob/blobActionHandlers.ts
@@ -22,16 +22,20 @@ export async function deleteBlob(context: IActionContext, treeItem: BlobTreeItem
         return;
     }
 
-    for (const [n, node] of selection.entries()) {
-        // When deleting a directory, go through the remaining selections and check to see if it's a child we've already deleted
-        if (isTreeItemDirectory(node)) {
-            for (let i = n + 1; i < selection.length; i++) {
-                const unverifiedNode = selection[i];
-                if (isSubpath(node.fullId, unverifiedNode.fullId)) {
-                    selection.splice(i, 1);
-                    i--;
-                }
+    // When deleting a directory, go through the remaining selections and check to see if it's a child we've already deleted
+    const dirPaths: string[] = [];
+    let shouldContinue;
+    for (const node of selection) {
+        shouldContinue = false;
+        for (const dirPath of dirPaths) {
+            if (isSubpath(dirPath, node.fullId)) {
+                shouldContinue = true;
+                break;
             }
+        }
+        if (shouldContinue) continue;
+        if (isTreeItemDirectory(node)) {
+            dirPaths.push(node.fullId);
         }
         await node.deleteTreeItem(context);
     }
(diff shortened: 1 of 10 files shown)


nturinski:
> To avoid not found errors, I check to see if each node is a directory. If it is, I go through the remaining nodes and check to see if they are subPaths. If they are, I remove them from the remaining list. The implementation uses a nested for loop which I wasn't thrilled about, but I didn't have an obvious alternative that I liked and felt that the average user's selection size probably wasn't anything that would scale out of control. If you have a better suggestion though, please let me know. Thanks!

What if you keep a list of the id/dir path when you delete them? Then check if the currently deleting file is a subpath of any of deleted directories?

alexweininger:
> What if you keep a list of the id/dir path when you delete them? Then check if the currently deleting file is a subpath of any of deleted directories?

Yeah I think this is the best approach. Or you could attempt to delete everything and swallow the `not found` errors 😎

src/commands/blob/blobActionHandlers.ts
```diff
@@ -4,14 +4,37 @@
...
+import { isSubpath } from "../../utils/fs";
 
 export function registerBlobActionHandlers(): void {
     registerCommand("azureStorage.deleteBlob", deleteBlob);
-    registerCommand("azureStorage.deleteBlobDirectory", async (context: IActionContext, treeItem: BlobDirectoryTreeItem) => await treeItem.deleteTreeItem(context));
 }
 
-export async function deleteBlob(context: IActionContext, treeItem: BlobTreeItem): Promise<void> {
-    await treeItem.deleteTreeItem(context);
+export async function deleteBlob(context: IActionContext, treeItem: BlobTreeItem | BlobDirectoryTreeItem): Promise<void> {
```
bwateratmsft, line 17: The standard way to handle multi-select commands passes the current selection as the second argument (in our case, third, since we inject an `IActionContext`). So, it should look like this:
```typescript
export async function deleteBlob(context: IActionContext, _treeItem?: BlobTreeItem | BlobDirectoryTreeItem, treeItems?: BlobTreeItem[] | BlobDirectoryTreeItem[]): Promise<void> {
```

With this approach, you don't need to pull the `selection` from `ext.rgApi.appResourceTreeView`. You can also ignore `_treeItem` since it will be included in the array.

bwateratmsft: I don't know if this behavior is actually documented anywhere, but either way I wish it was more clear.

MicroFish91:
> > What if you keep a list of the id/dir path when you delete them? Then check if the currently deleting file is a subpath of any of deleted directories?
> 
> Yeah I think this is the best approach. Or you could attempt to delete everything and swallow the `not found` errors 😎

I actually had an implementation like this but ended up changing it because I thought people might not be happy if I handled it this way.  Is this an okay way to handle this??

nturinski:
> > Or you could attempt to delete everything and swallow the `not found` errors 😎
> 
> I actually had an implementation like this but ended up changing it because I thought people might not be happy if I handled it this way 😂 Is this an okay way to handle this?? (it certainly looks a lot cleaner 😎)

lol I'm personally against this as I don't really like making a bunch of calls to Azure just to have them fail especially if we could pretty easily prevent that, but if everyone else is okay with it, then I would concede my position.

alexweininger:
Yeah I think we can avoid the `not found` errors easily enough using Nathan's suggestion.

src/commands/blob/blobActionHandlers.ts
MicroFish91: Oh neat, I think I'd still have to keep the second argument as `treeItem` because of the reveal behavior I'm implementing (unless that seems like undesirable behavior), but this definitely cleans up that first line and is super useful for future use :)

MicroFish91:
> What if you keep a list of the id/dir path when you delete them? Then check if the currently deleting file is a subpath of any of deleted directories?

Yeah I'm just now thinking more about it, and my solution is suboptimal because of the splice method right?

I just thought of a way to do it without splice using a set for constant time access which would actually make it the fastest solution in my opinion, but it seems like maybe excessive/an overengineer for a small add like this, so probably not a big deal 😂

src/commands/blob/blobActionHandlers.ts
```diff
@@ -14,27 +14,29 @@ export function registerBlobActionHandlers(): void {
...
-        if (isTreeItemDirectory(node)) {
-            for (let i = n + 1; i < selection.length; i++) {
-                const unverifiedNode = selection[i];
-                if (isSubpath(node.fullId, unverifiedNode.fullId)) {
-                    selection.splice(i, 1);
-                    i--;
-                }
+    // When deleting a directory, go through the remaining selections and check to see if it's a child we've already deleted
+    const dirPaths: string[] = [];
+    let shouldContinue;
```
nturinski, line 26: shouldContinue is the same thing as skipping, right? The naming convention is a little ambiguous to me since continue can mean

1. We should continue this operation and delete it
2. We should skip this and continue to the next one

I'd recommend changing it to `shouldSkip` or `directoryDeleted` or something along those lines.

src/commands/blob/blobActionHandlers.ts
```diff
@@ -14,27 +14,29 @@ export function registerBlobActionHandlers(): void {
...
-    for (const [n, node] of selection.entries()) {
-        // When deleting a directory, go through the remaining selections and check to see if it's a child we've already deleted
-        if (isTreeItemDirectory(node)) {
-            for (let i = n + 1; i < selection.length; i++) {
-                const unverifiedNode = selection[i];
-                if (isSubpath(node.fullId, unverifiedNode.fullId)) {
-                    selection.splice(i, 1);
-                    i--;
-                }
+    // When deleting a directory, go through the remaining selections and check to see if it's a child we've already deleted
```
nturinski: nit: It's when deleting any resource, file or directory, right?

src/commands/blob/blobActionHandlers.ts
MicroFish91: Sure, in my mind it felt self explanatory since the command that gets run is `continue` in the if-statement directly below.

src/commands/blob/blobActionHandlers.ts
MicroFish91: Yeah I think you might be looking at an older commit, I changed it since then to: `// Check to see if it's a child we've already deleted`

MicroFish91:
So I looked into it and it was pretty easy to extend this functionality for File Shares as well, which is what my latest commit did.

I did notice that some of our `registerCommand` handler have a small bit of inconsistencies - for example when I originally consolidated `deleteBlob` and `deleteBlobDirectory` into one command, I did so because both were under the same register handler; however, in file shares they are separated out under two file and directory handlers, so it made a little less sense to do that there.

I'm happy to standardize in either direction, so let me know how you guys want to approach this moving forward.  Hope that makes some sort of sense... if not I can explain in a future meeting

Thanks :)

src/commands/blob/blobActionHandlers.ts
```diff
@@ -3,15 +3,6 @@
...
-import { BlobTreeItem } from "../../tree/blob/BlobTreeItem";
-
 export function registerBlobActionHandlers(): void {
-    registerCommand("azureStorage.deleteBlob", deleteBlob);
-    registerCommand("azureStorage.deleteBlobDirectory", async (context: IActionContext, treeItem: BlobDirectoryTreeItem) => await treeItem.deleteTreeItem(context));
-}
-
-export async function deleteBlob(context: IActionContext, treeItem: BlobTreeItem): Promise<void> {
-    await treeItem.deleteTreeItem(context);
+    // Placeholder
```
nturinski, line 7: Are you waiting on this? https://github.com/microsoft/vscode-azurestorage/pull/1115

nturinski (dismissed):

src/commands/blob/blobActionHandlers.ts
MicroFish91: I don't think that will fix it unfortunately.  

So the issue is that I consolidated all the `blob` and `fileshare` file/directory actions under basically two commands:  

image

image

 (these commands already existed, I just changed the function call reference to be `deleteFilesAndDirectories`

Since I consolidated all the old blob calls into just `deleteFile` and `deleteDirectory` that leaves this empty:

image

I kept it because I thought it was useful to have it here so that when people implemented future commands related to BlobActions, that they'd know where to put them.  If we want to leave that onus on the future user, I'm happy to get rid of this

nturinski: Probably should just get remove of it then. I think if we were to implement more futures, we'll probably have a different structure due to the new tree changes anyway.

alexweininger: Don't wait on #1115. I'm still trying to get it to work. 🙃

PR #1118, opened 2022-08-16
Declare limited support for virtual workspaces

alexweininger:
Browsing accounts, creating accounts, etc. still works. Haven't tested everything but I guess declaring limited support is better than being disabled by default.

### DIFF
--- a/package.json
+++ b/package.json
@@ -76,6 +76,12 @@
     ],
     "main": "main",
     "icon": "resources/storageAccount.png",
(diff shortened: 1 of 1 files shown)


bwateratmsft (approved):

nturinski:
I just realized, but unless we enable Resource Groups and Azure Account extension to be limited support for virtual workspaces, this change won't actually do anything since they are dependencies.

PR #1126, opened 2022-08-22
Add local workspace support options to 'Create Queue' command via palette

### LINKED ISSUE #1043: There is no option "Attached Storage Accounts" in the command list when executing command from command palette
**OS:** All
**Azure Resource Extension Build Version:** 20220505.12 (unified)
**Azure Storage Extension Build Version:** 20220505.2 (unified)
**Regression:** Not a regression

**Repro Steps:**
1. F1 -> Azure Storage: Create Queue -> Select a subscription.
2. Check whether there is an option "Attached Storage Accounts" in the command list.

**Expect:**
There is an option "Attached Storage Accounts" in the command list.
image

**Actual:**
1. There is no option "Attached Storage Accounts" in the command list.
image
2. Can not to execute all commands from command palette for attached storage accounts.

MicroFish91:
Fixes #1043 

I'm proposing to make changes to `pickAppResource` [here](https://github.com/microsoft/vscode-azureresourcegroups/pull/369) and [here](https://github.com/microsoft/vscode-azuretools/pull/1228).

Basically my change does this - when creating queue from the `command palette`, the new change will first ask if the node we are looking for is `remote` or `local`.  Based on this, we will either traverse the app resources or the workspace resources.

Screen Shot 2022-08-22 at 10 28 47 AM

### DIFF
--- a/src/commands/queue/queueGroupActionHandlers.ts
+++ b/src/commands/queue/queueGroupActionHandlers.ts
@@ -15,10 +15,6 @@ export function registerQueueGroupActionHandlers(): void {
 }
 
 export async function createQueue(context: IActionContext, node?: QueueGroupTreeItem): Promise<void> {
-    if (node?.root.isEmulated && !(await isAzuriteInstalled())) {
-        warnAzuriteNotInstalled(context);
-    }
-
     if (!node) {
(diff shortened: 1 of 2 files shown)


MicroFish91:
Leaving this implementation for v2 of the API

PR #1133, opened 2022-08-26
Add Delete Blob Support to Activity Logs

motm32:


src/tree/blob/BlobTreeItem.ts
```diff
@@ -67,18 +69,26 @@ export class BlobTreeItem extends AzExtTreeItem implements ICopyUrl, IDownloadab
...
     public async deleteTreeItemImpl(context: ISuppressMessageContext): Promise<void> {
-        let result: MessageItem | undefined;
-        if (!context.suppressMessage) {
-            const message: string = `Are you sure you want to delete the blob '${this.label}'?`;
-            result = await window.showWarningMessage(message, { modal: true }, DialogResponses.deleteResponse, DialogResponses.cancel);
-        }
-        if (result === DialogResponses.deleteResponse || context.suppressMessage) {
+        if (context.suppressMessage) {
             const blobClient: azureStorageBlob.BlobClient = createBlobClient(this.root, this.container.name, this.blobPath);
             await blobClient.delete();
```
alexweininger, line 74: We should follow the pattern we used in [`BlobDirectoryTreeItem#deleteTreeItemImpl`](https://github.com/microsoft/vscode-azurestorage/blob/4258e1acbb3e16bd46a7f7c82aa3c6d1bd8d3b61/src/tree/blob/BlobDirectoryTreeItem.ts#L97)

In the current implementation, if `context.suppressMessage` is true, then no activity will appear in the activity log.

src/commands/deleteBlob/DeleteBlobStep.ts
```diff
@@ -0,0 +1,33 @@
...
+        const blobName = nonNullProp(wizardContext, 'blobName');
+        const blob = nonNullProp(wizardContext, 'blob');
+        const deletingBlob: string = localize('deletingBlob', 'Deleting blob "{0}"...', blobName);
+        ext.outputChannel.appendLog(deletingBlob);
+        progress.report({ message: deletingBlob });
+        const blobClient: azureStorageBlob.BlobClient = createBlobClient(blob.root, blob.container.name, blob.blobPath);
+        await blobClient.delete();
+        const deleteSuccessful: string = localize('successfullyDeletedBlob', 'Successfully deleted blob "{0}".', blobName);
+        ext.outputChannel.appendLog(deleteSuccessful);
+        void window.showInformationMessage(deleteSuccessful);
```
alexweininger, line 27: We should only show this information message when `wizardContext.suppressNotification` is false. This value is true by default since we don't want to show a notification and show the activity log by default.

src/tree/blob/BlobTreeItem.ts
```diff
@@ -67,18 +69,26 @@ export class BlobTreeItem extends AzExtTreeItem implements ICopyUrl, IDownloadab
...
+            const message: string = localize('deleteBlob', "Are you sure you want to delete the blob '{0}'?", this.label);
+            const wizard = new AzureWizard(wizardContext, {
+                promptSteps: [new DeleteConfirmationStep(message)],
+                executeSteps: [new DeleteBlobStep()]
+            });
+            await wizard.prompt();
+            await wizard.execute();
+        } else {
             const blobClient: azureStorageBlob.BlobClient = createBlobClient(this.root, this.container.name, this.blobPath);
             await blobClient.delete();
```
alexweininger, line 90: > We should follow the pattern we used in [`BlobDirectoryTreeItem#deleteTreeItemImpl`](https://github.com/microsoft/vscode-azurestorage/blob/4258e1acbb3e16bd46a7f7c82aa3c6d1bd8d3b61/src/tree/blob/BlobDirectoryTreeItem.ts#L97)
> 
> In the current implementation, if `context.suppressMessage` is true, then no activity will appear in the activity log.

The pattern I was referring to was only calling `blobClient.delete()` in `DeleteBlobStep`. And only calling `wizard.prompt()` if `!context.suppressMessage`.

https://github.com/microsoft/vscode-azurestorage/blob/4258e1acbb3e16bd46a7f7c82aa3c6d1bd8d3b61/src/tree/blob/BlobDirectoryTreeItem.ts#L112-L114

That way we are only calling `blobClient.delete()` in one place. And this ensures an activity appears regardless of what `context.suppressMessage` is set to.

src/commands/deleteBlob/DeleteBlobStep.ts
alexweininger: in case you missed this 😄

src/tree/blob/BlobTreeItem.ts
bwateratmsft: Would it be possible to nest all the blob nodes under the directory node?

alexweininger: It's possible, but I don't think we want that UX here. I'd prefer if we just had a single activity appear for the delete-an-entire-blob command.

alexweininger (approved):

Issue #998, opened 2021-08-18
Accessing a file from Azure Blob Storage using VSC


minsa110:
As a student, I want to be able to download a file from Azure Blob Storage using VSC extensions. 

**Observation:** This is not currently possible without an Azure account/if you're not downloading a file that someone else is hosting in Blob Storage. Current experience for students in classroom is:
1.	A professor provides a link to a data file hosted in Blob Storage
2.	Students paste the link to a browser, then download it to their machine
3.	Students move the data file into their local repo location and opens VSC

**Suggestion:** New feature where you can paste and download a file from a URL, and it will open in VS Code. AKA remove step 2 in the process, where students would be able to download the data from Azure Blob by pasting the link directly on VSC (via the Azure Storage extension) without going to the browser.

**Benefit:** Can access files that others share with you via URL from storage right from VS Code, without going to browser. Provide a smooth experience for sharing datasets.

nturinski:
As part of this feature, I'm planning to add the following:

- [ ] Generate a SAS token and URL for Blob Containers and File Shares and their children
- [ ] Downloading a container or file share by right click
- [ ] Downloading a container, file share, directory, or file by passing in a SAS url
- [ ] Uploading a container, file share, directory, or file by passing in a SAS url

PR #1151, opened 2022-09-07
Add a download from Sas url command to local Workspace view

nturinski:
Fixes https://github.com/microsoft/vscode-azurestorage/issues/998

I renamed everything to use `Sas` rather than `SAS`. It was hard on the eyes.

I'm not married to the idea, but for now, the entry point is on the local workspace bar:
image

I figured once we have an actions view, we can migrate it to there but for now, I thought it looked rather nice here. Hard to discover though.

Rather than dealing with many entry points for one command, I refactored download to have tree items and SAS url as two separate entry points, but then both use the `download` function to minimize code changes.

EDIT: Planning to add unit tests, but wanted to get the PR out first.

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,7 +50,8 @@
         "onCommand:azureStorage.createBlockBlob",
         "onCommand:azureStorage.uploadFiles",
         "onCommand:azureStorage.uploadFolder",
-        "onCommand:azureStorage.download",
+        "onCommand:azureStorage.downloadTreeItems",
+        "onCommand:azureStorage.downloadSasUrl",
         "onCommand:azureStorage.createFileShare",
         "onCommand:azureStorage.deleteFileShare",
         "onCommand:azureStorage.deleteStorageAccount",
@@ -72,7 +73,7 @@
         "onCommand:azureStorage.startBlobEmulator",
         "onCommand:azureStorage.startQueueEmulator",
         "onCommand:azureStorage.reportIssue",
-        "onCommand:azureStorage.generateSASUrl",
+        "onCommand:azureStorage.generateSasUrl",
         "onView:azureWorkspace"
     ],
     "main": "main",
@@ -199,9 +200,15 @@
                 "category": "Azure Storage"
             },
             {
-                "command": "azureStorage.download",
+                "command": "azureStorage.downloadTreeItems",
                 "title": "Download...",
                 "category": "Azure Storage"
+            },
+            {
+                "command": "azureStorage.downloadSasUrl",
+                "title": "Download by SAS Url...",
+                "category": "Azure Storage",
+                "icon": "$(cloud-download)"
             },
             {
                 "command": "azureStorage.createFileShare",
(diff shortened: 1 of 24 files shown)


src/commands/downloadFile.ts
```diff
@@ -61,13 +69,30 @@ export async function download(context: IActionContext, treeItem: IDownloadableT
...
+    const wizard: AzureWizard<ISasDownloadContext> = new AzureWizard(context, {
+        promptSteps: [new SasUrlPromptStep()],
+        executeSteps: []
+    });
+
+    await wizard.prompt();
+    return await download(context, nonNullProp(context, 'sasUrl'));
+}
+export async function downloadTreeItems(context: IActionContext, treeItem: IDownloadableTreeItem, treeItems?: IDownloadableTreeItem[]): Promise<void> {
+    treeItems = treeItems || [treeItem];
```
alexweininger, line 82: I think we can be fancier here 🍸 

```suggestion
    treeItems ??= [treeItem];
```

src/commands/downloadFiles/SasUrlPromptStep.ts
```diff
@@ -0,0 +1,51 @@
...
+        }
+
+        context.sasUrl = await context.ui.showInputBox({
+            prompt: localize('enterSasUrl', 'Enter a Azure Storage SAS URL'),
+            validateInput: async (s): Promise<string | undefined> => await this.validateInput(context, s),
+            value
+        });
+    }
+
+    public shouldPrompt(_wizardContext: ISasDownloadContext): boolean {
```
alexweininger, line 26:
```suggestion
    public shouldPrompt(): boolean {
```

alexweininger (dismissed):

src/commands/downloadFile.ts
alexweininger: Or maybe `||=`, not sure which is better here...

package.json
```diff
@@ -199,10 +200,16 @@
                 "category": "Azure Storage"
             },
             {
-                "command": "azureStorage.download",
```
alexweininger, line 202: Just making sure, no extensions depend on this command right?

alexweininger:
Something is up with the build 🤔

package.json
nturinski: Pretty sure they aren't. I don't think we allow any other extensions to download storage resources. We probably would have an API if that were the case.

src/commands/downloadFiles/SasUrlPromptStep.ts
```diff
@@ -0,0 +1,51 @@
...
+                !url.searchParams.get('se') ||
+                !url.searchParams.get('sig')
+            ) {
+                return localize('enterValidToken', 'The SAS token is missing a parameter. Enter a valid SAS token.',)
+            }
+        } catch (err) {
+            return localize('enterValidSasUrl', 'The URL "{0}" is not valid. Enter a valid URL', value);
+        }
+
+        return;
```
bwateratmsft, line 49:
```suggestion
        return undefined;
```
I'd prefer `return;` only be used when the return type is `void`.

src/commands/downloadFile.ts
```diff
@@ -61,13 +69,30 @@ export async function download(context: IActionContext, treeItem: IDownloadableT
...
 }
 
-async function getAzCopyDownloads(context: IActionContext, destinationFolder: string, treeItems: IDownloadableTreeItem[]): Promise<IAzCopyDownload[]> {
+export async function downloadSasUrl(context: ISasDownloadContext): Promise<void> {
+    const wizard: AzureWizard<ISasDownloadContext> = new AzureWizard(context, {
+        promptSteps: [new SasUrlPromptStep()],
+        executeSteps: []
+    });
+
+    await wizard.prompt();
```
bwateratmsft, line 78: Why not make download part of the wizard?

nturinski: It'd be a bit of a refactor since I'm shimming this into a pre-existing function, but I'm down to do that. I'd prefer to get this merged and then refactor so that the implementation changes are more obvious though.

bwateratmsft (approved):

alexweininger (approved):

PR #1152, opened 2022-09-07
Fixed file uploaded to existing directory showing up in the wrong place

### LINKED ISSUE #1143: File uploaded to existing directory shows up in the wrong place
**Repro steps**
1. Have an existing directory in a blob container
2. Right click on the blob container and click upload files
3. Type in existing directory for destination directory
4. Choose a file to upload and click Upload

**Expect**
The file to appear under the existing directory 

**Actual**
File appears under the blob container
image

After refresh the file moves under the directory

motm32:
Fixes #1143. I modeled this change after UploadLocalFolder. Not sure if we would want to do it this way but it does solve the issue.


nturinski (approved):
> I modeled this change after UploadLocalFolder. Not sure if we would want to do it this way but it does solve the issue.

We were doing this because there were many cases where the validation wasn't actually working, but we were still creating the tree item anyway. It's just easier to poll Azure afterwards to see if it actually created or not.

Issue #1147, opened 2022-09-02
The default specify quota is not 5120 when creating a new file share during uploading files


v-chenzha:
**OS:** Windows 10
**Build:** 20220901.11

**Repro Steps:**

1. Create a new storage account -> F1 -> Azure Storage: Upload Files... -> Select a file -> Select "File Share" -> Select "+ Create new File Share..." -> Type a name -> Press Enter.
2. Check whether the Specify quota is 5120 by default.

**Expect:**
The default specify quota is 5120.
![image](https://user-images.githubusercontent.com/107091828/188072908-1175eb1e-aa88-49d1-b992-21b19e108be1.png)

**Actual:**
The default specify quota is the same as the File Share name.
![image](https://user-images.githubusercontent.com/107091828/188072838-e4407715-d49b-4e05-8960-1fe4420ab095.png)

**More Info:**
This issue also reproduces when creating a new file share during executing "Upload Folder..." command.

motm32:
This issue is caused by the cache values being mixed up since there are multiple `showInputBox` calls being made in the same wizard step. See [FileShareGroupTreeItem](https://github.com/microsoft/vscode-azurestorage/blob/main/src/tree/fileShare/FileShareGroupTreeItem.ts#:~:text=public%20async%20createChildImpl(context%3A%20ICreateChildImplContext)%3A%20Promise%3CFileShareTreeItem%3E).

alexweininger:
Caused by https://github.com/microsoft/vscode-azurestorage/pull/1084

MicroFish91:
Ahh I see, it looks like involving a wizard means adding a cached value check behind the scenes:

![Screen Shot 2022-09-06 at 10 33 04 AM](https://user-images.githubusercontent.com/40250218/188702055-948f8729-be54-4e57-86ea-d2e749196ab3.png)

Could we do something like this to fix this issue:

`inputBox.value = options.value || wizard?.getCachedInputBoxValue() || '';`

Or do you still want to wait on a separate change related to v2?

alexweininger:
> Could we do something like this to fix this issue:
> 
> `inputBox.value = options.value || wizard?.getCachedInputBoxValue() || '';`
> 
> Or do you still want to wait on a separate change related to v2?

That would break a lot of wizards since it would always override the cached value with whatever the initial value was.

We will need to discuss as a team. The correct solution could be to create a new context for these extra prompts in a wizard step. Or to base the cache key on something more unique than just the wizard step.

PR #1153, opened 2022-09-08
Refactor createFileShare into wizard prompt steps

nturinski:
Fixes https://github.com/microsoft/vscode-azurestorage/issues/1147

### DIFF
--- a/src/tree/fileShare/createFileShare/FileShareNameStep.ts
+++ b/src/tree/fileShare/createFileShare/FileShareNameStep.ts
@@ -1,0 +1,46 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { AzureWizardPromptStep } from "@microsoft/vscode-azext-utils";
+import { localize } from "../../../utils/localize";
+import { IFileShareWizardContext } from "./IFileShareWizardContext";
+
+export class FileShareNameStep extends AzureWizardPromptStep<IFileShareWizardContext> {
+    public async prompt(context: IFileShareWizardContext): Promise<void> {
+        context.name = await context.ui.showInputBox({
+            placeHolder: localize('enterFileShareName', 'Enter a name for the new file share'),
+            validateInput: this.validateFileShareName
+        });
+    }
+
+    public shouldPrompt(context: IFileShareWizardContext): boolean {
+        return !context.name;
+    }
+
+    private validateFileShareName(name: string): string | undefined | null {
+        const validLength = { min: 3, max: 63 };
+
+        if (!name) {
+            return localize('shareNameEmpty', "Share name cannot be empty");
+        }
+        if (name.indexOf(" ") >= 0) {
+            return localize('shareNameSpaces', "Share name cannot contain spaces");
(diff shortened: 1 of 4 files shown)


src/tree/fileShare/createFileShare/IFileShareWizardContext.ts
```diff
@@ -0,0 +1,11 @@
+/*---------------------------------------------------------------------------------------------
+*  Copyright (c) Microsoft Corporation. All rights reserved.
+*  Licensed under the MIT License. See License.txt in the project root for license information.
+*--------------------------------------------------------------------------------------------*/
+
+import { IActionContext } from "@microsoft/vscode-azext-utils";
+
+export interface IFileShareWizardContext extends IActionContext {
+    name?: string;
+    quota?: string;
```
bwateratmsft, line 10: Why not just make it a `number`?

src/tree/fileShare/createFileShare/StorageQuotaPromptStep.ts
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    public shouldPrompt(context: IFileShareWizardContext): boolean {
+        return !context.quota;
+    }
+
+    private validateQuota(input: string): string | undefined {
+        const value = Number(input);
+        if (isNaN(value)) {
+            return localize('quotaNum', "Input must be a number");
```
bwateratmsft, line 29: Nit: quote and term consistency
```suggestion
            return localize('quotaNum', 'Value must be a number');
```

src/tree/fileShare/createFileShare/FileShareNameStep.ts
```diff
@@ -0,0 +1,46 @@
...
+        return !context.name;
+    }
+
+    private validateFileShareName(name: string): string | undefined | null {
+        const validLength = { min: 3, max: 63 };
+
+        if (!name) {
+            return localize('shareNameEmpty', "Share name cannot be empty");
+        }
+        if (name.indexOf(" ") >= 0) {
```
bwateratmsft, line 28: Maybe more robust?
```suggestion
        if (/\s/.test(name)) {
```

bwateratmsft, line 28: Since all the `if` blocks are `return`s it's not strictly necessary, but IMO these should all be else-ifs, so that it's more immediately obvious that that is the true behavior.

src/tree/fileShare/createFileShare/IFileShareWizardContext.ts
nturinski: I dunno, I was just following what was already there without much thought 🤣

src/tree/fileShare/createFileShare/FileShareNameStep.ts
```diff
@@ -0,0 +1,41 @@
...
+            placeHolder: localize('enterFileShareName', 'Enter a name for the new file share'),
+            validateInput: this.validateFileShareName
+        });
+    }
+
+    public shouldPrompt(context: IFileShareWizardContext): boolean {
+        return !context.name;
+    }
+
+    private validateFileShareName(name: string): string | undefined | null {
```
alexweininger, line 22: Love that we are able to move this code out of FileShareGroupTreeItem.ts

alexweininger (approved):

bwateratmsft (approved):

PR #1157, opened 2022-09-13
Fix the regExp that matches the id of the tree item

nturinski:
Fixes https://github.com/microsoft/vscode-azureresourcegroups/issues/377

Due to https://github.com/microsoft/vscode-azureresourcegroups/commit/17033f24577c52ba19af7188f5169f598423b6ee, the id of the GroupTreeItems changed on every Azure Group Tree Item.  That changed the form of expected id, which AzureStorageFS needs to look up tree items in the tree view.

This is changing the regexp to match the new expected id form.


alexweininger (approved):
Can you add some context about this in the description of the PR?

Like why/when did the IDs change in this way

PR #1158, opened 2022-09-13
Fix context values for storage account create

nturinski:
Fixes https://github.com/microsoft/vscode-azureresourcegroups/issues/375


alexweininger (approved):
Now we can remove it from https://github.com/microsoft/vscode-azureresourcegroups/blob/e8bef7c4e36a149df583654ad60f35fa48d03f4b/src/azureExtensions.ts#L93

PR #1169, opened 2022-09-15
Replace `Copy to Clipboard` with `View Output` for uploading multiple files

### LINKED ISSUE #1164: [Suggestion] It would be better to copy all URLs for the uploaded files when clicking "Copy to Clipboard" button
**OS:** All
**Build Version:** 20220913.12

**Repro Steps:**
1. Right click one blob container -> Select "Upload Files..." -> Select multiple files to upload -> Click "Copy to Clipboard" button in the below notification.
image
2. Check whether all URLs are copied or not.

**Actual:**
Only the URL of the last uploaded file is copied.

**Suggestion:**
It would be better to copy all URLs for the uploaded files when clicking "Copy to Clipboard" button.

MicroFish91:
Closes #1164 

These toasts/information message windows should eventually be replaced with informational/commandId Activity Log children.  For now, for the case of uploading multiple files, instead of providing a copy button, we can just show an option to view output.

Let me know what you guys think!  I thought removing informational boxes entirely would look too minimal without full implementation of the activity log children.

### DIFF
--- a/src/utils/uploadUtils.ts
+++ b/src/utils/uploadUtils.ts
@@ -77,16 +77,33 @@ export function outputAndCopyUploadedFileUrls(parentUrl: string, fileUrls: strin
         ext.outputChannel.appendLog(url);
     }
 
-    void vscode.window.showInformationMessage(
-        localize('outputAndCopyFinished', 'Finished uploading {0} {1}.', fileUrls.length, fileUrls.length === 1 ? 'file' : 'files'),
-        localize('copyToClipboard', 'Copy to Clipboard')
-    ).then(async (result) => {
-        const shouldCopy: boolean = !!result;
-        if (shouldCopy) {
-            const lastFileUrl: string = `${parentUrl}/${fileUrls[fileUrls.length - 1]}`;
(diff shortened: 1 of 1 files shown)


src/utils/uploadUtils.ts
```diff
@@ -77,16 +77,33 @@ export function outputAndCopyUploadedFileUrls(parentUrl: string, fileUrls: strin
...
+            if (shouldCopy) {
+                const lastFileUrl: string = `${parentUrl}/${fileUrls[fileUrls.length - 1]}`;
+                await copyAndShowToast(lastFileUrl, 'File URL');
+            }
+        });
+    } else {
+        void vscode.window.showInformationMessage(
+            outputAndCopyFiles,
+            viewOutput
+        ).then(async (result) => {
```
nturinski, line 100:
```suggestion
        ).then(result => {
```

nturinski: Technically don't need the async since nothing is awaited.

nturinski:
A little unconventional, but since we want to give the build to CTI, I'll merge this PR.

PR #1194, opened 2023-01-12
Fix AzureStorageFS id parsing

alexweininger:
Our Ids are no longer... "messed up" like in v1, so we have to change the regex back to normal.


bwateratmsft (dismissed):
Is this compatible with both v1 and v2?

alexweininger:
> Is this compatible with both v1 and v2?

It's not. 🤦 I'll need to try to account for both types of ids.

bwateratmsft (approved):

PR #1195, opened 2023-01-12
Initial conversion to a v1.5 extension

alexweininger:


bwateratmsft (approved):
Ugh. I don't love command registrations being scattered across so many files...

PR #1210, opened 2023-02-07
Add refresh to upload file/folder

### LINKED ISSUE #1208: Have to refresh the tree in order to see uploaded files and folders
_v2 forward compat_

Occurs for both blob containers and file shares. 

Note: This issue happens when using the context menu and the command palette

motm32:
Fixes #1208.

### DIFF
--- a/src/commands/uploadFiles/uploadFiles.ts
+++ b/src/commands/uploadFiles/uploadFiles.ts
@@ -43,5 +43,7 @@ export async function uploadFiles(
     }
 
     await wizard.execute();
+    await nTreeItem.refresh(context);
+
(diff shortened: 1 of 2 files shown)


alexweininger (commented):
I figured out that #1208 is caused by `clearCache` not getting set to true on refresh in the RGs compat shim AzExtTreeDataProvider. Your changes do fix this by calling directly calling refresh on the item, which uses the old AzExtTreeDataProvider, which sets `clearCache = true`. 

However, this fix needs to be applies in RGs and not here. See https://github.com/microsoft/vscode-azureresourcegroups/pull/554 for how I did that. Please let me know if you have questions because I know this is overly complex.

PR #1246, opened 2023-06-09
Add nice error when creating directories is not supported

### LINKED ISSUE #1245: Fail to create a new folder in Explorer after opening a blob container in Explorer
**OS:** All
**Build Version:** 20230607.4

**Repro Steps:**
1. Create a storage account -> Create a blob container.
2. Right click the blob container -> Click "Open in Explorer..." -> Select "Open in new window".
3. Click the "New Folder..." button in Explorer to create a new folder.
4. Check whether succeeds to create a new folder in Explorer.

**Expect:**
Succeed to create a new folder in Explorer.

**Actual:**
Fail to create a new folder in Explorer with an error.
image

**More Info:**
1. This issue does not reproduce when creating a new file.
2. This issue also reproduces on the current released extension 0.15.2.

alexweininger:
Fixes #1245

### DIFF
--- a/src/BlobContainerFS.ts
+++ b/src/BlobContainerFS.ts
@@ -337,13 +337,13 @@ export class BlobContainerFS implements vscode.FileSystemProvider {
     async createDirectory(uri: vscode.Uri): Promise<void> {
         const result = await callWithTelemetryAndErrorHandling<void>("azureStorageBlob.createDirectory", async (context) => {
             context.errorHandling.rethrow = true;
+            context.errorHandling.suppressDisplay = true;
 
             const { storageAccount, accountKey } = await this.getStorageAccount(uri, context);
             const isHnsEnabled = !!storageAccount.isHnsEnabled;
 
             if (!isHnsEnabled) {
-                // Virtual directories don't need to be created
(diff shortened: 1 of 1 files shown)


src/BlobContainerFS.ts
```diff
@@ -337,13 +337,13 @@ export class BlobContainerFS implements vscode.FileSystemProvider {
     async createDirectory(uri: vscode.Uri): Promise<void> {
         const result = await callWithTelemetryAndErrorHandling<void>("azureStorageBlob.createDirectory", async (context) => {
             context.errorHandling.rethrow = true;
+            context.errorHandling.suppressDisplay = true;
```
bwateratmsft, line 340: Did you mean `suppressReportIssue` here? If not, should a `UserCancelledError` be thrown instead?

alexweininger: VS Code displays the errors thrown inside these methods automatically. If we don't suppress display here then the error shows up twice since we're rethrowing. We have to rethrow to tell VS Code that the operation didn't succeed.

bwateratmsft (approved):

PR #1277, opened 2023-09-11
Isolate transfer related code/create transfer-tech agnostic function for downloading

MRayermannMSFT:
In order to be compatibile with web/vscode.dev, we need to be able to transfer data without using AzCopy. Currently there are many different places which import `azcopy-node` types and APIs. My goal is to for ther to be only one module which has exposure to `azcopy-node` types and APIs. Thus, this change:

1. Lays the groundwork for centralizing transfer related code paths, such that we can have a technology agnostic module that can start uploads & downloads. This module is what ultimately decides if it should use AzCopy or something else.
2. Implements downloads in the module. Note: I did some slight refactoring of existing logic, let me know if y'all want this as a separate change.

Code for uploads is much more spread out and numerous, which will result in a much larger change when it is moved. Therefore I am putting this change out first to get review on the overall design.


src/commands/transfers/downloadFiles.ts
```diff
@@ -0,0 +1,95 @@
+/*---------------------------------------------------------------------------------------------
```
MRayermannMSFT, line 1: The code for this file used to exist in/effectively came from DownloadFilesStep.ts

alexweininger (approved):

PR #1278, opened 2023-09-12
Remove `getResourceUri` and `getSasToken` util functions, add them to `ITransferSrcOrDstTreeItem` (formerly `IDownloadableTreeItem`) instead

MRayermannMSFT:


src/commands/downloadFiles/GetAzCopyDownloadsStep.ts
```diff
@@ -50,7 +47,7 @@ export class GetAzCopyDownloadsStep extends AzureWizardExecuteStep<IDownloadWiza
                     remoteFilePath,
                     localFilePath: join(destinationFolder, treeItem.blobName),
                     isDirectory: false,
-                    resourceUri,
+                    resourceUri: treeItem.resourceUri,
```
nturinski, line 50: nit: Could create a `const resourceUri = treeItem.resourceUri` in the for loop to keep the objects being pushed cleaner.

src/tree/blob/BlobDirectoryTreeItem.ts
```diff
@@ -63,6 +64,21 @@ export class BlobDirectoryTreeItem extends AzExtParentTreeItem implements ICopyU
         return new vscode.ThemeIcon('folder');
     }
 
+    public get resourceUri(): string {
```
nturinski, line 67: As far as I can tell, all of the blobTreeItems share the same `getters` for `resourceUri` and `transferSasToken`. Would you be open to creating a base class that implements it that could just be extended?

src/tree/fileShare/DirectoryTreeItem.ts
```diff
@@ -51,6 +52,21 @@ export class DirectoryTreeItem extends AzExtParentTreeItem implements ICopyUrl,
         return path.posix.join(this.parentPath, this.directoryName);
     }
 
+    public get resourceUri(): string {
```
nturinski, line 55: Similar comment, but for fileshares.

nturinski (dismissed):
I'd be okay if you didn't want to abstract the commands out since this is already a clean, vast improvement.

src/tree/fileShare/DirectoryTreeItem.ts
MRayermannMSFT: Will leave for someone who can take a look across all storage tree item classes.

src/tree/blob/BlobDirectoryTreeItem.ts
MRayermannMSFT: Will leave for someone who can take a look across all storage tree item classes.

nturinski (approved):

PR #1281, opened 2023-09-14
Identify which storage imports are type vs runtime imports as prep for stg sdk lazy load

MRayermannMSFT:


src/tree/blob/BlobDirectoryTreeItem.ts
```diff
@@ -65,14 +66,28 @@ export class BlobDirectoryTreeItem extends AzExtParentTreeItem implements ICopyU
...
+        const containerClient: ContainerClient = this.root.createBlobServiceClient().getContainerClient(this.container.name);
         return containerClient.url;
     }
 
     public get transferSasToken(): string {
-        const accountSASSignatureValues: azureStorageBlob.AccountSASSignatureValues = {
+        const accountSASSignatureValues: AccountSASSignatureValues = {
             expiresOn: new Date(Date.now() + threeDaysInMS),
-            permissions: azureStorageBlob.AccountSASPermissions.parse('rwl'), // read, write, list
+            permissions: {
```
MRayermannMSFT, line 76: I went ahead and changed this to not use the `.parse` because I anticipate that being a problem once I make lazy load SDK functions/classes thing. :(

alexweininger (approved):

PR #1282, opened 2023-09-14
Polyfill before every import of storage SDKs, update webpack config for needed fallbacks and aliases

MRayermannMSFT:
I originally intended to use this PR to swap all non-type of imports of the storage SDKs to be async imports, however the changes were incredibly numerous. I talked to Nathan about it and we decided I should try to see if we could just get away with making sure we polyfill before importing a storage SDK/s. The answer to this was yes, so I'm going with this solution.


src/tree/blob/BlobContainerTreeItem.ts
```diff
@@ -63,21 +68,7 @@ export class BlobContainerTreeItem extends AzExtParentTreeItem implements ICopyU
...
-                add: false,
-                create: false,
-                update: false,
-                process: false,
-                tag: false,
-                filter: false,
-                setImmutabilityPolicy: false,
-                permanentDelete: false
-            },
+            permissions: AccountSASPermissions.parse("rwl"), // read, write, list
```
MRayermannMSFT, line 71: I decided to bring back use of `.parse` since I don't have to async import the Storage SDKs.

alexweininger (approved):
I'm totally fine with these changes. However, I do have one idea you could explore if you want.

In some personal projects, I've had success using [patch-package](https://www.npmjs.com/package/patch-package) to apply small edits or fixes to packages without the need to fork and republish.

I wonder if you could utilize patch-package to add the polyfill to the storage SDKs themselves   to avoid having to load the polyfill manually wherever we use the SDKs.

Just a suggestion, may or may not actually be better.

PR #1302, opened 2024-03-27
Simplify AzureStorageFS readFile and pass along contents when creating a new blob in writeFile

### LINKED ISSUE #1286: Drag & drop to upload to Azurite doesn't really drop the files/blobs
Reported by PaulYuk.

Repro:

go to Azure panel in VS Code
Workspace-> Local Emulator -> Blob containers -> images
open images in explorer (new window)
drag and drop from Finder or Windows Explorer into that container
It sees files and looks uploaded.  but then you can't open image, or in code/streams

image
image

Same steps, but with Azure storage account, all is Good.

I was able to reproduce this.  The workaround is to use the context menu to upload/download.  It's likely we didn't wire up drag & drop if/when Azurite is used.  Not sure if technically it's feasible.  Need some investigations.

richardMSFT:
Fixes #1286 

Previously, trying to view the blobs in an emulated blob container by opening a new explorer window would error when loading said blobs. This is resolved by using `client.downloadToBuffer()` similar to what is done in `readFile()` of BlobContainerFS.ts instead of manually handling and reading the chunks of the stream.

Also, dragging and dropping files into the explorer of an emulated blob container was not behaving as expected because `writeFile()` inside `AzureStorageFS.ts` which handles writing operations for emulated containers (opened in a new explorer window) creates an empty text file for newly created blobs. This fix passes along the blob contents to the existing `containerClient` calls where we perform blob uploads.

out

### DIFF
--- a/src/AzureStorageFS.ts
+++ b/src/AzureStorageFS.ts
@@ -3,8 +3,8 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import type { BlobClient, BlobDownloadResponseModel, BlobGetPropertiesResponse, BlockBlobClient } from "@azure/storage-blob";
-import type { FileDownloadResponseModel, FileGetPropertiesResponse, ShareFileClient } from "@azure/storage-file-share";
+import type { BlobClient, BlobGetPropertiesResponse, BlockBlobClient } from "@azure/storage-blob";
+import type { FileGetPropertiesResponse, ShareFileClient } from "@azure/storage-file-share";
 
 import { AzExtTreeItem, IActionContext, UserCancelledError, callWithTelemetryAndErrorHandling, parseError } from "@microsoft/vscode-azext-utils";
 import * as path from "path";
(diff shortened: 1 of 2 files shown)


src/AzureStorageFS.ts
```diff
@@ -233,30 +233,28 @@ export class AzureStorageFS implements vscode.FileSystemProvider, vscode.TextDoc
...
                 if (treeItem instanceof FileShareTreeItem) {
                     client = createFileClient(treeItem.root, treeItem.shareName, parsedUri.parentDirPath, parsedUri.baseName);
+                    buffer = await client.downloadToBuffer();
                 } else {
                     client = createBlobClient(treeItem.root, treeItem.container.name, parsedUri.filePath);
+                    buffer = await client.downloadToBuffer();
                 }
-                downloaded = await client.download();
-                result = await this.streamToString(downloaded.readableStreamBody);
+                return buffer;
```
nturinski, line 250: nit: Couldn't you just

`return await client.downloadToBuffer()` here? Since both clients seem like they have the exact same function definition, I don't think that'd cause any issues.

nturinski (approved):
Slick! Looks good; thanks for doing this.

src/AzureStorageFS.ts
richardMSFT: I considered doing that but es-lint doesn't approve because the function signatures for `ShareFileClient` and `BlobClient` are different for their respective `downloadToBuffer()` function call. Even though both are passing in no arguments because we are using the signature with all optional parameters, `ShareFileClient` expects the optional parameter be `FileDownloadToBufferOptions` while `BlobClient` expects `BlobDownloadToBufferOptions`.

This seemed to be the simplest solution even though it is repetitive. Let me know tho if there is something else that is preferred!

alexweininger:
Epic

PR #1322, opened 2024-07-03
Update long running tests check

MicroFish91:
See context from this PR: https://github.com/microsoft/vscode-azurecontainerapps/pull/713


nturinski (approved):

package.json
```diff
@@ -833,7 +833,11 @@
         "webpack-cli": "^4.6.0"
     },
     "dependencies": {
+        "@azure-tools/azcopy-darwin": "^10.25.1",
```
nturinski: We need to do something about this. I think npm install is automatically putting all of the azcopys in the package.json, but that causes errors in the build.

However, it's necessary for when we package the vsix since we need all 3 versions of azcopy bundled with the extension.

Anyway, just go ahead and remove these lines for now.

nturinski (approved):

package.json
nturinski: Never mind-- from a stale commit. But we should still do something about it.

PR #1354, opened 2024-12-04
Fix file permissions for some AzCopy dependencies

### LINKED ISSUE #1346: Fail to deploy a folder to the storage account with an error "UnexpectedQuit"
**OS:** Mac (M1 Pro)/Linux
**Build Version:** v0.16.2-alpha
**Regression**: Previous release(v0.16.1)

**Repro Steps:**
1. Open the "[dist.zip](https://github.com/user-attachments/files/17289959/dist.zip)" folder in VS Code.
2. Deploy the folder to a storage account.
3. Check whether succeeds to deploy the folder to the storage account.

**Expect:**
Succeed to deploy a folder to the storage account.

**Actual:**
Fail to deploy a folder to the storage account with an error "UnexpectedQuit“.
Mac (M1 Pro):
Screenshot 2024-06-05 at 09 51 52
Linux:
image

**More Info:**
1. This issue doesn't reproduce on Windows OS.
2. This issue also reproduces when uploading a file/folder to Blob Containers/File Shares.
3. This issue also reproduces when downloading a file/folder from Blob Containers/File Shares.

MicroFish91:
Fix #1346

### DIFF
--- a/gulpfile.ts
+++ b/gulpfile.ts
@@ -28,6 +28,10 @@ async function cleanReadme(): Promise<void> {
 async function setAzCopyExePermissions(): Promise<void> {
     cp.exec(`chmod u+x ${path.join(__dirname, 'node_modules/@azure-tools/azcopy-darwin/dist/bin/azcopy_darwin_amd64')}`);
     cp.exec(`chmod u+x ${path.join(__dirname, 'node_modules/@azure-tools/azcopy-linux/dist/bin/azcopy_linux_amd64')}`);
+
+    // Related: https://github.com/microsoft/vscode-azurestorage/issues/1346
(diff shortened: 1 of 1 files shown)


alexweininger (approved):

gulpfile.ts
```diff
@@ -28,6 +28,10 @@ async function cleanReadme(): Promise<void> {
 async function setAzCopyExePermissions(): Promise<void> {
     cp.exec(`chmod u+x ${path.join(__dirname, 'node_modules/@azure-tools/azcopy-darwin/dist/bin/azcopy_darwin_amd64')}`);
```
MicroFish91, line 29: We may be able to get rid of these and just replace with the new paths, but I'm not 100% sure so I'm keeping both for now until one of us has more time to verify

Issue #1385, opened 2025-04-03
Support for custom configuration of Azurite


ethan-giordano-cf:
I have an issue where the port numbers Azurite runs on by default are claimed by some corporate spyware. So I moved them to another port range.

This extension has no way to configure the port numbers like the Azurite extension does. Ideally this extension would just read the settings from the Azurite extension if possible.

This was previous reported here: https://github.com/microsoft/vscode-azurestorage/issues/620

ethan-giordano-cf:
Woohoo!! Thanks!

nturinski:
My pleasure! It's not released yet, but we should be doing a release sometime early next week.

Issue #1348, opened 2024-10-16
Unable to connect to local emulator over HTTPS, and unable to detach default configuration


mnbf9rca:
I recognise that there may be 2 issues in here: unable to connect, and unable to detach default config.

Does this occur consistently? Yes
Repro steps:
1. Install Azurite (in my case as a container) and configure to use HTTPS (as per [Azurite documentation](https://learn.microsoft.com/en-us/azure/storage/common/storage-use-azurite?tabs=visual-studio%2Cblob-storage#certificate-configuration-https))
2. Install Azure Storage VS Code extension
3. try to connect to the emulated storage using the [https connection string](https://learn.microsoft.com/en-us/azure/storage/common/storage-use-azurite?tabs=visual-studio%2Cblob-storage#https-connection-strings).

Expected: you can connect to the account

Attempted resolution:
Detach the "default" conenction from the "Detach storage account" option in the command pallet but the default connection does not show up:
![Image](https://github.com/user-attachments/assets/42a08247-15cf-4fda-a395-b512052c468e)

Action: azureStorage.attachStorageAccount
Error type: Error
Error Message: Storage Account "devstoreaccount1" is already attached.

Version: 0.16.1
OS: linux
OS Release: 6.5.0-1025-azure
Product: Visual Studio Code
Product Version: 1.94.2
Language: en

<details>
<summary>Call Stack</summary>

```
AttachedStorageAccountsTreeItem.<anonymous> extension.bundle.js:2:4488928
fulfilled extension.bundle.js:2:4486357
```

</details>

MicroFish91:
Thanks for filing this issue, we will need to investigate this one and get back to you

duncanwd:
Allowing the 'default' connection to be detached could also help resolve issues when custom ports are needed due to conflicts.

nturinski:
Fixed by: https://github.com/microsoft/vscode-azurestorage/pull/1390

PR #1390, opened 2025-05-21
Use azurite settings for listening to emulator

nturinski:
Fixes https://github.com/microsoft/vscode-azurestorage/issues/1385

The sdk recognized `UseDevelopmentStorage=true` as a native connection string, however that would limit users to only being able to use the default host and port.

This will read Azurite's settings and use those if they exist.

### DIFF
--- a/src/tree/AttachedStorageAccountTreeItem.ts
+++ b/src/tree/AttachedStorageAccountTreeItem.ts
@@ -144,7 +144,8 @@ class AttachedStorageRoot extends AttachedAccountRoot {
             const blobPort = getWorkspaceSetting('blobPort', undefined, azuriteKey) || '10000';
 
             const sharedKeyCredential = new StorageSharedKeyCredentialBlob(emulatorAccountName, emulatorKey);
-            return new BlobServiceClient(`http://${blobEndpoint}:${blobPort}/${emulatorAccountName}`, sharedKeyCredential);
+            const protocol = getTransferProtocol();
+            return new BlobServiceClient(`${protocol}://${blobEndpoint}:${blobPort}/${emulatorAccountName}`, sharedKeyCredential);
         }
 
         return BlobServiceClient.fromConnectionString(this._connectionString, this._serviceClientPipelineOptions);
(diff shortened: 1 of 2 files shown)


src/tree/AttachedStorageAccountTreeItem.ts
```diff
@@ -133,11 +134,19 @@ class AttachedStorageRoot extends AttachedAccountRoot {
...
         ).toString();
     }
 
     public async createBlobServiceClient(): Promise<BlobServiceClient> {
+        if (this.isEmulated) {
+            const blobEndpoint = getWorkspaceSetting('blobHost', undefined, azuriteKey) || defaultEmulatorHost;
+            const blobPort = getWorkspaceSetting('blobPort', undefined, azuriteKey) || '10000';
+
+            const sharedKeyCredential = new StorageSharedKeyCredentialBlob(emulatorAccountName, emulatorKey);
+            return new BlobServiceClient(`http://${blobEndpoint}:${blobPort}/${emulatorAccountName}`, sharedKeyCredential);
```
MRayermannMSFT, line 147: Is there an Azurite setting for protocol?

MRayermannMSFT (dismissed):

src/tree/AttachedStorageAccountTreeItem.ts
nturinski: That's a good question. I guess this would require us changing the protocol?

{10FE66FB-4FE4-44EE-8150-584CB1C8164C}

It might be a bit overkill for this feature, but if users ask about it, I'd probably go back and implement https

MRayermannMSFT: You best know the requirements. If y'all don't need to worry about HTTPS enabling right now, then don't worry about my comment. :)

nturinski: https://github.com/microsoft/vscode-azurestorage/issues/1348

Actually, it seems like you were right to be concerned 😁

nturinski: Context on how https is done in Azurite: https://marketplace.visualstudio.com/items?itemName=Azurite.azurite#https-setup

MRayermannMSFT (dismissed):
Noice. LGTM. Would normally suggest you don't manually construct URLs, use [built-in modules ](https://developer.mozilla.org/en-US/docs/Web/API/URL)instead, but these cases are pretty simple.

PR #1432, opened 2025-08-27
Fix blob container creation for attached storage accounts

### LINKED ISSUE #1411: An error occurs when creating a blob container for attached storage account
**OS:** All
**Build Version:** 20250718.2

**Repro Steps:**
1. Create a storage account in VS Code.
2. Attach this storage account in "WORKSPACE" view.
3. Expand the attached storage account -> Expand "Blob Containers" node, no error displays under "Blob Containers" node.
4. Create a blob container for attached storage account.
5. Check whether succeeds to create a blob container without any error.

**Expect:**
Succeed to create a blob container for attached storage account.

**Actual:**
An error occurs when creating a blob container for attached storage account.
Image

**More Info:**
1. This issue also reproduces when creating a blob container in Local Emulator.
2. This issue does not reproduce for other services.

nturinski:
- Attached storage accounts can create blob containers without errors
- Azure subscription storage accounts continue to work as before

Fixes #1411

### DIFF
--- a/src/tree/blob/BlobContainerGroupTreeItem.ts
+++ b/src/tree/blob/BlobContainerGroupTreeItem.ts
@@ -52,7 +52,7 @@ export class BlobContainerGroupTreeItem extends AzExtParentTreeItem implements I
 
         const isAttached = this.fullId.startsWith("/attachedStorageAccounts/");
 
-        let containersResponse: ListContainerItem[] | ListContainersSegmentResponse;
+        let containersResponse: ContainerItem[];
         try {
             containersResponse = await this.listContainers(isAttached, this._continuationToken);
         } catch (error) {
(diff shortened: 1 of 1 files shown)


src/tree/blob/BlobContainerGroupTreeItem.ts
```diff
@@ -117,21 +132,31 @@ export class BlobContainerGroupTreeItem extends AzExtParentTreeItem implements I
         return await vscode.window.withProgress({ location: vscode.ProgressLocation.Window }, async (progress) => {
             context.showCreatingTreeItem(name);
             progress.report({ message: `Azure Storage: Creating blob container '${name}'` });
-            return await BlobContainerTreeItem.createBlobContainerTreeItem(this, await this.createBlobContainer(name));
+            const container = await BlobContainerTreeItem.createBlobContainerTreeItem(this, await this.createBlobContainer(name));
+            void this.refresh(context)
+            return container;
         });
+
```
motm32, line 124: nit: Might want to remove this extra line

src/tree/blob/BlobContainerGroupTreeItem.ts
```diff
@@ -126,11 +128,18 @@ export class BlobContainerGroupTreeItem extends AzExtParentTreeItem implements I
...
+        this._continuationToken = undefined;
         const containerClient: ContainerClient = await createBlobContainerClient(this.root, name);
         await containerClient.create();
         const isAttached = this.fullId.startsWith("/attachedStorageAccounts/");
+        const containersResponse: ListContainerItem[] = [];
+        // load all of the containers until we find the one we just created
+        do {
+            const containers = await this.listContainers(isAttached, this._continuationToken);
+            containersResponse.push(...containers);
+        } while (this._continuationToken);
```
MicroFish91, line 141: Is `this.listContainers` responsible for resetting `this._continuationToken` to exit the loop?

src/tree/blob/BlobContainerGroupTreeItem.ts
```diff
@@ -87,7 +87,7 @@ export class BlobContainerGroupTreeItem extends AzExtParentTreeItem implements I
         const response: AsyncIterableIterator<ServiceListContainersSegmentResponse> = blobServiceClient.listContainers().byPage({ continuationToken, maxPageSize: maxPageSize });
 
         const value = (await response.next()).value as ListContainersSegmentResponse;
-        this._continuationToken = value.continuationToken;
+        this._continuationToken = value.continuationToken ? value.continuationToken : undefined;
```
MicroFish91, line 90: Does this ternary end up doing the same thing, or is there another falsey value besides `undefined` that can be obtained via `value.continuationToken`?

nturinski: When `continuationToken` reaches the end of the list, it's set to `""` (empty string). However, that causes issues when you pass it into the next list call because `""` is read as a token, not as undefined.

Since `'"` is a falsey value, I'm using it to set the token explicitly to undefined.

